PR #11, opened 2014-05-08
First draft of medicalEntity extension proposal 

twamarc:
Edit: please see #492 for 2015 status of this work, using the new http://schema.org/docs/extension.html mechanism.

Dear colleagues,
I am pleased to introduce to you this healthcare community effort initiated in June 2013, to extend the medicalEntity vocabulary of schema.org. This aims to enable the use of schema.org not only in indexing health records, web, healthcare documents but also as a pillar source of medical ontology/vocabulary for formalization of healthcare information, exchange and interoperability and use in HL7 standards, CCD/C-CDA elements based APIs. 
The intention here is neither to have a comprehensive medical ontology nor to define or codify a new controlled medical vocabulary, but to provide a scalable, use-driven vocabulary schema, in-line with the schema.org philosophy. In this design, few granular classes will be created, the priority is given to use the vocabulary in clinical information models which use classes from  existing vocabularies and terminology (SNOMED, ICD, LOINC, ATC, RxNorm,etc.)  and all properties from schema.org as pillar source.
This approach have been tested and used within SALUS EU FP7 project  and was successful. All tools and services produced in SALUS EU FP7 project   are public and accessible.  

Expected use:
In line with healthcare IT community new initiatives (HL7 FHIR, CIMI, OpenEHR etc) and the real need in interoperability, the trend is now to develop clinical models not using local/internal ontology or a proper new ontology but only 3rd party ontology. This is the momentum for schema.org to contribute to the healthcare information, exchange and interoperability. The consumers and publishers will find this extension also interesting as they can now enlarge the amount of indexable content, and beyond that they can index not only information and meta-data but also the data itself.

Feel free to provide your feedback, suggestions and comments, to improve the proposal.

Regards
Marc Twagirumukiza, Agfa HealthCare N.V.
Belgium.


danbri:
@twamarc and others -


(to recap)
The Medical/health vocabulary is one we have long struggled with at schema.org. It is a hugely important topic, but also very detailed and technical. Our original vocabulary was poorly integrated, using many generic terms in med-specific ways. We were even considering removing it when this proposal arrived last year. Meanwhile, we have also been working on an improved mechanism for extending schema.org which partially decentralizes the creation of schemas, while keeping them integrated within a flat namespace. Yesterday's update to schema.org v2.0 included a revision to the extensions page, presenting that idea: http://schema.org/docs/extension.html 

We now anticipate handling some bibliographic and automobile terminology as extensions. Each of these is supported by a W3C Community Group, and would show up as e.g. bib.schema.org, auto.schema.org. Within markup the prefixes could be omitted when needed in syntax (e.g. Microdata) or used for clarity (in JSON-LD).

What do you think about this route as a way forward for medical/health schemas at schema.org?

Apologies for the long time it has taken us to respond to your hard work here. I look forward to getting things moving again...

twamarc:
I personally support this approach as a way forward (but I have not yet discussed with the group which was behind the proposal last year).
We-all of us I guess, followed with interest the recent discussions about schema extension mechanism and how things keep evolving at schema.org. I am still convinced that the medical vocabulary existing in schema.org should be improved and making it an extension, maintained by some people would help to make it robust and up-to-date.
Just looking at the proposal, some of suggested concepts have been already integrated in schema---there is a need to align the proposal to the existing schema version (eliminating duplicates, etc)

In parallel there was a big move towards having robust medical vocabulary and within the W3C Life science Group started the the HL7 ITS subgroup on RDF for Semantic Interoperability. The group is nowadays working among others on HL7 FHIR Ontology which my help/overlaps with the proposal.
Sure the group will not back the proposal but the existing of the extension  will be an added value.

The 2 questions I have here (for next steps):
-should we still use the same namespace schema: or we will use something else depending to the extension, eg. med.schema: ?
-Is there a mechanism (a developer's site ) to maintain the proposal or we need to that in an external API?

danbri:
@twamarc thanks for your reply, and your patience. Let's move ongoing discussion into the #492 issue.

jamrob:
Fully support!

MartinDesigner:
imghome
So, as a supplement, however as a main point try this amazing free app. Createer.com is an online software for easily designing online ads, responsive landing pages and print materials such as flyers, posters, covers, invitations and more. Createer lets you easily share your designs on social media, send them via email, download as HTML, image or PDF. Createer also includes 3,000 predesigned templates (browse here: https://beta2.createer.com/Templates?ghi) and over 250,000 images (illustrations, photos and icons) to work with in its user-friendly drag’n’drop enviroment. And finally this amazing stuff is here: https://beta2.createer.com/?ghi

danbri:
I made a new PR https://github.com/schemaorg/schemaorg/pull/1056 targetting sdo-deimos, and have (with some manual tweaks) just merged it into that branch for review.

Dataliberate:
This needs some review.

From a quick look, some core properties have been wrongly pulled into the extension. For example 'status' and 'purpose'  They may be modified in the extension (eg. extending the domain and/or range) but their original definition belongs in the core.

danbri:
Yes, review is what it needs. There is a very rough first cut up at http://health-lifesci.webschemas.org/ but note that I've not yet made the general front page text, so it just gives a *big* ugly list of terms from the extension.  However you can also navigate in from http://webschemas.org/docs/meddocs.html 

(btw think of webschemas.org as a volatile work-in-progress upstream test version, in sync with the latest proposed changes committed to our next release branch)

twamarc:
Good progress! Thanks 
I will review again to fix those core properties  inadequately pulled into the extension.

egonw:
MedicalScholarlyArticle seems needless. I understand a wish to introduce the pubType field, but the terms listed by MeSH are generic and apply to any ScholarlyArticle. I would prefer to see it added then to that type.

danbri:
@egonw for better or worse, we have had http://schema.org/MedicalScholarlyArticle since http://blog.schema.org/2012/06/health-and-medical-vocabulary-for.html in 2012. The bulk of the work in this set of changes is to move the super-detailed medical and health vocabulary that we already have into a dedicated health-lifesci extension, where other related efforts can also live.

danbri:
@twamarc if you have further change suggestions it may be best to outline them here textually, rather than via Pull Request. If you do make a pull request, it should be based on the current state of the sdo-deimos branch here to avoid complexity.

danbri:
(I'm going to close the pull request since it has effectively been merged, but we can continue to discuss here)

twamarc:
@danbri : Thanks. Maybe we can completery close this thread and follow the comments, improvements at 
* https://github.com/schemaorg/schemaorg/issues/492
on start another one, what do you think?

danbri:
Sounds good, #492 it is.

twamarc:
Agree! 
@All: Please do not make comments on this issue anymore. We want to centralize all threads and we would invite you to make all your comments/suggestions/discussion/edits at :
https://github.com/schemaorg/schemaorg/issues/492

Thanks
Marc.

PR #106, opened 2014-08-27
Proposal for OrderItem and parcelDelivery addition to the Orders Schema

dunglas:
This update has been discussed first [on the public vocabs mailing list](http://lists.w3.org/Archives/Public/public-vocabs/2014Jan/0161.html).

# Overview (extracted from the mailing list)

The current Orders Schema has some drawbacks:
- Many merchants such as Amazon, CDiscount or La Redoute handle status at the order line level: one item can be "Processing" while another of the same order (ordered in the same time) is "In Transit". In the current Orders Schema, the status is stored at the Order level, not at the Order item level, so it's impossible to represent case explained above.
- In the current Orders Schema, there is no notion of quantity for an order item. However, in almost all e-commerce platforms, a customer can order several times the same "Offer". (For instance, you can buy 3 t-shirts with SKU ABCDE).
- The Orders Schema does not include shipment details. However, a ParcelDelivery type already exists and a reverse property from Order to ParcelDelivery is sufficient.

# Vocabulary

The main additions / changes are:
- The new type Thing > Intangible > OrderItem
- Two new properties on Thing > Intangible > Order

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -3679,6 +3679,12 @@
       <span property="rdfs:comment">An order is a confirmation of a transaction (a receipt), which can contain multiple line items, each represented by an Offer that has been accepted by the customer.</span>
        <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Intangible">Intangible</a></span>
     </div>
+    
+    <div typeof="rdfs:Class" resource="http://schema.org/OrderItem">
+      <span class="h" property="rdfs:label">OrderItem</span>
+      <span property="rdfs:comment">An order item is a line of an order. It includes the quantity and details of a bought offer.</span>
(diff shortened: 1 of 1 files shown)


danbri:
Thanks for your patience and for the detailed proposals! We think we should do something around this. I'll fold it into a test build for public review.

dunglas:
:+1:

danbri:
@vholland I see this pull request still open, yet we already have OrderItem merged. Can we safely close this request now?

vholland:
Yes, I think so. I'm not sure what happened here, but the vocabulary is already there.

wuda-io:
Just wanted to point out, that the online version in schema.org differs from this approach. Check out OrderItem https://schema.org/OrderItem
So i like this one more.

PR #135, opened 2014-09-20
Create property "homepage" and add it to Person and Organization

lanthaler:
I'd like to propose to create a new property `homepage` with a domain of `Person` and `Organization` and a range of `WebPage`. There might be other types that could benefit from this property but let's keep it simple for the time being.

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -6431,6 +6431,13 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/ContactPoint">ContactPoint</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Place">Place</a></span>
     </div>
+    <div typeof="rdf:Property" resource="http://schema.org/homepage">
+      <span class="h" property="rdfs:label">homepage</span>
+      <span property="rdfs:comment">The homepage.</span>
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Person">Person</a></span>
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Organization">Organization</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/WebPage">WebPage</a></span>
+    </div>
     <div typeof="rdf:Property" resource="http://schema.org/honorificPrefix">
       <span class="h" property="rdfs:label">honorificPrefix</span>
       <span property="rdfs:comment">An honorific prefix preceding a Person&#39;s name such as Dr/Mrs/Mr.</span>


danbri:
I'm still not seeing a clear case that distinguishes a homepage from an url. Do you have a specific definition in mind that could clarify this?

thadguidry:
@danbri then if you have url ... what happens if a person has many urls to represent us, but we PREFER only 1 of them, how would you designate that preference that some Persons or Organizations have ?

Just answer this @danbri :  How does an organization capture their Web Presence preference for only 1 particular url , but still listing all 5 that they have ?  Would that be by using "url" and then a bunch of "sameAs" ?  If yes, then we just need to give an example and then change the definition of  "url" to say something like "the official url of this item" ... it appears that "sameAs" actually has the term "official website" in it's definition...that makes it hard to understand when should "sameAs" or "url" be used for capturing the preference.

The "homepage" idea is about Preference and designating that somehow I would say.

danbri:
I think at this point we have quite a collection of related properties (url, mainEntityOfPage, sameAs at least), so we're unlikely to expand to 'homepage' too. I'll close out this issue. If the issue comes back to life (and generally each pull request should be wrapped in a "what problem does this solve" issue) we can always dig this from the archives. Thanks for the proposal!

dsimmons:
Apologies, I don't mean to bring this back from the grave, but I too (like @lanthaler) desired this property.

Here's an example use case. Say I was modeling a user's Medium profile.

```JSON
{
    "@type": "Person",
    "name": "Ev Williams",
    "url": "https://medium.com/@ev",
    "sameAs": [
      "http://evhead.com/",
      "https://twitter.com/ev",
      "https://facebook.com/500085339"
    ]
}
```

I've purposefully omitted other properties in the intereste of brevity.

Here, I want to treat `http://evhead.com/` different from the other `sameAs` links and call out explicitly that it's a domain owned by that entity (Person, in this case). That's where I feel like `website` (or whatever the name) could come in.

You could imagine a similar scenario for a Wikipedia page, for instance. Really, from my perspective, it's about differentiating between *one* representation of the entity (which may or may not be owned by that entity) and the "home" for that entity (which is owned by the very same entity, such as a blog or a website). Put another way, I don't want to lump the "website" in with other `sameAs` links, as I may want to treat it fundamentally different.

Probably up for debate is whether the usage of `url` here is correct. According to the website:

> URL of the item.

Accordingly, I would expect `url` to be reserved for the URL of the entity being viewed. If it's the case that `url` should ***actually*** be `http://evhead.com/` when used properly, I'd say that `url` feels to be a bit of a misnomer.

If I'm way off, feel free to say so! Even having read https://schema.org/docs/datamodel.html, `url` and/or `@id` vs `sameAs` vs `mainEntityOfPage` is confusing to say the least! :confused:

betehess:
I didn't know about this issue. Right now, engines like Google consider that the homepage must have the path `/`, [expecting to find markup specifically *on* that webpage](https://developers.google.com/structured-data/customize/overview). But we at Apple have several homepages, for different countries/languages. We would certainly like a way to say that explicitly in our markup.

And maybe that way, Google's search box and other properties could be better localized too.

betehess:
I see the value in a property, but also in a proper type e.g. `schema:Homepage`.

vholland:
A region- and language-specific homepage designation would be useful.

PR #185, opened 2014-11-23
Remove additional itemprop/property attribute on 'Book' example

unor:
A Microdata/RDFa example had two `itemprop`/`property` elements on the same element, one of which contained an invalid property value (`Fiction`).

I guess it was intended to use `content="Fiction"` instead, but note that Microdata doesn’t allow this on `span` (see https://github.com/rvguha/schemaorg/issues/184) and the JSON-LD example uses "fictional" instead of "Fiction", too.


danbri:
Thanks! Merged in via https://github.com/schemaorg/schemaorg/pull/272 (I pulled it into sdo-stantz, our upcoming release, rather than master branch).

PR #190, opened 2014-12-09
Add "name" properties for CreativeWork examples

dbs:
As pointed out by Martin Hepp, several of the types in the CreativeWork
examples are missing "name" properties, resulting in empty, typed
entities without any sort of attached property. That's worse than a
blank node; call it a void node, maybe ;)


danbri:
Merged via https://github.com/schemaorg/schemaorg/pull/278 - thanks again, Dan!

Issue #53, opened 2014-06-21
Linter warnings on schema.org examples


gkellogg:
Update: https://gist.github.com/danbri/c5fef76dcf89bc23bdb6 has a fresh run against 2015-01-28th snapshot of sdo-stantz branch. -- @danbri

FYI, the following are warnings produced by the Structured-Data Linter for current schema.org examples:

[ earlier results removed for readability, see link above ] 

This was done through the following:
- clone http://github.com/structured-data/linter
- bundle install
- run the schema_examples rake task to generate all example files
- run `script/parse schema.org/*.html --lint --quiet

Changes to the schema.org RDFS need to be synchronized by updating the vocabulary in the RDF.rb gem, typically done after each release.

danbri:
Ok @gkellogg I finally got the linter running locally. Do you have a script handy that unpacks and checks the schema.org examples?

scor:
@danbri side note: might be worth to run all the example through the new Google validator at https://developers.google.com/webmasters/structured-data/testing-tool/ - you might have a way to send all those examples in batch and get the warnings/errors back. Maybe you have already done so, in which case ignore this comment. Happy to create a new issue if not.

danbri:
Yeah, we should do that! Thanks for the nudge, it's already on my todo list...

gkellogg:
From the linter directory, you can run `rake schema:warnings` which will generate `etc/schema-warnings.txt` over the examples included in the linter. These examples can be updated using `rake schema:examples` which takes them from Guha's master branch, but could easily be modified to take them from somewhere else.

danbri:
Thanks Gregg! Note that Guha's master branch has graduated to an organization account within github now. Most Web links seem to redirect nicely, hopefully it will continue to work for you too.

-> https://github.com/schemaorg/schemaorg

There's still some admin/workflow setup to do but at least the basic shape of things is settling down.

danbri:
FYI the 'rake schema:examples' command wiped out the examples collection because sinatra-linkeddata failed (I think it wasn't installed properly)

danbri:
It needed 

          gem install sinatra-linkeddata
          gem install active_support-custom_logger

... and then generates a load of (very useful) output, then appears to hang at schema.org/UpdateAction-163-jsonld.html, file is...  41665 Jan 19 22:14 etc/schema-warnings.txt

gkellogg:
Thanks, I'll update the dependencies and check out the UpdateAction issue.

[...]

PR #199, opened 2014-12-12
Example fixes

gkellogg:
I went through and fixed many examples, one of which was actually causing my RDFa processor to hang! Everything else should be non-controversial, but there are some issues that indicate properties might be misnamed.

There are many more markup issues, and it would be great if new examples passed some form of validation before being incorporated. These were identified using a command-line tool in the linter.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -9905,7 +9905,7 @@ JSON:
   "reservationStatus": "http://schema.org/ReservationConfirmed",
   "reservationFor": {
     "@type": "BusTrip",
-    "operatingCompany": {
+    "provider": {
       "@type": "Organization",
       "name": "Bolt NYC"
     },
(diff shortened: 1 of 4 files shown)


danbri:
Just to say a huge thanks for doing these. There's a bit of administrivia to do first before integrating but ... thanks - much appreciated :)

danbri:
Ok I've merged in https://github.com/schemaorg/schemaorg/pull/260 ... can we close this version? 

I made the %20 tweak afterwards.

Suggestions for better workflow welcomed - would you prefer I ask the original pull request to be reconfigured to another branch? in this case I wanted it against sdo-stantz rather than master branch, and it seemed easier just to make another quick pull request. I couldn't see a way to change this one myself.

scor:
You can change the target branch once the pull request is created, you have to create a new one, as inconvenient as it is...

What if you set the default branch and adjusted it accordingly after each release? https://help.github.com/articles/setting-the-default-branch/ - pull request will be created against that branch by default.

danbri:
Thanks @scor just what I was looking for! :)

danbri:
Related - I've updated https://github.com/schemaorg/schemaorg/issues/53 with a link to the results of re-running this linter against today's version of sdo-stantz: https://gist.github.com/danbri/c5fef76dcf89bc23bdb6

Issue #141, opened 2014-09-29
 AggregateOffer type triggers error with the Structured Data Testing Tool


Zinebb:
When trying to debug a case on the Google's FR webmasters forum, I noticed a problem in the schema.org documentation:

The second example in https://schema.org/Offer triggers errors when tested with the Structured Data Testing Tool: http://www.google.com/webmasters/tools/richsnippets?q=uploaded:80050431808bb1e0607ce88eb20d2a05

Is this a markup error or is it a bug in the testing tool?
Should AggregateOffer be added as a type that expects an 'offers' property?

Thanks,
Zineb

danbri:
Fixed in https://github.com/schemaorg/schemaorg/pull/209

eldk:
Hello,
Just test AggregateOffers with offers in https://developers.google.com/structured-data/testing-tool/ with example given in http://schema.org/AggregateOffer second and last exemple (copy/paste the code).

Property offers used with AggregateOffers is still throwing and error : "Nous ne reconnaissons pas la propriété offers pour un objet de type AggregateOffer." or we don't know offers property for a AggregateOffer object'

Is it fix on schema.org last released but not for google structured data testing tool ?

Greatings,

Eric

![google_structured_data_error_aggregate_offer_offers](https://cloud.githubusercontent.com/assets/11083657/6279940/e6dfce7a-b8a6-11e4-9df9-f9f7b3e2d52c.jpg)

PR #209, opened 2014-12-28
Fix AggregateOffer lossage

sesuncedu:
The example microdata is missing a </div> tag.

The domain of schema:offers is missing AggregateOffer (contra the examples); the rich snippets testing tool follows the rdfa and not the examples.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -935,14 +935,15 @@ MICRODATA:
     to <span itemprop="highPrice">$1495</span>
     from <span itemprop="offerCount">8</span> sellers
 
-  Sellers:
-  <div itemprop="offers" itemscope itemtype="http://schema.org/Offer">
-    <a itemprop="url" href="save-a-lot-monitors.com/dell-30.html">
-     Save A Lot Monitors - $1250</a>
-  </div>
-  <div itemprop="offers" itemscope itemtype="http://schema.org/Offer">
-    <a itemprop="url" href="jondoe-gadgets.com/dell-30.html">
-     Jon Doe's Gadgets - $1350</a>
+    Sellers:
+    <div itemprop="offers" itemscope itemtype="http://schema.org/Offer">
+        <a itemprop="url" href="save-a-lot-monitors.com/dell-30.html">
+        Save A Lot Monitors - $1250</a>
+    </div>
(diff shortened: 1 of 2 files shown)


danbri:
Thanks. Regarding domain, we do point to a parent type. I can't offer a formal proof that it's therefore ok to apply to a subtype but it would seem plausible?

        <http://schema.org/offers> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/1999/02/22-rdf-syntax-ns#Property> .
        <http://schema.org/offers> <http://www.w3.org/2000/01/rdf-schema#label> "offers" .
        <http://schema.org/offers> <http://schema.org/domainIncludes> <http://schema.org/Product> .
        <http://schema.org/offers> <http://schema.org/domainIncludes> <http://schema.org/Event> .
        <http://schema.org/offers> <http://schema.org/domainIncludes> <http://schema.org/CreativeWork> .
        <http://schema.org/offers> <http://schema.org/rangeIncludes> <http://schema.org/Offer> .
        <http://schema.org/offers> <http://www.w3.org/2000/01/rdf-schema#comment> "An offer to provide this item&#x2014;for example, an offer to sell a product, rent the DVD of a movie, or give away tickets to an event." .

sesuncedu:
I don't think Offer is in the domain -  just the range?

danbri:
/me blames his new glasses

You're entirely right. Which is the best kind of right. We should have a branch here shortly for the next release, into which this'll get pulled. Thanks!

danbri:
When I stop to think I realise this is the same issue as https://github.com/schemaorg/schemaorg/issues/141

danbri:
@sesuncedu  would you mind retargetting these fixes at the sdo-stantz branch?

danbri:
Merged via https://github.com/schemaorg/schemaorg/pull/274

danbri:
fixes https://github.com/schemaorg/schemaorg/issues/141

eldk:
Hello,
This seems to be not fixed in google micro-data testing tool
as this is the same as https://github.com/schemaorg/schemaorg/issues/141, I haved put my comment on this conversation.
Greatings,
Eric

PR #229, opened 2015-01-18
Mark up MedicalScholarlyArticle abstract as description

dbs:
schema:description provides "a short description of an item", while an abstract
provides a short description of an article. Rather than leaving the abstract in
the MedicalScholarlyArticle example without any structure, use
schema:description to mark up the abstract as we do in the example for Article.


danbri:
Thanks Dan!

danbri:
Could you retarget this at sdo-stantz branch? (or I'm happy to make another PR that does)

danbri:
Merged via https://github.com/schemaorg/schemaorg/pull/273 - thanks Dan :)

Issue #216, opened 2015-01-06
Schema.org has no way to describe invoices/bills, add Invoice type + vocab


danbri:
See proposal from Vicki, http://lists.w3.org/Archives/Public/public-vocabs/2014Dec/0031.html

"Schema.org currently has http://schema.org/Offer for offering items for
sale/lease. It has http://schema.org/Order to model when a purchase has been made,
but there is no type for marking up Invoices requesting payment.

To that end, we are proposing adding a new type: schema.org/Invoice where
an invoice is defined as "a statement of the money due for goods or
services; a bill.""

danbri:
Status update 2015-01-22: @vholland is working on some examples which should help clarify the difference between an invoice/bill and an order.

danbri:
Implemented in https://github.com/schemaorg/schemaorg/pull/279

Staged at http://sdo-stantz.appspot.com/docs/releases.html#sdo-stantz e.g. http://sdo-stantz.appspot.com/Invoice

This should go out (and issue will be closed) as part of our next release - final comments and tweaks welcomed here or on the public-vocabs list.

thadguidry:
Hmm... I thought we were going to have a property to hold the NET D, i.e. "Net 60 days", or "Due upon Receipt", which is used by many small businesses and their automated invoicing templates that get generated by various small office software products out there.  Reference for NET D, http://en.wikipedia.org/wiki/Net_D

vholland:
There is a paymentDue for the due date, but that is not quite the same as NET D. "Due upon receipt" is the tricky one as, I am not sure the programmer solution of setting it to Net 0 is useful to others.

thadguidry:
@vholland @danbri  In UBL 2.1 , the NET D is (I think, if I recall correctly) encapsulated in the Payment Terms ...yeah, looks like it is....see the XLS reference at http://docs.oasis-open.org/ubl/os-UBL-2.1/UBL-2.1.html#T-INVOICE

danbri:
I think with @vholland 's most recent commits we're done for this release. Ok to close this issue against sdo-stantz? @tmarshbing  ?

PR #279, opened 2015-01-22
Sdo bills

danbri:
This pulls in @vholland 's Invoice/Bills work.

discussion: http://lists.w3.org/Archives/Public/public-vocabs/2015Jan/0100.html

issue: https://github.com/schemaorg/schemaorg/issues/216


danbri:
I have trouble reading what's going on here but it looks at first take as if the bills work is against too old a version of the schemas?

vholland:
I've cleaned up the branch to match sdo-stantz. It should auto merge now.

danbri:
This fixes https://github.com/schemaorg/schemaorg/issues/216

PR #283, opened 2015-01-25
Changed "materials" property to singular "material"

lazaruscorporation:
Changed "materials" property to singular "material", as per my proposed change to the data file


danbri:
Thanks!

Issue #145, opened 2014-10-07
serviceAudience vs. audience


elf-pavlik:
http://schema.org/audience
http://schema.org/serviceAudience

related email: http://lists.w3.org/Archives/Public/public-vocabs/2014Oct/0023.html

I would suggest removing serviceAudience and using generic audience. If some reasons exist for not doing so maybe at least defining serviceAudience as subPropertyOf audience?

danbri:
@elf-pavlik 's proposal seems sensible to me. @vholland - any thoughts?

vholland:
Sounds good to me.

elf-pavlik:
PR #285

danbri:
Thanks. Historically it is pretty rare for us to actually erase terms, so far we usually just mark as supersededBy. In this case maybe that is too conservative and best to avoid clutter. Any thoughts?

sesuncedu:
You could mark the obsolete terms in HTML with <s>. In the type rendering
you could group the obsoleted terms together with the term that obsoleted
them.

I've been treating superseded as implying sub property and obsolete .

The sub property relationship might be wrong, if a superseding property can
have a narrower meaning than the older term.

elf-pavlik:
@danbri if you like I can revert | drop this change and then use *schema:supersededBy* instead!

danbri:
Yes, I think schema:supersededBy would be most appropriate here. @elf-pavlik could you tweak your branch that #285 draws from? I've filed this for the next release, and can't see it proving controversial.

Simon is right that we could say more about what it means. We have never said that it implies subPropertyOf, since it is easy enough to assert that independently at the same time where needed.

In this case I believe serviceAudience would also be a sub-property of audience.

danbri:
I'm going to take the pull request, then add back in 

-    <div typeof="rdf:Property" resource="http://schema.org/serviceAudience">		
-      <span class="h" property="rdfs:label">serviceAudience</span>		
-      <span property="rdfs:comment">The audience eligible for this service.</span>		
-      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Service">Service</a></span>		
-      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Audience">Audience</a></span>		
-    </div>

and use supersededBy instead of deletion.

danbri:
[...]

PR #285, opened 2015-01-25
removed serviceAudience and reused audience rel #145

elf-pavlik:
> The intended audience of the subject, i.e. the group for whom given item was created.

may need better wording


danbri:
I merged this into stantz by mistake. Have implemented by hand instead, and am using "An intended audience, i.e. a group for whom something was created." as definition.

PR #287, opened 2015-01-26
NOT FOR MERGING - debugging a problem w/ Vicki

danbri:


lazaruscorporation:
I'm not sure but I *think* that some of the above related to VisualArtwork will actually revert it to an earlier version than is currently in schemaorg:sdo-stantz

danbri:
Yes exactly this is super stale, I just made this PR as Vicki and I were discussing reworking these changes and it helped to show the diff. It will not be merged in.

PR #291, opened 2015-01-27
Add preferredLanguage property to Person.

vholland:


chaals:
Seems reasonable, modulo the fact that many people have more than one preferred langauges. I presume we are assuming they can add more than one without any problems, and if they want to go into details then they deserve the action item of figuring out how to specify those details, no?
(For example, in general my preferred language is english, except that in Spain it is usually spanish… and then there are some odd cases for other langauges).

vholland:
Yes. I thought about going into detail about location and medium (preferredVerbalLanguage and preferredWrittenLanguage), but the model started to get complicated quickly. I am assuming people can set more than one language and if people start to need more details, we can add them later.

chaals:
+ reply@
- notifications@

OK. Works for Me. I support adding this.

28.01.2015, 17:24, "vholland" <notifications@github.com>:
> Yes. I thought about going into detail about location and medium (preferredVerbalLanguage and preferredWrittenLanguage), but the model started to get complicated quickly. I am assuming people can set more than one language and if people start to need more details, we can add them later.
>
> —
> Reply to this email directly or view it on GitHub.

--
Charles McCathie Nevile - web standards - CTO Office, Yandex
chaals@yandex-team.ru - - - Find more at http://yandex.com

thadguidry:
+1 for preferredLanguage

We have similar in Freebase, called "Languages", but slightly different context, "Languages that this person can communicate fluently in."

Also, for Movies and Country, Freebase has "languagesSpoken" to hold all the languages spoken in them.

danbri:
+1 on adding this. Inkel's old [speaks-reads-writes vocabulary](http://lists.foaf-project.org/pipermail/foaf-dev/2003-March/005191.html) was more complex but even that didn't meet everyone's needs. Having preferredLanguage is far better than nothing.

danbri:
Premature/accidental commit to stantz reverted, though seems like a good idea to do this (Q1? next release?). Any thoughts @pmika @tmarshbing  ?

sesuncedu:
One thing to note is that the definition class Language is a bit too broad
a range,

Language does not have any subclasses, but is defined as a union of two
(almost disjoint) classes.

The definition is:
"Natural languages such as Spanish, Tamil, Hindi, English, etc. and
programming languages such as Scheme and Lisp."

I prefer English and Common Lisp ; lisp-common is not BCP 47 approved
(Larry Masinter can't be everywhere).

There probably ought to be  subclasses; the version for natural language
should have a property for a BCP 47 language tag.

There is also a case to be made for having required languages be usable on
job postings (and a case for having them be interpreted as BCP language
ranges).

If skills (skill? skillz? ) had a range of Skill (or Capability) , the
modality issue, and competence level could be handled in language
capability subclasses -    Farsi, DLPT 3, etc.

tmarshbing:
I also support adding this, but I would like to see a little more discussion before we do. How about targeting sdo-gozer for this change?

Based on the comments from @thadguidry on Freebase, I would also like to see some language property on Country. Would preferredLanguage be the right property there, or should we create one or more additional properties for Country, such as officialLanguage and/or the Freebase name languageSpoken?

I was hoping we could have consistent naming across Person, Movie, and Country, but I don't see how to make that work. Particularly when Movie already has inLanguage from CreativeWork.

inkel:
First of all: thanks @danbri for remind me of that time (I can't believe that email address, why wasn't I using a personal one instead of the one at work?)

As you've said, that vocabulary was a bit more complex than this proposal, even though it was simple enough. I recall the discussions on the mailing list, language specification isn't an easy task. @chaals pointed out something very true: when in Spain his/mine preferred language is Spanish, but in some other countries we might rather speak in English instead. And let's not even start with programming languages as @sesuncedu suggested :smile_cat: 

I think `preferredLanguage` is better than nothing, but I also think that it falls short on its premise, even when more than one could be indicated. Again, is not an easy topic.

Last but not least, I wouldn't go into a more detailed description like `preferredVerbalLanguage` or `preferredWrittenLanguage`, I'd start with something simpler.

rvguha:
Something like this would be very useful, but I agree that it needs more
discussion. I'd like to see a set of examples and use cases before we can
decide.

guha

danbri:
Three points: repetition of the property, better example of using external authority lists, and permissible shortcut representations:

Can a single entity have multiple different preferredLanguage properties?

Two arguments in favour of multiplicity:

* The Language type currently covers both programming and human languages (and perhaps hybrid and artificial languages - Klingon, OWL etc.)
* Many people are bi-lingually competent, and it seems better to simply allow repetition than to make it a formal inconsistency to say that someone prefers e.g. both English and yet also Spanish (to other languages).

We currently only have one rather weak example of the Language type on the site.

http://schema.org/Language using http://schema.org/availableLanguage and defining the specific language via a 'name' property on an instance of Language. I suggest adding at least the use of (1.) a Wikipedia URL (2.) a code from http://tools.ietf.org/html/bcp47 ("Tags for Identifying Languages"). What properties sould relate a Language instance to a wikipedia URL or a BCP47 string? sameAs? name?

Finally, do we want to say anything about shortcut idioms such as  "preferredLanguage": "en-GB". Generally schema.org says it tolerates these but doesn't usually anticipate them formally via rangeIncludes. Should we say that if used in string form, preferredLanguage takes bcp-47 codes? what about the name property applied to an instance of Language?

danbri:
looks like this was prematurely merged and everyone thought it was a good idea but asked for more discussion. 

Vicki's draft:

    <div typeof="rdf:Property" resource="http://schema.org/preferredLanguage">
       <span class="h" property="rdfs:label">preferredLanguage</span>
       <span property="rdfs:comment">The language the person prefers to use when communicating.</span>
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Person">Person</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Language">Language</a></span> 
    </div>

danbri:
Proposals in pull requests sometimes get overlooked - I've made a tracking issue #1084 to make sure we catch this one.

PR #293, opened 2015-01-28
VisualArtwork - added URL as type for 3 properties

lazaruscorporation:
Added URL as an expected type (in addition to Text) to the VisualArtwork properties of "artform", "surface", and "material" to allow use of linked data, as suggested by Niklas Lindström at https://lists.w3.org/Archives/Public/public-vocabs/2014Dec/0115.html


danbri:
Thanks! Nearly missed this one, @tmarshbing caught that slip.

Issue #222, opened 2015-01-12
'producer' and 'Organization'


unor:
http://schema.org/producer

The description of the [`producer` property](http://schema.org/producer) currently is:

> The person or organization who produced the work (e.g. music album, movie, tv/radio series etc.).

Problems:

* It says "person or organization", but it lists only `Person` under "Values expected to be one of these types", not `Organization`.

* In addition to `CreativeWork `, it lists `Organization` under "Used on these types". But can an organization have a producer?

Maybe `Organization` should be moved from "Used on these types" to "Values expected to be one of these types"?

thadguidry:
The person or organization who produced the CreativeWork.

Inverse on Person would be workProduced - "The work produced by this Person".
Inverse on Organization would be workProduced - "The work produced by this Organization".

IMHO.

vholland:
I see. There is a copy/paste error where Organization is the domain instead of the range.

danbri:
Fixed, and queued up for sdo-stantz release - http://sdo-stantz.appspot.com/producer

PR #299, opened 2015-01-28
Issue #222: Fixed copy/paste error on producer property where Organizati...

vholland:
...on was the domain instead of the range.


danbri:
This seems to be already fixed in sdo-stantz by #222, 

* verifiable at http://sdo-stantz.appspot.com/producer
* src: https://github.com/schemaorg/schemaorg/blob/sdo-stantz/data/schema.rdfa#L7904
* commit: https://github.com/schemaorg/schemaorg/commit/bccb6134d85bd3718d661302c471fb5dadd97169

I don't understand why the 'files changed' tab of the pull request shows the old version though - i.e. https://github.com/vholland/schemaorg/commit/a201f136b23acf4d38f3990f8208a441ce2fb1db - I would have expected the PR to track subsequent changes to the target repo. Should I RTFM?

Issue #341, opened 2015-02-17
Range of programmingLanguage: Language instead of Thing?


unor:
The [`programmingLanguage` property](http://schema.org/programmingLanguage) currently has a range of `Thing`.

Its definition ("The computer programming language.") suggests that it could/should be [`Language`](http://schema.org/Language) instead:

> Natural languages such as Spanish, Tamil, Hindi, English, etc. **and programming languages** such as Scheme and Lisp.

danbri:
Thanks. Yes this could do with more care (as could 'Language' itself). We don't want to include lists of actual languages (human, computer or hybrid) as part of the schema.org core, but we should document ways of using external enumerations e.g. Wikipedia/Wikidata and it might make sense to break out subtypes of Language.

danbri:
While there are other issues here (Code has a problematic name, see also unrelated 'code' property), writing Language rather than Thing seems a harmless improvement. I'll do that now and close out this specific issue. Thanks again!

danbri:
Done - http://sdo-gozer.appspot.com/programmingLanguage

PR #342, opened 2015-02-17
Range of programmingLanguage: Language instead of Thing

unor:
Suggested in #341


unor:
The change was made in https://github.com/schemaorg/schemaorg/commit/526abaf1be04fc069a9a67760f995cf787204767

Issue #367, opened 2015-03-01
Label of arrivalAirport is departureAirport


gooodguy:
I found this inconsistency:
```xml
    <div typeof="rdf:Property" resource="http://schema.org/arrivalAirport">
      <span class="h" property="rdfs:label">departureAirport</span>...
    </div>
```

PR #370, opened 2015-03-02
Wrong label for arrivalAirport

unor:
Fixes https://github.com/schemaorg/schemaorg/issues/367


chaals:
Looks good. Although labels could afford to be human text (generally - this case is consistent with a lot of other stuff)

Issue #263, opened 2015-01-22
Add a generic Property-Values mechanism for long-tail use (e.g. e-commerce, EXIF)


danbri:
Martin Hepp has a proposal: http://lists.w3.org/Archives/Public/public-vocabs/2014Dec/0057.html

danbri:
Used in https://github.com/schemaorg/schemaorg/issues/262 for Automobiles.

danbri:
I've asked Martin to break out the property-value piece from his larger proposal (currently all in one branch).

mfhepp:
I just created an individual pull request for the property-values contribution for better tracking, see here:

https://github.com/schemaorg/schemaorg/pull/377

As for the comments raised by Tom Marsh in https://lists.w3.org/Archives/Public/public-vocabs/2015Jan/0004.html:

"I am supportive of adding the proposed property-value and EXIF changes into schema.org, but I would like to see them separated out from the other changes so we can approve and incorporate them independently and so that there is a clearer change history for people to follow in GitHub."

This is implemented with this pull request.

"Assuming we make this change, however, I think it is essential that we provide guidance on when it is acceptable to use these constructs. In particular, if publishers start to use property-value pairs where there are equivalent schematized properties, it will significantly dilute the value of the vocabulary. Therefore, I think we need to document a requirement that the name in the property-value pairs cannot match a schema.org property (it would therefore be considered invalid markup if the name did match)."

I added a note to the additionalProperty property, stating

"Note: Do not use additionalProperty if there is a specific property for this characteristic readily defined in schema.org."

As for prohibiting a property-value pair with a name that exists in schema.org, I am not recommending that, because

a) publishers may have properties in their local databases that accidentally clash with schema.org names (e.g. from a table of 200 product features for a technical product). Catching those and implementing a specific handling can be a bit challenge for implementers.

b) the local database schemas might use names for properties with a different meaning (e.g. weight for the package weight).

It is clear that we should encourage publishers to use specific properties when possible and that the mechanism must not be used to dilute the core vocabulary. I think the current proposal strikes a balance.
We might want to clarify this in a blogpost or implementation notes for this feature.

"I think we should also have an informal agreement as a community that we will make additions to the vocabulary for any properties that turn out to be widely used in property-value pairs so that we can encourage more normalized and consistent representations."

I agree, but this is something that should be mentioned in a blogpost or implementation notes for this feature.

tmarshbing:
[...]

PR #377, opened 2015-03-11
Property values proposal as a separate pull request

mfhepp:
This is the property-values proposal turned into a separate contribution, as per

https://github.com/schemaorg/schemaorg/issues/263

It can be used for product properties, EXIF data, and will be useful for the separate automotive extension.

For the full background of the proposal, see https://www.w3.org/wiki/WebSchemas/PropertyValuePairs.

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -9180,6 +9180,7 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Number">Number</a></span>
 	  <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
 	  <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Boolean">Boolean</a></span>
+	  <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/StructuredValue">StructuredValue</a></span>
     </div>	
     <div typeof="rdf:Property" resource="http://schema.org/valueAddedTaxIncluded">
       <span class="h" property="rdfs:label">valueAddedTaxIncluded</span>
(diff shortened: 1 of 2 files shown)


data/schema.rdfa
```diff
@@ -9167,10 +9174,13 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/value">
       <span class="h" property="rdfs:label">value</span>
-      <span property="rdfs:comment">The value of the product characteristic.</span>
+      <span property="rdfs:comment">The value of the quantitative value or property value node. For QuantitativeValue, the recommended type for values is &apos;Number&apos;. For PropertyValue, it can be &apos;Text;&apos;, &apos;Number&apos;, &apos;Boolean&apos;, or &apos;StructuredValue&apos;.</span>
```
tmarshbing: Did you mean to add StructuredValue to the range? It is mentioned in the comment, but not listed in the ranges.

data/schema.rdfa
```diff
@@ -11330,6 +11346,47 @@ postponing for 1.6.
...
+<div typeof="rdf:Property" resource="http://schema.org/additionalProperty">
+    <span class="h" property="rdfs:label">additionalProperty</span>
+    <span property="rdfs:comment">A property-value pair representing an additional characteristics of the entitity, e.g. a product feature or another characteristic for which there is no matching property in schema.org. &lt;br /&gt;&lt;br /&gt;
+	
+	Note: Do not use additionalProperty if there is a specific property for this characteristic readily defined in schema.org. 
+</span>
+    <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Product">Product</a></span>
+    <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Place">Place</a></span>
+    <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/QualitativeValue">QualitativeValue</a></span>
+    <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/QualitativeValue">QualitativeValue</a></span>
```
tmarshbing: Did you intend for this to be QuantitativeValue instead?

mfhepp: Oops, yes; thanks for the catch! Will commit a fix shortly.

data/schema.rdfa
mfhepp: Yes, thanks - forgot that one! Fix comes in a minute!

mfhepp:
I fixed the description for unitText as requested in the email dated Jan 9, too.

Note that unitText is really important - one one hand, we do want people to use proper UNCEFACT Common Codes when they can, because they are much more reliable and allow unit conversion etc. On the other hand, we want people to publish unit information as text when they cannot do better. Having two separate properties for this maintains backward compatibility with the original GoodRelations model, tools, and data, and reduces the task for a consuming client.

mfhepp:
This pull request is now ready for inclusion in sdo-gozer, following yesterday's discussions.

chaals:
And works for me...

danbri:
Git says I can't merge this automatically, but should be easy enough. Will finish merge and update gozer test site within the week...

mfhepp:
Maybe this is because after you merged the auto proposal, this is no longer from the same codebase.

I can try to update the pull request.

mfhepp:
Hi Dan:

I fixed this, the pull request can now be automatically merged:

https://github.com/schemaorg/schemaorg/pull/377

The cause, afaics, was that the property-values proposal and the automotive proposal added lines in the very same region and git assumed this to be potential conflicts, while they were just sequential additions.

Issue #262, opened 2015-01-22
Add vocabulary for describing vehicles / automobiles / automotive


danbri:
http://schema.org/Car
http://schema.org/Vehicle

A substantial and detailed proposal has been prepared by Martin Hepp, see associated pull request: https://github.com/schemaorg/schemaorg/pull/80

Announcement / discussion: http://lists.w3.org/Archives/Public/public-vocabs/2014Dec/0057.html

Original W3C wiki link https://www.w3.org/wiki/WebSchemas/Vehicles

See [2015-01-09 review](https://lists.w3.org/Archives/Public/public-vocabs/2015Jan/0002.html) from @tmarshbing which suggests amongst other things partitioning the proposal into smaller sections.

See also #263 for the property-value proposal, which supports the proposal.

mfhepp:
I have just created a pull request 

    https://github.com/schemaorg/schemaorg/pull/379

that is a trimmed-down version of the automotive part of the original proposal and addresses the comments received in the meantime.

mfhepp:
Note that we need https://github.com/schemaorg/schemaorg/issues/263 (Property-Values) for this proposal to be consistent, because the unitText property is referenced but not contained in this pull request.

danbri:
http://www.fileformat.info/info/unicode/char/1f697/automobile.svg

mfhepp:
Is the logo for the respective blogpost ;-) ?

danbri:
This is partially queued up for sdo-gozer and partially coming later as an extension. I'll therefore leave the issue open and update the milestone.

See http://sdo-gozer.appspot.com/docs/releases.html#automotive

danbri:
Q: I see "Typical unit code(s): C62" in several definitions but don't understand this. Searching around it seems to be in the original OWL ontology from @mfhepp too, so I don't think I introduced it in error.

mfhepp:
C62 is the correct un/cefact common code for "piece" or "no unit".

martin

---------------------------------------
martin hepp
www:  http://www.heppnetz.de/
email: mhepp@computer.org

danbri:
I thought you were up a mountain!

[...]

PR #379, opened 2015-03-13
Automotive

mfhepp:
Dear all:

This is the new automotive extension proposal, initially based on https://www.w3.org/wiki/WebSchemas/Vehicles. This pull request addresses the requests to 

2. separate the automotive part from the property-values proposal,
3. remove the bicycles, boats, etc. elements until there will be fully-fledged proposals for those domains and
4. various detailed comments from Tom March from Jan 9, 2015.

In the following, I summarize how these comments and requests have been addressed.

"In general, I think we should review and make changes to schema.org in smaller self-contained units. In this case, that would mean splitting out the changes into:
(1) the value-prop changes, which I will send a separate mail on.
(2) the auto changes
(3) the bicycle changes
(4) the boats changes"

--> OK, done.

"For the bicycle and boat changes, I think we should defer the changes for now. In both cases, I did a quick survey of boats/bicycles for sale and manufacturer sites, and I saw little overlap between the main attributes they used to describe the boats/bicycles vs. what is present in the Bicycle, Watercraft, Boat, MotorBoat, and SailingBoat proposals. I think we should revisit this as a separate change once there are more complete and detailed proposals for these types with their own use-cases, etc. (the use cases in the current proposal are all auto-specific)."

--> OK, done: removed Bicycle, Boat, Watercraft, Motorboat, Sailing Boat and removed/updated domain/range definitions.

"One small note on Bicycle, when we add it: why does Bicycle allow for “some bicycles have a small combustion or electric engine that assists with pedaling” when there is a separate MotorizedBicycle type for that?"

--> The idea was that, at least in Europe, there are MotorizedBicycles, which are essentially very light Motorcycles (their pedals are not typically used, except for starting the engine), and bikes with a very small support engine, like e-bikes. Since we removed all bikes parts for the moment, this issue is closed.

"Car vs. Van vs. Truck – Can this be modeled as body styles (now called bodyType) instead? The three example sites seem to do this, and it strikes me as cleaner from a vocabulary point of view, especially as companies continue to blur the lines between these (SUVs, crossovers, etc.)."

--> OK, done: Van and Truck have been removed. This will also make it easier for used car listing sites that may not have proper classification information in their DBs. Also note that Wikipedia has recently changed its definition for car to to "A car is a wheeled, self-powered motor vehicle used for transportation." I copied that. I also added an acknowledgment to Wikipedia to the GAO text, to give proper credit that some class and property definitions are inspired by Wikipedia.

So we now have the following subtypes of Vehicle:

MotorizedRoadVehicle
	BusOrCoach
	Car
	Motorcycle
	MotorizedBicycle

MotorizedRoadVehicle will be handy for non-cars, like trikes, and it also allows bundling all engine-related properties nicely.

"For EmissionStandardValue, TransmissionTypeValue, EngineTypeValue, and BodyTypeValue, there doesn’t seem to be any distinction in the usage vs. using Text, so why do we need these? If we were to have enumeration values for these, then having separate types would make more sense to me."

--> The initial idea was to to use / allow the use of URIs, like Freebase, DBPedia. 
DONE - I changed the range of all respective properties to Text OR QualitativeValue or URL and removed the classes for these types. That should cover all use-cases and reduce the size of the proposal.

"For FuelTypeValue, I saw on the US version of Autotrader (one of the examples cited on the wiki) values of Alternative, Diesel, Electric, Gasoline, and Hybrid. These are missing from the enumeration, and I think they should be added. Also, I haven’t seen a reference to TwoStrokeMixture (the only current enum value for this type) from the sorts of sites in the use case list before. Is that for motorcycles? Or if the entire purpose of this type is to add TwoStrokeMixture to the lists already available on DBPedia/Freebase/etc., then I would argue people can just enter text for that, and we can remove the FuelTypeValue in favor of having people directly use QualitativeValue."

--> DONE. I removed TwoStrokeMixture an the FuelType type,  and changed of all properties to Text or QualitativeValue or URL. This will allow standardization bodies to establish canonical URLs/URIs for important fuel types and foster the use of Wikipedia, DBPedia and Freebase IDs.

"For CarUsageType, the name and description don’t imply past usage to me. Should we rename this as something like PastCarUsageType instead? Otherwise, when I first read it, I was expecting this type to be used to describe the current use (i.e., available for rental now, not available for sale to a private owner but previously used as a rental car)."

--> Note that the property name is specialUsage, which I think fits. For the enumerations, I would like to keep CarUsageType, because we may want to reuse this for terms and conditions of rental offers or insurance contracts.
DONE: I changed the description of the values RentalUsage, DrivingSchoolUsage, TaxiUsage to prepare for such future usages but kept specialUsage for the property and CarUsageType for the enumerations.

"Can we rename ACRISSCode to acriss and VIN to vin? Other similar properties (isbn, gtin13, gtin14, etc.) follow this naming convention."
--> OK, DONE: Changed ACRISSCode to acrissCode and VIN to vin.

"Can we rename acceleration to accelerationTime? This would make the usage slightly clearer and also remove the issues in the notes about there not being unit codes for acceleration."

--> OK, DONE. 

"Can we rename colorInterior to interiorColor to be consistent with things like tongueWeight, seatingCapacity, modelDate, etc.?"

--> OK, DONE.

"How should a publisher specify the condition (new/used/etc.)? It seems like we could reuse the existing itemCondition property (with updated description) for this."

--> For the general condition, it will be itemCondition. For damages, there is the damages property.

"configurationName seems to be duplicating a bunch of other properties (transmission, enginePower, bodyType, etc.). If people are going to do this, IMO, they should just use the name property. There is a concept of trim level that the US version of AutoTrader uses, which I think is missing and would be useful. That would capture probably just the “ST 2.5” part of the example. For a Subaru Legacy, on the US AutoTrader site, for example, you’ll see values such as “2.5GT Limited”, “2.5i”, “3.6R Limited”, etc. "

--> I would like to keep that, because

1. the resulting strings may be very valuable for NLP processing (links to knowledge graphs);
2. I am thinking of an extension for modeling configurable products (cars and other), and for that I would like to have a generic name already; trim is very car-specific and does not even fit for other vehicles.
3. schema:name may include the full make and model, while configurationName will just contain strings referring to the car configuration. In European cars, this may be much more detailed than just trim information, because they are very highly configurable. The configuration string in there will be too long to be used as a short name.
4. Yandex initially requested this property, if I remember correctly.

"The engine properties (displacement, name, power, and type) should probably be separated out into their own type, especially if we want to model multiple engines per vehicle (as some of the descriptions suggest). Otherwise, it will not be explicit which property sets go together."

--> While we could add a type Engine and set the domain of all these properties to MotorizedRoadVehicle OR Engine, and define a property "engine" with a domain of MotorizedRoadVehicle and a range of Engine, I would like to keep that as a future extension, because most cars have only one engine, and the proposal is already pretty substantial.

"For enginePower, should we also list the unit code for horsepower (BHP?)?"

--> I decided against listing it, because horsepower is to my knowledge an obsolete unit. As far as I can see, the proper code would be N12 ("Pferdestaerke", with 1 PS = 735,498 75 W, btw. It could still be used, but I think we should not actively promote it.

"For fuelConsumption and fuelEfficiency, it would be good to have examples of these in addition to the existing notes. Also, I am not sure it makes sense to put the “wrong” units in the unitCode property. Why not guide people to use unitText instead? I thought that unitText was introduced for just this situation; i.e., when there is no standard code to refer to."

--> I fixed the description to recommend unitText. As for examples, I will put them on my radar, but I think such can be added in a second step.

"fuelType is on Vehicle, but fuelConsumption and fuelEfficiency are only on the more specific types. I think the placement of these three properties should be consistent, and I like putting them on the more specific types."

-> OK, DONE: I changed the domain of fuelType to MotorizedRoadVehicle. 

"For gears, the description says forward and reverse gears, but I think it is customary to only list the forward gears. The example for Car also seems to list only the forward gears. Should we update the description?"

--> Agreed. OK - DONE.

"For interiorType, if this is intended to describe just the material, should we rename it to interiorMaterial?"

--> I would like to keep interiorType, because it could also be a type that is not based on a material but on a target usage or audience ("Hunter edition with gun rack" or "1970s disco interior").

"It looks like numberOfOwners was renamed to previousOwners. It would be nice to have the wiki updated if so."

--> OK, DONE - but while I tried to sync the Wiki page with the github work, it is very resource-consuming and error-prone, so we should consider the Wiki page deprecated.

"What is the use case for productionDate? I’m not sure we need this property, but if we do, can we reuse the existing dateCreated property instead? It sounds like it would have the same meaning."

--> In Europe, the production date is often an essential characteristic for new and used car, because some cars stand on stock for several months depending on factory capacities and demand. The property was also present in the older Yandex proposal. So I suggest to keep that.

"Can we rename purchasingDate to purchaseDate and firstRegistration to firstRegisteredDate? Our date property naming in schema.org isn’t very self-consistent, but I think these suggestions will be more consistent."

--> OK, DONE: Renamed purchasingDate to purchaseDate and firstRegistration to dateOfFirstRegistration, which I think is better English.

"transmission mentions sending power to propellers, but this property is on MotorizedRoadVehicle. Should it instead be on Vehicle? Or should the description be updated?" 

--> OK, DONE: I changed the description so that it remains generic without irritating typical users.

"There are some typos:
In the description for fuelCapacity: staorage => storage"

--> OK, DONE.

"In the description for unitText: “Useful if you cannot provide a standard unit code for” what?"

--> This is an issue for https://github.com/mfhepp/schemaorg/tree/property-values. I will add it there.

"In the descriptions for payload and weightTotal: permited -> permitted"

-> OK, DONE. (Also found a few more).

I also updated the examples accordingly.

Again, a huge thanks for your detailed review! I hope this is ready for inclusion now.

PS: For tracking what I changed between the two version, see

    https://github.com/mfhepp/schemaorg/commit/ed235c80edfd60000e0d1b6cb58f04284f734ec1

Best wishes

Martin

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -11332,7 +11332,6 @@ postponing for 1.6.
 </div>
 
 <!-- Vehicles extension proposal, as per https://www.w3.org/wiki/WebSchemas/Vehicles -->
-<div>
     <h1>Vehicles</h1>
     <div typeof="rdfs:Class" resource="http://schema.org/MotorizedRoadVehicle">
         <span class="h" property="rdfs:label">MotorizedRoadVehicle</span>
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -11686,8 +11686,8 @@ postponing for 1.6.
         <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
         <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#Automotive_Ontology_Working_Group">GAO</a></span>
     </div>
-    <div typeof="rdf:Property" resource="http://schema.org/configurationName">
-        <span class="h" property="rdfs:label">configurationName</span>
+    <div typeof="rdf:Property" resource="http://schema.org/configuration">
+        <span class="h" property="rdfs:label">configuration</span>
         <span property="rdfs:comment">A short text indicating the configuration of the vehicle, e.g. &apos;5dr hatchback ST 2.5 MT 225 hp&apos; or &apos;limited edition&apos;.</span>
         <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/MotorizedRoadVehicle">MotorizedRoadVehicle</a></span>
         <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -9369,11 +9369,12 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
       <span property="rdfs:label">BibExTerm</span>
       <span property="rdfs:comment">The W3C &lt;a href="http://www.w3.org/community/schemabibex/"&gt;Schema Bib Extend&lt;/a&gt; (BibEx) group led the work to improve schema.org for bibliographic information, including terms for periodicals, articles and multi-volume works. The design was inspired in places (e.g. pageStart, pageEnd, pagination) by the &lt;a href="http://bibliontology.com/"&gt;Bibliographic Ontology&lt;/a&gt;, 'bibo'.</span>
     </div>
-
+	
+	<div typeof="http://schema.org/Organization" resource="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#Automotive_Ontology_Working_Group">
+	  <span property="rdfs:label">AutomotiveOntologyWGClass</span>
+	  <span property="rdfs:comment">This class is based on the work of the Automotive Ontology Working Group, see &lt;a href=&quot;http://www.automotive-ontology.org&quot;&gt;http://www.automotive-ontology.org&lt;/a&gt; for details.</span>
+	</div>
     <hr />
-
-
-
 
     <h1>Reservations</h1>
 
@@ -9494,13 +9495,13 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
     </div>
     <div typeof="rdfs:Class" resource="http://schema.org/Vehicle">
       <span class="h" property="rdfs:label">Vehicle</span>
-      <span property="rdfs:comment">A vehicle.</span>
+      <span property="rdfs:comment">A vehicle is a device that is designed or used to transport people or cargo over land, water, air, or through space.</span>
       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Product">Product</a></span>
     </div>
     <div typeof="rdfs:Class" resource="http://schema.org/Car">
       <span class="h" property="rdfs:label">Car</span>
-      <span property="rdfs:comment">An automobile.</span>
-      <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Vehicle">Vehicle</a></span>
+      <span property="rdfs:comment">An automobile, motor car, or car is a wheeled motor vehicle used for transporting passengers, which also carries its own engine or motor.</span>      
+      <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/MotorizedRoadVehicle">MotorizedRoadVehicle</a></span>
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/reservationId">
       <span class="h" property="rdfs:label">reservationId</span>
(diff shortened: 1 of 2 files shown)


mfhepp:
This is https://github.com/schemaorg/schemaorg/issues/262 in the tracker.

tmarshbing:
"configurationName seems to be duplicating a bunch of other properties (transmission, enginePower, bodyType, etc.). If people are going to do this, IMO, they should just use the name property. There is a concept of trim level that the US version of AutoTrader uses, which I think is missing and would be useful. That would capture probably just the “ST 2.5” part of the example. For a Subaru Legacy, on the US AutoTrader site, for example, you’ll see values such as “2.5GT Limited”, “2.5i”, “3.6R Limited”, etc. "
--> I would like to keep that, because
1.	the resulting strings may be very valuable for NLP processing (links to knowledge graphs);
2.	I am thinking of an extension for modeling configurable products (cars and other), and for that I would like to have a generic name already; trim is very car-specific and does not even fit for other vehicles.
3.	schema:name may include the full make and model, while configurationName will just contain strings referring to the car configuration. In European cars, this may be much more detailed than just trim information, because they are very highly configurable. The configuration string in there will be too long to be used as a short name.
4.	Yandex initially requested this property, if I remember correctly.

[Tom] If we want to keep this more general, then how about just configuration? This opens the possibility in a future extension for us to create a new type to capture additional aspects of the configuration itself, which would also be useful when modeling what products can be purchased (vs. already existing ones as are covered by the current proposal).


"The engine properties (displacement, name, power, and type) should probably be separated out into their own type, especially if we want to model multiple engines per vehicle (as some of the descriptions suggest). Otherwise, it will not be explicit which property sets go together."
--> While we could add a type Engine and set the domain of all these properties to MotorizedRoadVehicle OR Engine, and define a property "engine" with a domain of MotorizedRoadVehicle and a range of Engine, I would like to keep that as a future extension, because most cars have only one engine, and the proposal is already pretty substantial.

[Tom] I still think we should create the Engine type and make the domain of all of these properties *only* Engine, not MotorizedRoadVehicle. It doesn’t make the proposal significantly larger (just one new type), and it prevents us from having to either keep two nearly-equivalent ways of saying the same thing (using the engine property or adding directly to MotorizedRoadVehicle) or deprecating the MotorizedRoadVehicle domain for these properties when we *do* support multiple engines in the future (which, while still uncommon, is becoming much more common in the US at least). 


"For enginePower, should we also list the unit code for horsepower (BHP?)?"
--> I decided against listing it, because horsepower is to my knowledge an obsolete unit. As far as I can see, the proper code would be N12 ("Pferdestaerke", with 1 PS = 735,498 75 W, btw. It could still be used, but I think we should not actively promote it.

[Tom] Horsepower is still the most commonly used unit for engine power in the US publications I have read. I checked www.caranddriver.com and www.roadandtrack.com for some examples, and they were all in horsepower.


"For interiorType, if this is intended to describe just the material, should we rename it to interiorMaterial?"
--> I would like to keep interiorType, because it could also be a type that is not based on a material but on a target usage or audience ("Hunter edition with gun rack" or "1970s disco interior").

[Tom] OK, but can you update the description to make that clearer?


"What is the use case for productionDate? I’m not sure we need this property, but if we do, can we reuse the existing dateCreated property instead? It sounds like it would have the same meaning."
--> In Europe, the production date is often an essential characteristic for new and used car, because some cars stand on stock for several months depending on factory capacities and demand. The property was also present in the older Yandex proposal. So I suggest to keep that.

[Tom] I still don’t understand the difference vs. dateCreated. I agree it’s important to know when the car was produced, but I think that’s the same as saying it’s important to know when the car was created.


"Can we rename purchasingDate to purchaseDate and firstRegistration to firstRegisteredDate? Our date property naming in schema.org isn’t very self-consistent, but I think these suggestions will be more consistent."
--> OK, DONE: Renamed purchasingDate to purchaseDate and firstRegistration to dateOfFirstRegistration, which I think is better English.

[Tom] If we are going to put date as a prefix, how about dateFirstRegistered, to parallel the existing dateCreated, datePublished, dateModified, etc. We can leave purchaseDate as is (it matches several other existing properties that end with “Date”).

mfhepp:
"configurationName seems to be duplicating a bunch of other properties (transmission, enginePower, bodyType, etc.). If people are going to do this, IMO, they should just use the name property. There is a concept of trim level that the US version of AutoTrader uses, which I think is missing and would be useful. That would capture probably just the “ST 2.5” part of the example. For a Subaru Legacy, on the US AutoTrader site, for example, you’ll see values such as “2.5GT Limited”, “2.5i”, “3.6R Limited”, etc. "
--> I would like to keep that, because
1. the resulting strings may be very valuable for NLP processing (links to knowledge graphs);
2. I am thinking of an extension for modeling configurable products (cars and other), and for that I would like to have a generic name already; trim is very car-specific and does not even fit for other vehicles.
3. schema:name may include the full make and model, while configurationName will just contain strings referring to the car configuration. In European cars, this may be much more detailed than just trim information, because they are very highly configurable. The configuration string in there will be too long to be used as a short name.
4. Yandex initially requested this property, if I remember correctly.

[Tom] If we want to keep this more general, then how about just configuration? This opens the possibility in a future extension for us to create a new type to capture additional aspects of the configuration itself, which would also be useful when modeling what products can be purchased (vs. already existing ones as are covered by the current proposal).

[mfhepp] - Okay, implemented in extension and  cross-checked examples (that do not use this so far). See https://github.com/mfhepp/schemaorg/commit/be853cba63c997042a446f68baedbd9cc0940624

mfhepp:
As for:
"What is the use case for productionDate? I’m not sure we need this property, but if we do, can we reuse the existing dateCreated property instead? It sounds like it would have the same meaning."
--> In Europe, the production date is often an essential characteristic for new and used car, because some cars stand on stock for several months depending on factory capacities and demand. The property was also present in the older Yandex proposal. So I suggest to keep that.

[Tom] I still don’t understand the difference vs. dateCreated. I agree it’s important to know when the car was produced, but I think that’s the same as saying it’s important to know when the car was created.

[mfhepp]: I would stick to productionDate, because dateCreated would be very uncommon terminology in the auto industry, same as if you used dateCreated for a person's birthday (which might be biologically incorrect anyway ;-).

mfhepp:
"For enginePower, should we also list the unit code for horsepower (BHP?)?"
--> I decided against listing it, because horsepower is to my knowledge an obsolete unit. As far as I can see, the proper code would be N12 ("Pferdestaerke", with 1 PS = 735,498 75 W, btw. It could still be used, but I think we should not actively promote it.

[Tom] Horsepower is still the most commonly used unit for engine power in the US publications I have read. I checked www.caranddriver.com and www.roadandtrack.com for some examples, and they were all in horsepower.

[mfhepp] ok, agreed. I added BHP and N12 as unit codes for brake horsepower and metric horsepower (PS). See https://github.com/mfhepp/schemaorg/commit/468335e38a2134571d42f2fb180c11300219a625

mfhepp:
"For interiorType, if this is intended to describe just the material, should we rename it to interiorMaterial?"
--> I would like to keep interiorType, because it could also be a type that is not based on a material but on a target usage or audience ("Hunter edition with gun rack" or "1970s disco interior").

[Tom] OK, but can you update the description to make that clearer?

[mfhepp] Done. See https://github.com/mfhepp/schemaorg/commit/37aa6b70ddd6c5d6020d126e4aac843d254bfde5

mfhepp:
"Can we rename purchasingDate to purchaseDate and firstRegistration to firstRegisteredDate? Our date property naming in schema.org isn’t very self-consistent, but I think these suggestions will be more consistent."
--> OK, DONE: Renamed purchasingDate to purchaseDate and firstRegistration to dateOfFirstRegistration, which I think is better English.

[Tom] If we are going to put date as a prefix, how about dateFirstRegistered, to parallel the existing dateCreated, datePublished, dateModified, etc. We can leave purchaseDate as is (it matches several other existing properties that end with “Date”).

[mfhepp] ok, done and cross-checked that no examples are affected. See https://github.com/mfhepp/schemaorg/commit/aa057f230a49fb90c5318f51bda9297b7680de0c

mfhepp:
Last item on the list ;-)
"The engine properties (displacement, name, power, and type) should probably be separated out into their own type, especially if we want to model multiple engines per vehicle (as some of the descriptions suggest). Otherwise, it will not be explicit which property sets go together."
--> While we could add a type Engine and set the domain of all these properties to MotorizedRoadVehicle OR Engine, and define a property "engine" with a domain of MotorizedRoadVehicle and a range of Engine, I would like to keep that as a future extension, because most cars have only one engine, and the proposal is already pretty substantial.

[Tom] I still think we should create the Engine type and make the domain of all of these properties only Engine, not MotorizedRoadVehicle. It doesn’t make the proposal significantly larger (just one new type), and it prevents us from having to either keep two nearly-equivalent ways of saying the same thing (using the engine property or adding directly to MotorizedRoadVehicle) or deprecating the MotorizedRoadVehicle domain for these properties when we do support multiple engines in the future (which, while still uncommon, is becoming much more common in the US at least).

[mfhepp] I implemented the proposed change and think it makes the model much cleaner. See  https://github.com/mfhepp/schemaorg/commit/3f4abba2ed25efd3ed12da475efa42e592083416 for details.

Basically what I did was:

1. Change engineName to engine with a range of EngineSpecification and a domain of vehicle
2. Move all engine properties to EngineSpecification (a few went to Vehicle or EngineSpecification to cater for simpler cases)
engineDisplacement
enginePower
engineType
torque
3. deleted MotorizedRoadVehicle + all references; mostly lifted to Vehicle, sometimes Car OR BusOrCoach.
4. Updated examples in all sytaxes.

The pull request is now ready to go from my perspective. Give that this last change is non-trivial, it might be a good idea to check that I did not introduce any new bugs ;-)


But again - this is basically ready for inclusion into sdo-gozer from my point of view.

tmarshbing:
The changes look good to me. Thanks for all the hard work, @mfhepp!

PR #381, opened 2015-03-15
New 'difficulty' property for Recipe

sergioredondo:
Many recipes websites label their recipes depending on the difficulty of elaboration, so this could be a new property for Recipe schema.

### DIFF
--- a/data/sdo-difficulty-examples.txt
+++ b/data/sdo-difficulty-examples.txt
@@ -1,0 +1,43 @@
+<div typeof="rdf:Property" resource="http://schema.org/difficulty">
+      <span class="h" property="rdfs:label">difficulty</span>
+      <span property="rdfs:comment">Difficulty of the recipe.</span>
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Recipe">Recipe</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
+</div>
+
+Example:
+
+<div vocab="http://schema.org/" typeof="Recipe">
+  <span property="name">Mom's World Famous Banana Bread</span>
+  By <span property="author">John Smith</span>,
+  <meta property="datePublished" content="2009-05-08">May 8, 2009
+  <img property="image" src="bananabread.jpg"
+    alt="Banana bread on a plate" />
+  <span property="description">This classic banana bread recipe comes
+  from my mom -- the walnuts add a nice texture and flavor to the banana
+  bread.</span>
+  Prep Time: <meta property="prepTime" content="PT15M">15 minutes
+  Cook time: <meta property="cookTime" content="PT1H">1 hour
+  Yield: <span property="recipeYield">1 loaf</span>
+  Difficulty: <span property="difficulty">Easy</span>
+  <div property="nutrition"
+     typeof="NutritionInformation">
+    Nutrition facts:
+    <span property="calories">240 calories</span>,
+    <span property="fatContent">9 grams fat</span>
+  </div>
+  Ingredients:
+  - <span property="ingredients">3 or 4 ripe bananas, smashed</span>
+  - <span property="ingredients">1 egg</span>
+  - <span property="ingredients">3/4 cup of sugar</span>
+  ...
+  Instructions:
+  <span property="recipeInstructions">
+  Preheat the oven to 350 degrees. Mix in the ingredients in a bowl. Add
+  the flour last. Pour the mixture into a loaf pan and bake for one hour.
+  </span>
+  140 comments:
+  <meta property="interactionCount" content="UserComments:140" />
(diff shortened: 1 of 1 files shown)


vholland:
The term 'difficulty' might apply to a variety of things. Can we rename to something like 'recipeDifficulty' to avoid name collisions in the future?

rvguha:
I agree, but the concept of 'how difficult these instructions are to
follow' applies to a number of things. Whilst 'difficulty' is too
ambiguous, 'recipeDifficulty' is too specific. Can we do something that
covers all 'howDifficultTheseInstructionsAreToFollow' with a better name
than that?

thanks,

guha

thadguidry:
For recipes... I typically just count(verbs)...to get a taskDifficulty score.  Seems to work pretty good for my usage.  verbs like (bake, cover, mix, stuff, mash, grind, season, fry, toss, etc...)

sergioredondo:
When I thought about this new property I meant 'Recipe Dificculty'. I imagine this new property with three possible values: Easy, Intermediate and Difficult. Of course, is up to the user who marks up the content to use whatever he/she wants, but in order to limit its use I think it's probably better to find a way to use only 3 possible values.

I like the option 'taskDifficulty', because it can be used not only for 'Recipes', but more types, so it can be a property for an upper level like 'CreativeWork' or 'Thing', although I prefer it being exclusive to 'Recipe'.

thadguidry:
I want taskDifficulty (or an even better term 'difficultyRating') to not be
limited to an enum of 3 values of: Easy, Intermediate, and DIfficult.

I want support for a https://schema.org/PropertyValueSpecification
that way, we do not limit others but provide

I think difficultyRating might be a better choice on 2nd thought...subtle
alignment with AggregateRating and such which can later on aggregate the
most difficult and least difficult, etc...

sergioredondo:
You're right Thad. I agree with you. 'difficultyRating' could be a great solution.

vholland:
I like 'difficultyRating'. Instructional sites are common. I can see eventually having a general TaskInstruction type that could use this property.

thadguidry:
Now we just need to decide...

Is a really great Recipe... one with a default difficultyRating of 1 or 5 ? :-)
I think the best recipes and styles are those around difficulty 2, like Jamie Oliver's 15 min quick recipes...a little effort and a lot of POW ! :-)

Seriously, we might want to look more at AggregateRating also... sorting out bestRating and worstRating.

Probably should not re-use those, but instead create new easiestRating and hardestRating properties underneath ?  Dunno.

Thoughts?

chaals:
Well, assuming 1 means "only for experts with a well-equipped culinary laboratory and team" and 5 means "your 2-year-old can do it"…

difficultyRating looks good to me.

sergioredondo:
I like 'difficulty Rating' option, but, against chaals's opinion, I imagine 1 as easy and 5 as difficult. But, thinking on snippets, the thing here is that this rate cannot be shown as stars by search engines, so I think is imperative to find another option to indicate this type of rating, thus distancing of 'aggregateRating'. That's why my initial thought of this rating was word-based, using some scale from Easy to Difficult (or from very easy to very difficult), so snippets could only shown a word, something like:
Difficulty: Easy
Don't know. What do you think about it?

vholland:
Perhaps we create a new schema.org/DifficultyRating similar to schema.org/Rating. The properties are something like:

easiestRating
mostDifficultRating
ratingValue

Someone looking to provide a single word could use ratingValue.

Someone looking to provide a rating on a 1-5 scale could tell us how to interpret 1 and 5.

thadguidry:
@sergioredondo  I agree that Stars might not be the best option to represent difficulty of tasks, recipes, etc.  The search engines can perhaps come up with a uniform symbol for this...let's let the industry choose that.

@vholland  Agreed on having ratingValue to hold text.
Agreed on schema.org/DifficultyRating (it will make publishers lives easier this way actually)
whatever you think is best for the property names, but I sorta prefer:

schema.org/DifficultyRating
- easiestRating
- hardestRating
- ratingValue

sergioredondo:
Ok, I think it could be the best option. 

Excuse me, but I'm a newbie here. What's the next step?

GraceMassaLanglois:
Just throwing this out, levelOfDifficulty - beginner, intermediate, advanced.

fatso83:
This effort started nicely, but it seems to have died out. Re @sergioredondo's question, what can be done to further this initiative?

thadguidry:
@sergioredondo Can you add an additional example file that uses the
numbered steps form (rather than paragraph) for the
property="recipeInstructions"  We'll need to support both
​forms ​
and teach users how to do it.
​  Real Example here:
http://allrecipes.com/recipe/241398/honey-balsamic-chicken/​

thadguidry:
Once we have the other file example show numbered instructions...

I am +1 for getting this into Pending.

Issue #109, opened 2014-08-28
StructuredValue description is confusing


danbri:
"If you are messing around near StructuredValue, it would probably be worthwhile to fix up the comment on StructuredValue, which says that structured values are strings, but all the subclasses of StructuredValue are not strings." -- http://lists.w3.org/Archives/Public/public-vocabs/2014Aug/0205.html

http://schema.org/StructuredValue "Structured values are strings—for example, addresses—that have certain constraints on their structure."

Subtypes: 

More specific Types
ContactPoint
GeoCoordinates
GeoShape
NutritionInformation
OpeningHoursSpecification
OwnershipInfo
PriceSpecification
QuantitativeValue
TypeAndQuantityNode
WarrantyPromise

danbri:
See also https://github.com/schemaorg/schemaorg/issues/394

mfhepp:
There is a pull request for this at https://github.com/schemaorg/schemaorg/pull/395

danbri:
Moving my comment in here from #395 pull request. It was:

""""Structured values are values that consist of sub-elements and thus have some kind of internal structure."

Thanks for making an exact suggestion. We haven't so far used the term "element" heavily; it sounds rather XML-oriented. 

Maybe something like "Structured values are used when the value of a property has more complex structure than simply being a textual value or a reference to another thing."?"""

/cc @vholland @rvguha

chaals:
WfM

reavels:
I'm still confused by this to be honest. 
I can understand the difference between simple and complex (structured) types, such that a simple type is string, datetime, number, etc which don't have properties and a structured type has properties. 
Obviously, this definition doesn't work in Schema.org because everything is inherited from Thing which has properties...but then how is a StructuredValue different to say Person as the value type of the owner property, or Event as the value type of the event property? 
...hopefully someone can clear up my cloudy view here...

thadguidry:
Its a Parent Type for holding ANY structured value.

16R22L12R

Like my old high school locker combination.
Instead of simply saying its a Text value.
I could instead say its a StructuredValue

Then proceed to describe the structure.

Name=LockCombination
Type = StructuredValue
Property=firstRotation
Value=16R
Property=secondRotation
Value=22L
Property=thirdRotation
Value=12R

This is a quick hack and not very applicable, but should help you
understand that StructuredValue lets you build structure for values...where
Schema.org might not have a Type for your Structure yet... Like my
LockCombination subtype above.

[...]

Issue #394, opened 2015-03-24
Better wording for StructuredValue


mfhepp:
The wording for http://schema.org/StructuredValue is no longer appropriate, since it refers to strings, while this type is now used as a supertype for all kinds of structured value entities.

danbri:
True, but duplicate of #109 - closing this in favour of the existing record.

PR #395, opened 2015-03-24
Fixed wording for StructuredValue (Issue 394)

mfhepp:
This fixes issue 394.
https://github.com/schemaorg/schemaorg/issues/394

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -689,7 +689,7 @@
 
     <div typeof="rdfs:Class" resource="http://schema.org/StructuredValue">
       <span class="h" property="rdfs:label">StructuredValue</span>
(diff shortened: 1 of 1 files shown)


mfhepp:
This also fixes issue #109

danbri:
"Structured values are values that consist of sub-elements and thus have some kind of internal structure."

We haven't so far used the term "element" heavily; it sounds rather XML-oriented. Maybe something like "Structured values are used when the value of a property has more complex structure than simply being a textual value or reference to another thing"?

mfhepp:
@danbri 
Hi Dan, 
I have seen that you fixed this locally, but I spotted two minor mistakes, which this pull request fixes.
Martin

danbri:
Thanks! The "a" was a stylistic thing rather than a typo, but I can't claim same for ".." :)

PR #400, opened 2015-03-25
Attribution for Car

mfhepp:
I had forgotten to add the attribution markup to the Car type, because it previously existed. But given that the majority of the new features are from the automotive extension and that it was just a skeleton before, I think that is justified.


mfhepp:
@danbri 
I just rebased this pull request after your recent changes in the sdo-gozer branch. Please try to pull this into sdo-gozer. Thanks!

PR #409, opened 2015-04-02
Change range for fileSize in SoftwareApplication

tilid:
In http://schema.org/SoftwareApplication property fileSize has range http://schema.org/Integer but text description contains example with unit of measurement (18MB) what mean that actually range of fileSize is not integer, it's text


danbri:
Thanks! Merged. We should clarify these text formats at some point but for now this is the best fix.

Issue #365, opened 2015-02-27
'award' property also for Organization and Product?


unor:
The [`award` property](https://schema.org/award) is defined for `CreativeWork` and `Person`:

> An award won by this person or for this creative work.

I think it would make sense to allow it also for [`Organization`](http://schema.org/Organization) ([use case](http://stackoverflow.com/q/28724723/1591669 "Stack Overflow: How do I use the “Award” Schema for an Organization, not a Person?")) and [`Product`](http://schema.org/Product).

danbri:
Sounds plausible to me. Any objections?

Dataliberate:
+1
~Richard

On 16 Mar 2015, at 14:22, Dan Brickley <notifications@github.com<mailto:notifications@github.com>> wrote:


Sounds plausible to me. Any objections?

thadguidry:
+1 for award property on Organization and Product

tmarshbing:
+1

vholland:
+1

rvguha:
Looks good, lets do it.

chaals:
(And a positive endorsement)

mfhepp:
Yes, +1

danbri:
Thanks @unor, and everyone. I've merged #365 into gozer.

http://sdo-gozer.appspot.com/award

PR #413, opened 2015-04-07
extended domain of award/awards (#365)

unor:
https://github.com/schemaorg/schemaorg/issues/365

Added `Organization` and `Product` as domain of `award`/`awards`, and adjusted the description.


unor:
I changed it also for the deprecated `awards` property. Was this correct?

I’m not sure which description to use:

* "An award won by this organization/person or for this product/creative work."

* "An award won by or for this item."

* …

Issue #414, opened 2015-04-08
Bad HTML Tag on RDFA


Downchuck:
The "awards" entry in schema rdfa file has a trailing end tag on the span for label.
https://github.com/schemaorg/schemaorg/blob/sdo-gozer/data/schema.rdfa#L5035

danbri:
Thanks, fixed by @unor

PR #415, opened 2015-04-08
HTML fix (#414)

unor:
Fixes https://github.com/schemaorg/schemaorg/issues/414


danbri:
Fixed (manually, but thanks for finding it and the pull request!)

Issue #232, opened 2015-01-19
2.4 children? We say Number where we mean Integer 


danbri:
http://schema.org/numAdults
http://schema.org/numChildren
http://schema.org/additionalNumberOfGuests
http://schema.org/Integer
http://schema.org/Number
http://schema.org/LodgingReservation 

We should review http://schema.org/Number "Instances of Number may appear as values for the following properties". 

Also suspicious: http://schema.org/additionalNumberOfGuests ...

Reported by Simon Spero @sesuncedu in http://lists.w3.org/Archives/Public/public-vocabs/2015Jan/0057.html

vholland:
I changed the range for:
- numAdults
- numChildren
- numberOfItems
- numberOfEpisodes
- numberOfSeasons
- partySize
- playersOnline
- ratingCount
- reviewCount

See pull request #420.

danbri:
Thanks :)

PR #420, opened 2015-04-10
Issue #232: Changed numberOfItems, numberOfEpisodes, numberOfSeasons, ra...

vholland:
...tingCount, reviewCount, partySize, playersOnline, numAdults and numChildren to use Integer instead of Number.


danbri:
Thanks!

PR #422, opened 2015-04-13
fix json-ld syntax error in BreadcrumbList example

ryanofsky-google:


danbri:
Thanks! This is already fixed in sdo-ganymede branch, which will be merged into master shortly so I'll close this now.

PR #442, opened 2015-04-22
Change equvalentClass for DataCatalog to dcat:Catalog

gkellogg:
... as dcat:DataCatalog does not exist.
@danbri can we get this into sdo-gozer?


danbri:
Thanks!

Issue #439, opened 2015-04-21
numberOfItems clarification


unor:
The current description of [`numberOfItems`](http://schema.org/numberOfItems):

> The number of items in an ItemList. Note that some descriptions might not full describe all items in a list (e.g. multi-page pagination).

I think the second sentence could be more clear; [a Stack Overflow user is also not sure about it](http://stackoverflow.com/q/29764039/1591669#comment47682178_29764039).

It is my understanding that `numberOfItems` should *always* specify the total number, even for paginated or teaser lists. Correct?

Example: The homepage contains a sidebar, "My favorite movies", teasing only 2 of 6 movies, with a link to the page with the full list. In both cases, `numberOfItems` should be `6`.

```html

  <!-- http://example.com/ -->

  <div itemscope itemtype="http://schema.org/ItemList">
      <meta itemprop="numberOfItems" content="6"/>
      <p itemprop="itemListElement">…</p>
      <p itemprop="itemListElement">…</p>
      <a itemprop="url" href="/favorite-movies">Full list</a>
  </div>

  <!-- http://example.com/favorite-movies -->

  <div itemscope itemtype="http://schema.org/ItemList">
      <link itemprop="url" href="/favorite-movies"/>
      <meta itemprop="numberOfItems" content="6"/>
      <p itemprop="itemListElement">…</p>
      <p itemprop="itemListElement">…</p>
      <p itemprop="itemListElement">…</p>
      <p itemprop="itemListElement">…</p>
      <p itemprop="itemListElement">…</p>
      <p itemprop="itemListElement">…</p>
  </div>

```

I guess it would be different if the `ItemList` items had different URLs, as we’d have multiple lists then, not the same represented differently. Correct?

danbri:
Thanks. We also managed to typo "full" for "fully".

Suggestion: 

* current: "The number of items in an ItemList. Note that some descriptions might not full describe all items in a list (e.g. multi-page pagination)."
* proposed: "The number of items in an ItemList. Note that some descriptions might not fully describe all items in a list (e.g. multi-page pagination); in such cases, the numberOfItems would be for the entire list."

halindrome:
Minor nit:  e.g. should always have a comma after it:

proposed: "The number of items in an ItemList. Note that some descriptions might not fully describe all items in a list (e.g., multi-page pagination); in such cases, the numberOfItems would be for the entire list."

unor:
[...]

PR #443, opened 2015-04-22
clarified numberOfItems description

unor:
Suggestion from @danbri (improved by @halindrome) for #439

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -6916,7 +6916,7 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
 
     <div typeof="rdf:Property" resource="http://schema.org/numberOfItems">
       <span class="h" property="rdfs:label">numberOfItems</span>
-      <span property="rdfs:comment">The number of items in an ItemList. Note that some descriptions might not full describe all items in a list (e.g. multi-page pagination).</span>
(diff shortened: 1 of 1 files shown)


data/schema.rdfa
```diff
@@ -6916,7 +6916,7 @@ Note that Event uses startDate/endDate instead of startTime/endTime, even when d
 
     <div typeof="rdf:Property" resource="http://schema.org/numberOfItems">
       <span class="h" property="rdfs:label">numberOfItems</span>
-      <span property="rdfs:comment">The number of items in an ItemList. Note that some descriptions might not full describe all items in a list (e.g. multi-page pagination).</span>
+      <span property="rdfs:comment">The number of items in an ItemList. Note that some descriptions might not fully describe all items in a list (e.g., multi-page pagination); in such cases, the numberOfItems would be for the entire list.</span>
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/ItemList">ItemList</a></span>
```
gskema: Perhaps `should be` or `should be of` would be clearer?

gskema: Perhaps `should be` or `should be of` would be clearer?

danbri: thanks for the thought, but I think 'would' works fine in this context.

PR #451, opened 2015-04-23
Bugfixes for GoodRelations texts

mfhepp:
This pull request adds missing line breaks to the lists of recommended values for GoodRelations properties. Would be nice to have this in sdo-gozer.


danbri:
Thanks:)

PR #454, opened 2015-04-27
Merge sdo scripts into gozer

danbri:
I have hand-merged the python tweaks in sdo-gozer into sdo-scripts, so this should be relatively ok.


danbri:
Merged manually.

PR #461, opened 2015-04-29
Merge bib extension 1.0 proposed types, properties, and examples

dbs:
This branch includes the complete history for the comics RDFa and examples, and also includes a commit with the core bib extension 1.0 RDFa and examples credited in git history to @Dataliberate (Richard Wallis).

This should merge smoothly into sdo-gozer.


danbri:
Ok I'll proceed to merge this. I think there are some things to consider and debate e.g. names, some parts for core etc., but we can handle that down-the-line. For now I want to have a consolidated candidate build so we can make sure that extensions are appropriately navigable.

I'll also later move data/bsdo-*-examples.txt down into data/ext/bib/ but I think they're harmless at the top level for now. Worst case (and it may be a positive feature) is that extension example "leak" into the core via examples being tagged with core and with extension term names. Rather than isolate all extension examples away, I think better to make some navigation / info alongside the examples to indicate that some terms are layered in from extensions.

/cc @tilid @tmarshbing @chaals @pmika @ajax-als @scor @mfhepp @rvguha

Issue #329, opened 2015-02-06
New schema to support TV listings


vholland:
Schema.org currently has types for BroadcastService and TelevisionStation, but there is no notion of a channel in a given TV line up as one would find from a satellite or cable provider.

This proposal is to add the following:

- New type TelevisionChannel for a specific channel in a provider lineup. Properties include:
    - channel: for the identifier (usually a number) in a provider's lineup
    - station: The TelevisionStation associated with the channel
    - inLineup: linking a channel to a BroadcastService (e.g. Comcast cable)
    - serviceTier: A text description of the service needed to access the channel (e.g. premium)
 
- New properties on TelevisionStation:
    - affiliateOf: The broadcast network(s) associated with a TelevisionStation.
    - displayName: The name the station uses on its display. For example, WAAY may display as "ABC".
    - timezone: The timezone the station uses when publishing its lineup.

Details and some examples can be found in my GitHub repository at https://github.com/vholland/schemaorg/commit/fe3fa6fea747f7080f1811a86bb35de8ac691989

tmarshbing:
For timezone, should we recommend that people use the ISO 8601 suffix (e.g., "−06:00" for Central Standard Time)?

For channel, should we name it more specifically, such as televisionChannel? The Wikipedia disambiguation for channel has a few other meanings that could conceivably end up in our vocabulary at some point, such as for geography, marketing, communication, etc.

Similarly, station may also need to be qualified.

We have properties from channel -> broadcast service, and channel -> station, but not reciprocal ones. Should we have the reciprocal ones, too?

vholland:
Agreed on the names. I'll think it through and come up with some better ones.

moustaki:
If the proposal is to describe such cable/satellite providers as BroadcastService (which I don't think was the original intent of that concept - the broadcast service examples include e.g. BBC Radio 4, Channel 4, NBC, etc.), then maybe the existing parentService and organization properties could be used to describe such relationships, without introducing new concepts.

In any case, I would recommend moving the proposed TelevisionChannel/TelevisionStation properties to concepts that could accommodate radio stations as well.

vholland:
I've [updated the property names](https://github.com/vholland/schemaorg/commit/adabe5d2d0b28f36d04424bcc439522aaf9a5a6c)  to be less generic and I hope in line with Radio where applicable.

**TelevisionStation**:
- "affiliateOf" changed to "broadcastAffiliateOf"
[...]

PR #464, opened 2015-04-30
Issue #417: Renamed types/properties per suggestions

vholland:


danbri:
The old vs_termStatus vocabulary we made up in the pub for FOAF had  this:
 
* 'unstable' - The meaning, deployment practices, documentation (or important associated software/services) associated with this term are liable to change arbitrarily at some point in the future. They may not, but stability is not guaranteed. Use with caution.
* 'testing' - The meaning, deployment practices, documentation and general understanding of this term are approaching some stability, but changes are still possible due to implementation experience or other unanticipated factors.
* 'stable' The term is relatively stable, and its documentation and meaning are not expected to change substantially.
* 'archaic' This term is marked as old-fashioned; although used, it is not considered typical of current best practice and alternative expressions may be preferable.

We preferred to say "archaic" instead of "deprecated", but maybe that's too cutesy? The point was to not make people feel bad for having adopted against earlier versions of the spec, and to acknowledge that the term would continue to mean whatever it was that it meant earlier rather than "dissapeared" completely. Non-spec at http://www.w3.org/2003/06/sw-vocab-status/note

If these definitions are a reasonable start, I can reflect them into a schema.org enumeration trivially. I'd prefer discussion here to confirm first.

danbri:
I've moved @chaals point about (broadcast)Timezone across to the original thread where this was discussed back in feb/march, https://github.com/schemaorg/schemaorg/issues/329#issuecomment-101326224

danbri:
also https://github.com/schemaorg/schemaorg/issues/481

Issue #300, opened 2015-01-28
General chaos around questions/answers in ReplyAction, AskAction vs Question, Answer, question property vs lack of answer property.


vholland:
http://schema.org/ReplyAction
http://schema.org/Answer

Original issue: I was looking at the lint errors outlined in issue #53. At some point in the proposal process, ReplyAction had an "answer" property, which was removed prior to official adoption. It would be useful to link ReplyAction to schema.org/Answer in some way, so the body of the reply can be modeled.  

Consider also #183 re RsvpAction

Notes: 

when the Question + Answer vocab was modeled we originally made an 'answer' property to point to answer. Later it became 'http://schema.org/suggestedAnswer' which was a superproperty of the more specific http://schema.org/acceptedAnswer.  Meanwhile the q/a-related action types AskAction, ReplyAction have an example (which uses AskAction but is catalogued against the ReplyAction type in data/examples.txt) which uses simple textual values (rather than e.g. 'text' property of an Answer or Question). This all needs some cleaning.

danbri:
Related vocabulary (see UserComments cleanup issue #170 : 

    <http://schema.org/replyToUrl> <http://schema.org/domainIncludes> <http://schema.org/UserComments> .

If we are modeling replies, we might pause briefly at least to consider possibility of modeling more of http://schema.org/EmailMessage (although modeling all of email headers is a big task).

danbri:
@vholland  

Vicki, I've filed this under Gozer for now. Any thoughts on whether that is feasible? Feels like a worthy thing to clean up alongside UserComments?

vholland:
It seems like we should be able to add the "comment" property to RsvpAction, so someone can provide something of type Comment.

The only question in my mind is whether "comment" will be renamed as part of issue #417.

danbri:
(I vote for leaving 'comment' as-is)

danbri:
@vholland so sdo-gozer already has 'comment' on RsvpAction, see http://sdo-gozer.appspot.com/docs/releases.html#sdo-gozer "F*x to #183: comment property now applicable to an RsvpAction, and can have Comment as values.".

Can you elaborate on how this relates to the ReplyAction and Answer situation? 

[...]

PR #467, opened 2015-05-01
Issue #300: Cleaned up AskAction and ReplyAction by tying them to Questi...

vholland:
...on and Answer respectively. Also cleaned up examples.


danbri:
Merged manually.

PR #477, opened 2015-05-07
Add JSON-LD example for workExample / exampleOfWork

### LINKED ISSUE #471: exampleOfWork and workExample lack JSON-LD example
schema.org/exampleOfWork and its inverse property, schema.org/workExample, both lack JSON-LD examples.  They both have microdata and RDFa examples, but the JSON-LD is present but empty (and without any explanation about _why_ it's empty).

dbs:
The workExample / exampleOfWork property examples were missing a JSON-LD
example. This fills in that missing piece, based on an example that Richard
Wallis provided.

Fixes #471.

### DIFF
--- a/data/sdo-periodical-examples.txt
+++ b/data/sdo-periodical-examples.txt
@@ -680,7 +680,67 @@ RDFA:
 </div>
 
 JSON:
- 
+
+<script type="application/ld+json">
+{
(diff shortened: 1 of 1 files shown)


danbri:
Merged, thanks @dbs !

Issue #77, opened 2014-07-30
Use "link" instead of "meta" element for URLs


unor:
Some (?) examples use a `meta` element for providing a URL. Instead, a `link` element should be used.

Example on http://schema.org/WebSite:

    <meta itemprop="url" content="http://www.example.com/"/>

    <meta property="url" content="http://www.example.com/"/>

----

According to HTML5 (W3C WD), the [`meta` element](http://www.w3.org/TR/2014/WD-html5-20140617/document-metadata.html#the-meta-element) is only to be used for metadata

> […] that cannot be expressed using the […] `link` […] elements

Microdata (W3C Note) also [requires](http://www.w3.org/TR/2013/NOTE-microdata-20131029/#values):

> If a property's value, as defined by the property's definition, is an absolute URL, the property must be specified using a URL property element.

danbri:
We should make these changes but also document the underlying situation more clearly.

dbs:
#478 addresses a similar issue in the existing examples in sdo-gozer, where a few instances of:

    <meta property="availability" href="http://schema.org/InStock">

(and similar `<meta itemprop-or-property href>` usages) had crept in over time.

danbri:
Punting this to next release as I am scared of corner-cases in our RDFa/RDFS pseudo-parser. But thank you for doing this, it is genuinely appreciated, honest :)

unor:
Some older pull requests that fix occurrences where `meta` was used for URI values:

* https://github.com/schemaorg/schemaorg/pull/599
* https://github.com/schemaorg/schemaorg/pull/600
* https://github.com/schemaorg/schemaorg/pull/593
* https://github.com/schemaorg/schemaorg/pull/591

RichardWallis:
Instances fixed in Htmlupdates request.

Thanks @dbs, @unor, @scor

PR #478, opened 2015-05-07
Change examples of <meta ... href=""> to <link ... href="">

dbs:
A few instances of <meta property="" href=""> had crept into examples, but
<link property="" href=""> is the preferred formulation.

Related to #77, which focuses on the more worrisome <meta property=""
content="http://example.com"> usage in examples.


danbri:
Punting this to next release as I am scared of corner-cases in our pseudo-parser.

danbri:
And btw, thanks for taking the time! Sorry not to have got it in this release...

RichardWallis:
merged into #787 


Thanks @dbs

PR #488, opened 2015-05-13
Fix typo in Stéphane Corlosquet's name

dbs:


danbri:
Oh :( thanks for catching this...

PR #497, opened 2015-05-14
fix typo of isAccessibleForFree

callahanchris:


jgoodell2:
f.y.i. An alternative (more flexible/powerful) way of handling "isAccessibleForFree" is modeled here: https://ceds.ed.gov/CEDSElementDetails.aspx?TermxTopicId=26893

The approach allows indicating specific Access Rights using URI/definition for options like FreeAccessWithRegistration, i.e. it is free to access ...but users are required to register...".


|   |
|   |   |   |   |   |
| Common Education Data Standards (CEDS)The Common Education Data Standards (CEDS) project is a national collaborative effort to develop voluntary, common data standards for a key set of education data elements to streamline the exchange and comparison of data across institutions and sectors. |
|  |
| View on ceds.ed.gov | Preview by Yahoo |
|  |
|   |

danbri:
Thanks for the bug report, @callahanchris. I've filed #508 rather than merge this patch since your diff is against a generated file. We build the file (including pages like http://schema.org/isAcccessibleForFree ) from the schemas in data/schema.rdfa, so I'll take it from here, thanks!

callahanchris:
Thanks for looking into this @danbri!

PR #500, opened 2015-05-15
Make change from an URL to a URL in schema.rdfa

shankarnat:


shankarnat:
Make a change from an URL to URL in schema.rdfa

danbri:
I don't understand what's going on in: https://github.com/schemaorg/schemaorg/pull/537 ... can you investigate @shankarnat ?

shankarnat:
Yes, I would review it. Thanks!

Issue #502, opened 2015-05-17
Confusion about PerformanceRole


Harald-Effenberg:
Hi!

I'm a little confused about the correct use of "PerformanceRole" - and obviously Google and Bing are confused, too.
AFAICS at http://schema.org/PerformanceRole#examples someone forgot to close a div in the example for "Microdata"?

If I try to imagine where the div should be closed, it seems to me that

      <div itemscope itemprop="performer" itemtype="http://schema.org/PerformanceRole">
      <meta itemprop="characterName" content="the charcter played" />
       <div itemscope itemprop="performer" itemtype="http://schema.org/Person">
              <a itemprop="url" href="http://example.com/"><span itemprop="name">The actor's name</span></a>
       </div>
     </div>

should work.

But Google's structured-data-testing-tool
https://developers.google.com/structured-data/testing-tool/?url=http%253A%252F%252Fjuri.effenberg.de%252Ftest.htm
wants a "name" for the PerformanceRole - although the attached "Person" already has a name.

The Yandex-Structured-data-validator thinks that everything is fine.

And the Bing-markup-testing-tool does not see any PerformanceRole at all.

Maybe the function and use of "PerformanceRole" should be communicated more clearly, so that Google, Bing and I can better understand how to work with it? ;-)

Warm regards
Harry

unor:
I think you are right about the missing `</div>` tags. I’ve created a pull request to fix this: https://github.com/schemaorg/schemaorg/pull/504

danbri:
Thanks for the fix.

As for external markup testing tools, I'll pass this along within Google. @tmarshbing might be in a position to do likewise at Microsoft.

PR #504, opened 2015-05-17
missing closing div tags in PerformanceRole example

unor:
Reported by @Harald-Effenberg in https://github.com/schemaorg/schemaorg/issues/502


danbri:
Thanks!:)

PR #511, opened 2015-05-19
Replaced meta tags with link tags for issue #77

stuartrobinson:
Changes to address the following issue:

https://github.com/schemaorg/schemaorg/issues/77

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -1441,7 +1441,7 @@ MICRODATA:
 
 <div itemprop="track" itemscope itemtype="http://schema.org/MusicRecording">
   <span itemprop="name">Rope</span>
-  <meta itemprop="url" content ="foo-fighters-rope.html">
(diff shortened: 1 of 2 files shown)


danbri:
Thanks! Should these be @content or @href?

stuartrobinson:
I think they should actually be @href. Let me update the branch.

unor:
+1 for the changes to `data/examples.txt` (the RDFa needs these fixes, too, but this could come at a later point).

The issues in `data/sdo-website-examples.txt` seem to have been fixed in the meantime (nitpick in the current version: the `url` value for the Microdata example should be `http://example.com/` instead of `http://www.example.com/`).

danbri:
Belatedly fixed and merged, including the s/www.example.com/example.com/ caught here by @unor. Thanks both!

danbri:
Merged to wrong branch, reverting and committing manually. Apologies for the noise and thanks again!

Issue #382, opened 2015-03-16
WebPage example does not use WebPage


Aaranged:
On:
http://schema.org/WebPage

We find this example (quoted is the "Without markup" tab)

> How to Tie a Reef Knot
> by John Doe
> This article has been tweeted 1203 times and contains 78 user comments.

However, the class schema.org/WebPage is not employed in any of the examples:  not in the microdata, RDFa or JSON-LD tabs.  The only class employed is schema.org/Article (and the only relation that Article has to WebPage is that they're both types of CreativeWork).

Accordingly, this example should either be changed to use WebPage, or removed.  The only likely consequence of using an "example" for WebPage that does not include WebPage is to confuse users that look at the example.

jvandriel:
Since @Aaranged started an example wishlist I'd like to add something to that.

In almost every use case I've encountered I've seen folks use the 'mainContentOfPage' property as if it's a 'mainEntity' property, meaning I've seen it be used in combination with types like Article, BlogPosting, Product, etc. but almost never have I seen it be used in combination with WebPageElement or it's subClasses.

Something I think finds it's origin in [a] lack of proper examples, and [b] lack of a mainEntity property. So If a proper example get's written could please make sure it also contains an example of how one is supposed to use the 'mainContentOfPage' property?

vholland:
Can we close this and take up mainContentOfPage and WebPageElement in another issue?

PR #512, opened 2015-05-19
Don't associated WebPage with Article example. Fix for issue #382.

stuartrobinson:
Fix for the following issue:

https://github.com/schemaorg/schemaorg/issues/382


vholland:
Thanks, Stuart!

Issue #137, opened 2014-09-20
replace interactionCount with interaction


elf-pavlik:
In one of last emails about differences between *potential* and *active/compleated* Action, I suggested introducing new property *interaction* which would supersede *interactionCount* -  http://lists.w3.org/Archives/Public/public-vocabs/2014Sep/0174.html

```js
{
  "@type": "WebPage",
  "@id": "http://example.net/",
  "interaction": [
    {
      "@type": "ViewAction",
      "@id": "http://mypds.net/joe/ew0r98a0ga",
      "actionStatus": "CompletedActionStatus",
      "object": "http://example.net/",
      "endTime": "2017-01-04T10:32:15+01:00"
    },
    {
      "@type": "ReadAction",
      "@id": "http://mypds.net/joe/0a9fd80gh7",
      "actionStatus": "CompletedActionStatus",
      "object": "http://example.net/",
      "startTime": "2017-01-04T10:32:15+01:00",
      "endTime": "2017-01-04T10:37:04+01:00"
    }
  ]
}
```

>While on it, I would like to bring attention (for the 4th time in last 2
years) to: http://schema.org/interactionCount

>And propose to replace it with *interaction* which would accompany
*potentialAction* on Thing.

>* Thing --{potentialAction}--> Action --{actionStatus}-->
PotentialActionStatus

>* Thing --{interaction}--> Action --{actionStatus}--> ActiveActionStatus
|| CompletedActionStatus

danbri:
Thanks for the detailed proposal. This clearly needs some attention now that Actions has shipped...

vholland:
For issue #170, we were going to deprecate all of UserInteraction, which makes interactionCount useless.

It would be useful to have some sort of AggregateInteractionCount class so pages could mention how many times something has been liked or tweeted, which I think was the original intent of UserInteraction, but that is for another release.

danbri:
I've generously assigned this to Vicki, since she is handling the UserInteraction area (and which looks likely to go away and take interactionCount with it).

Since the issue here is also forward looking, once interactionCount goes away, I'm moving it out from sdo-gozer into sometime-in-Q2.

vholland:
In the spirit of @elf-pavlik's proposal, I put together https://github.com/vholland/schemaorg/commit/a8c3e6c1a7bf4f2ba6f9b0a31d176cc06c36f3a2


The major difference is I did not leverage Action, for two reasons:
[...]

PR #514, opened 2015-05-20
Issue #137: Created InteractionCounter type and properties to supersede UserInteraction.

vholland:


danbri:
Thanks @vholland. @elf-pavlik - care to take a look?

zazi:
The term 'interaction' is included in every property and class that is related to this issue, I doubt that this is a best practice for good vocabulary design. Unfortunately, it is obviously necessary to keep everything under the schema.org namespace ... ;)

danbri:
@zazi yes it natural consequence of various design tradeoffs, to drift towards having soft / pseudo namespaces in various places. Although it might not look particularly elegant it does keep several things simpler for publishers/webmasters. I think we can live with it for now.

elf-pavlik:
@vholland LGTM :smile: thank you!

1) I wonder why *userInteractionCount* and not simply *interactionCount* ?
2) At some point we still could try to find a way of linking to the collection/container with actual references to all the interactions (e.g. Comment - all the comments, Share - all the shares, Follow - all the followers). In lines of what I tried to brainstorm in this comment https://github.com/schemaorg/schemaorg/issues/137#issuecomment-56270890)

vholland:
@elf-pavlik I was going to use `interactionCount`, but that already exists with different semantics. It seemed confusing to completely change the property's meaning. 

If people would rather, I can go back to that name.

elf-pavlik:
I just noticed
* https://schema.org/downvoteCount
* https://schema.org/upvoteCount
* https://schema.org/answerCount
* ...

https://schema.org/docs/search_results.html#q=count

danbri:
Merged in. We have some work on deprecation UI, but the new stuff shows up, e.g. http://webschemas.org/InteractionCounter (also in sdo-ganymede test site)

vholland:
SGTM. Feel free to kick back merges back to me.

unor:
Some HTML changes: https://github.com/schemaorg/schemaorg/pull/593

PR #517, opened 2015-05-20
Check APP_ENGINE env var before using hard-coded path

stuartrobinson:
Test script checks environment variable APP_ENGINE for path to Google AppEngine SDK before fall back on the hard-coded path.


danbri:
Thanks!

Issue #156, opened 2014-11-07
tests/test_graphs.py should test supercededBy and inverseOf don't point to same thing


danbri:
We've fixed bugs recently in which inverseOf and supercededBy link some url to itself. This is easy to test with, see tests/test_graphs.py for similar cases handled via SPARQL.

This would be a great thing to have contributed btw. Just sayin' !

(Otherwise I'll get to it eventually :)

PR #519, opened 2015-05-20
Fix for Issue #156. Minor cleanup. New tests for inverseOf and supercededBy

stuartrobinson:
Adds two new tests that make sure that inverseOf and supercededBy don't point to themselves. Fixes the following issue:

https://github.com/schemaorg/schemaorg/issues/156

### DIFF
--- a/scripts/run_tests.py
+++ b/scripts/run_tests.py
@@ -17,8 +17,10 @@ import argparse
 #
 # Alt: python -m unittest discover -s tests/ -p 'test_*.py' (problem as needs GAE files)
 
-def main(sdk_path, test_path, args):
-    sys.path.insert(0, sdk_path)
+def main(test_path, args):
+    sdk_path = getenv('APP_ENGINE',
+                      expanduser("~") + '/google-cloud-sdk/platform/google_appengine/')
+    sys.path.insert(0, sdk_path) # add AppEngine SDK to path
     import dev_appserver
     dev_appserver.fix_sys_path()
     print args, test_path
@@ -30,10 +32,7 @@ def main(sdk_path, test_path, args):
 
 
 if __name__ == '__main__':
-
-
     parser = argparse.ArgumentParser(description='Configurable testing of schema.org.')
-    
     parser.add_argument('--skipbasics', action='store_true', help='Skip basic tests.')
     args = parser.parse_args()
-    main(expanduser("~") + '/google-cloud-sdk/platform/google_appengine/', './tests/', args)
+    main('./tests/', args)
--- a/scripts/run_tests.py
+++ b/scripts/run_tests.py
@@ -2,7 +2,7 @@
 
 import optparse
 import sys
-from os import path
+from os import path, getenv
 from os.path import expanduser
 import unittest
 import argparse
--- a/tests/test_graphs.py
+++ b/tests/test_graphs.py
@@ -1,9 +1,16 @@
 import unittest
 import os
+from os import getenv
+from os.path import expanduser
 import logging # https://docs.python.org/2/library/logging.html#logging-levels
 import glob
 import sys
+
 sys.path.append( os.getcwd() ) 
+
+#sdk_path = getenv('APP_ENGINE',
+#                  expanduser("~") + '/google-cloud-sdk/platform/google_appengine/')
+#sys.path.insert(0, sdk_path)
 
 from api import *
 from parsers import *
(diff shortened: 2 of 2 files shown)


danbri:
Thanks @stuartrobinson. I'm still seeing a merge conflict but will pull this in and resolve any conflicts manually.

danbri:
Merged manually. Thanks for this @stuartrobinson and apologies it took so long! Aiming for "PR inbox zero" now...

jaygray0919:
May I ask a Google Structured Data Testing Tool (GSDTT) question here? Currently GSDTT does not process inverseOf. Will the planned changes here cascade into the way GSDTT processes inverseOf?

danbri:
the Google SDTT does process all properties. It doesn't support the experimental idea of an inverse-itemprop attribute (or similar) for Microdata. Is that what you had in mind?

jaygray0919:
Please ignore previous message - it was a brain fart. Our problem is not with `inverseOf`, which **is** properly processed by Google SDTT. Our problem is that we cannot use **@reverse** in schema.org. The **use case** is a follows: we use `Class` and `supercededBy` to define a tree. That allows us to specify the parent of a `Class` node. But we also need to specify that a `Class` node has children (i.e. `"@reverse": "rdfs:subClassOf"`). The reason is that we have an application that expects to see both  parent(s) and child/children when processing a tree structure. But we can't nudge a term into a schema.org `Class` structure that is not part of the data model (e.g. `"children": ("@reverse": "http://schema.org/Class"}`). Our solution is a total hack. We use `additionalType` as the term to specify children (we can tell our processor the term used for `children`). And then we hang our array of children on `additionalType`. Not elegant, but it works. Previously, I mentioned that we have the **MeSH** tree implemented using schema.org (and would contribute it to the greater good). This is the technique we use to build trees for our tree-processor.

Sorry to waste you time on an false aspersion. But it does allow me to explain one of the operating issues we face using schema.org. In many cases, we use `schema:Thing` and multiple **@graph** to handle extensions to a schema.org model. But we can't (or don't know how to) do it with trees. Hence the above hack.

FYI.

danbri:
Interesting work-around :) If you were trying to tweak the Google SDTT's knowledge of schema.org by passing in additional schema definitions, I can reassure you that it won't work for all kinds of unexciting reasons. Is the MeSH work posted anywhere?

jaygray0919:
First, here is MeSH in production (http://72.167.253.87/cgi-bin/flamenco.cgi/_ChEMATIC_v02_-_14-08-21_/Flamenco).

 [Tap F5 if the database is offline; it will load. This is a big database on an under-powered server, so please be patient when navigating the facets].

Currently, each facet is formulated using BFO and IAO. We now are converting the facet tree to schema.org, and I'll share when finished.

Separately, we have a tree that implements the ideas above. I'll post that to a Gist so you can see it. It is valid on Google SDTT. I'll also show you an application (our processor) that maps items to the tree. Those items are composed using schema.org. Further, the tree structure is properly rendered on @gkellogg Structure Data Linter (SDL).

But our processor needs to know the children; hence we purloin a term that Google SDTT will accept and use it as our key to define children. As an aside, our `d3.js` tree implementation (using the dendrogram pattern) **does not** need to know children. So that 'processor' can render a kosher schema.org tree.

Obviously, the right way to do this is with **@reverse**. When we use non-schema.org Controlled Vocabularies (CV) we can successfully use `@reverse rdfs:subClassOf`. In that case, the SDL  properly renders the tree.

Gimme a day or two to set this up; our folks are heads-down on composition tasks, so I'll need to do this myself. But your analysis is valuable so I'll set it the examples soon.

jaygray0919:
@danbri here is the gist:
  https://gist.github.com/jaygray0919/669fc6693e9d25712c81142c06b0f727
It is Google SDTT valid.

Here is an application of the schema.org tree:
  http://ontomatica.com/public_new/project/test/Exhibit/publications/bibliography/bib_B/index.html

At this moment, only one publication is mapped to the technology tree. Our goal was simply to test the gist for one item.

Bottom line: we generate files for three processors; a facet navigation processor, a d3.js processor, and a SPARQL processor.

Given that we can generate schema.org trees (Class structures), items and their maps (relationships), we now can migrate examples like MeSH to a Google SDTT-valid platform.

The above is less of a statement and more of a question: are we going in the right direction? TY in advance for advice.

jaygray0919:
@danbri Updated gist to remove null data.

Issue #531, opened 2015-05-22
MusicCompositions should have a way to specify the song lyrics


vholland:
MusicComposition has a `lyricist` property, but not a `lyrics` property. It would be nice to be able to specify the lyrics themselves.

vholland:
See pull request #532.

danbri:
I commented offtopic in the pull request, while meaning to comment offtopic in the issue :)

"Thanks! I was wondering if this should be attached more widely, but talked myself out of it.

If someone applies a 'lyrics' property to a thing typed only as e.g. VideoObject, some validators would probably complain. I think instead of adding 'lyrics' up at a safer generic level we should try to make the culture of schema.org validation softer. A reasonable interpretation of "x is a VideoObject and it has a lyrics property whose value is y" would be to conclude "oh, x is also probably a MusicComposition then". I don't believe its reasonable to require that every usage of "lyrics" on a VideoObject also requires the mention of a 2nd type, MusicComposition. This just adds weight for limited benefit."

danbri:
Fixed by #532

PR #532, opened 2015-05-22
Issue #531: Added lyrics property to MusicComposition

vholland:


danbri:
Thanks! I was wondering if this should be attached more widely, but talked myself out of it.

If someone applies a 'lyrics' property to a thing typed only as e.g. VideoObject, some validators would probably complain. I think instead of adding 'lyrics' up at a safer generic level we should try to make the culture of schema.org validation softer. A reasonable interpretation of "x is a VideoObject and it has a lyrics property whose value is y" would be to conclude "oh, x is also probably a MusicComposition then". I don't believe its reasonable to require that every usage of "lyrics" on a VideoObject also requires the mention of a 2nd type, MusicComposition. This just adds weight for limited benefit.

vholland:
I can see both sides of this. Coming from Freebase, it seems weird to put the lyrics on the MediaObject, but it makes the markup more complex as one needs to create a MusicComposition simply to specify the lyrics.

chaals:
+1 to approaching this through the validation... if something has a property X, whose range is Z, then perhaps something *is* of type Z even if we didn't say so?

danbri:
@chaals I guess you mean domain (our terminology follows http://www.w3.org/TR/NOTE-MCF-XML/simple.gif), in which case that makes sense. In practice we document the schemas with a weaker property http://schema.org/domainIncludes which allows us to evolve things without formally surprising anyone.

@vholland wouldn't the MediaObject simply also *be* a MusicComposition (whether explicitly stated or not), or do you have in mind an additional related entity?

chaals:
@danbri Yeah, if we used "domain" we would get that forced on us. It isn't quite a simple question, since there is value in people being able to validate and find that they didn't *declare* something implied by their domain, but this is probably a question of the culture around validator implementation rather than formal schema semantics...

danbri:
@chaals - yes, softer gentler validators could still give sensible "did you really mean...?" warnings.

Perhaps formally there is nothing saying that something cannot be simultaneously a CreativeWork and a Person, but in practice when you encounter such a combination it is much more likely to be due to markup errors e.g. bad tag nesting, than because they're describing artistic tattoos or a performance artist.

danbri:
Back re lyrics, I propose to merge this pull request. Is that OK with everyone here?

chaals:
Merge: Yes.

Person who is a CreativeWork: Sherlock Holmes. Who even has a home address.

But yeah, validators that said "are you sure Dan Brickley is a volcano?" would be useful…

danbri:
Merged! Pretty sure you're a CreativeWork too :)

vholland:
@chaals @danbri Coming from the Freebase world, I instinctively separate out the MediaObject from the MusicComposition, but there is no reason schema.org needs to work that way. Creating intermediate entities gets tedious and there is no reason we can't infer the relationship as @chaals suggested.

danbri:
Ok merged and I've updated our test build: http://sdo-ganymede.appspot.com/docs/releases.html#sdo-ganymede

Thanks @vholland for taking the time to include an edit for /docs/releases.html too, it makes all the difference :)

Issue #143, opened 2014-10-02
Examples on openingHours should not use the 'time' element


unor:
http://schema.org/openingHours

The examples on [`openingHours`](http://schema.org/openingHours) use HTML5’s `time` element:

    <time itemprop="openingHours" datetime="Tu,Th 16:00-20:00">Tuesdays and Thursdays 4-8pm</time>
    <time itemprop="openingHours" datetime="Mo-Su">Monday through Sunday, all day</time>

The values "Tu,Th 16:00-20:00" and "Mo-Su" don’t seem to be valid [according to W3C’s HTML5](http://www.w3.org/TR/2014/PR-html5-20140916/text-level-semantics.html#the-time-element) or [according to WHATWG’s HTML](https://html.spec.whatwg.org/multipage/semantics.html#the-time-element).

The examples on [`LocalBusiness`](http://schema.org/LocalBusiness) use the `meta` element instead, which seems to be appropriate. The [`data` element might also be an appropriate alternative](http://stackoverflow.com/a/26159198/1591669).

danbri:
I had a chat with Hixie about possibly getting HTML updated to allow this, if/when we get this micro-notation properly spec'd. But for now, I agree, time element is in-apropes.

danbri:
Another nudge that we should address this. https://twitter.com/Nettsentrisk/status/574859464316219392 

I'd suggest:

 - [ ] stop promoting <time> for this markup 
 - [ ] specify the openingHours microsyntax more formally, suggest it to HTML5 groups for use with <time>

danbri:
This needs doing, but bumping one more milestone.

Here's a grep of what's in the examples:

    <time itemprop="validFor" content="P1Y">1 year</time>
     <time property="validFor" content="P1Y">1 year</time>
     <div>asked <time itemprop="dateCreated" datetime="2010-11-04T20:07Z">Nov 11 '10 at 20:07</time></div>
       <div>answered <time itemprop="dateCreated" datetime="2010-12-01T22:01Z">Dec 1 '10 at 22:01</time></div>
       <div>answered <time itemprop="dateCreated" datetime="2010-12-06T21:11Z">Dec 6 '10 at 21:11</time></div>
     <div>asked <time property="dateCreated" datetime="2010-11-04T20:07Z">Nov 11 '10 at 20:07</time></div>
       <div>answered <time property="dateCreated" datetime="2010-12-01T22:01Z">Dec 1 '10 at 22:01</time></div>
[...]

PR #533, opened 2015-05-23
OpeningHoursSpecification example: various fixes

unor:
* `time` elements also for plain markup example
* using `meta` instead of `time` for `openingHours` (as "Mo,Tu,We,Th,Fr,Sa,Su 09:00-14:00" is not a valid `@datetime` value, see https://github.com/schemaorg/schemaorg/issues/143)
* no need for an empty `time` element for `validThrough`, as `@itemprop`/`@property` can have multiple properties
* JSON-LD example had different data than the markup examples, and the second `OpeningHoursSpecification` was missing

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -10995,16 +10995,16 @@ PRE-MARKUP:
 <div>
   <h1>Middle of Nowhere Foods</h1>
   <h2>Opening hours</h2>
-  <p>Normally open daily 9am-2pm except on:</p>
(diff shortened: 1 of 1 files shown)


danbri:
@unor - I merged this into sdo-ganymede rather than sdo-deimos without thinking. Fixing manually with a new pull request to sdo-deimos that has conflicts. Possibly these have been fixed already at least in part.

PR #537, opened 2015-05-24
Revert 500 urlchange

abol:


danbri:
Can you explain what's happening here?

danbri:
Ah, I see. The frozen snapshot in /version/ should not have been changed.

Issue #184, opened 2014-11-23
Some Microdata examples use a 'content' attribute for 'span'


unor:
(Coming from this Stack Overflow question: [HTML5 microdata: span content?](http://stackoverflow.com/q/27088904/1591669))

http://schema.org/Book contains Microdata examples containing

```
<span itemprop="price" content="6.99">$6.99</span>
<span itemprop="inLanguage" content="en">English-language</span>
<span itemprop="name" content="Tolkien, J. R. R. (John Ronald Reuel)">J. R. R. Tolkien</span>
```

But a `span` element can’t have a `content` attribute in Microdata (contrary to RDFa).

danbri:
We should also add a FAQ or Wiki page explaining the confusing situation w.r.t. @content in Microdata vs RDFa, and meta element in <body> in HTML5 vs earlier HTML versions.

unor:
(I fixed the mentioned cases in https://github.com/schemaorg/schemaorg/issues/552)

We should also adjust the usage guideline on [`price`](http://schema.org/price):

> Note that both RDFa and Microdata syntax allow the use of a "content=" attribute for publishing simple machine-readable values alongside more human-friendly formatting.

unor:
Here is a PR that fixes this in two more examples: #1102

danbri:
Looks like we have a lot of these (via "grep itemprop examples.txt | grep content"). Perhaps there's a case to have @content in Microdata? /cc @chaals

gkellogg:
It would be useful, but seems to miss the point of Microdata being simple. There's always RDFa for being precise. But, if the community agrees, I'd support adding @content to Microdata. Note that the HTML validator may continue to reject this, as it is now allowed because of special rules for RDFa, so HTML 5.1 May need to include it to be valid Microdata. Given that it seems to be in general use, it should get past the hixie filter.

PR #552, opened 2015-05-27
Book examples: fixes

unor:
In Microdata, only the `meta` element can have a `content` attribute (see https://github.com/schemaorg/schemaorg/issues/184):

* for `inLanguage`: using `meta` instead of `span` in Microdata (keeping `span` in RDFa)
* for `price`: omitting `content` in Microdata and RDFa, because it was not needed (moving the "$" outside of the `price` property)


RichardWallis:
#648 supersedes this - can be closed now

RichardWallis:
Superseded by #648

PR #556, opened 2015-05-28
Changed from an URL to a URL

shankarnat:


shankarnat:
#500 URL changes has got reverted and hence merging again to ensure that An URL changes to A URL

danbri:
Actually there was a fair point:  data/releases/2.0/schema.rdfa should not have been changed, as it was a frozen(ish:) release snapshot. The live schema in data/schema.rdfa is where we should be putting new improvements like "an URL -> a URL". However I don't believe that #500 did get reverted yet as I didn't merge the pull request. This is complicated as the pull is also targetted at sdo-gozer. I'll take a look and clean something up...

shankarnat:
Okay, Should I make revoke this pull request and make changes only in the data/schema.rdfa. I can do it; if that can work and create a new pull request.

danbri:
Sounds good, thanks! Target just sdo-ganymede, don't worry about the other branches :)

shankarnat:
Made changes to ensure only data/schema.rdfa is update. The Release 2.0 folder was reverted to the old release version. @danbri  please verify and merge. Thanks!

danbri:
Ok so Github (and git) can be confusing!

* here in Github UI I see this as a pending change of a few "an" to "a" in schema.org
* locally on my machine (after 'git pull --all') these are already written as "a"
* I figure it can't do any harm, so I merge it anyway.
* I do another "git pull --all" locally
* I verify that my local copy still has "a" in data/schema.rdfa, so we're ok.
* Looking at https://github.com/schemaorg/schemaorg/commit/06b858c7e2c3e8d826539cdeb0b4d40347b1c359 it confirms my suspicion that the pull request didn't actually change anything ("Showing  0 changed files  with 0 additions and 0 deletions.")

Anyway, thanks - I think we're fine now! This isn't the first time I've been tripped up by the UI around pull requests, I wonder if I'm misunderstanding something subtle...

shankarnat:
Thanks @danbri .

PR #562, opened 2015-06-01
CreativeWork/ItemList example: byArtist expects MusicGroup, not Person

unor:
The example for `CreativeWork` and `ItemList` used `Person` as value for the `byArtist` property, but the expected value is `MusicGroup` (even for solo musicians).


danbri:
(thanks @RichardWallis for taking a look)

PR #563, opened 2015-06-02
RDFa changing tag span to meta

Jonadabe:
Attribute content not allowed on element span.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -56,7 +56,7 @@ RDFA:
   <img src="janedoe.jpg" property="image" alt="Photo of Jane Joe"/>
 
   <span property="jobTitle">Professor</span>
-  <div property="address"  typeof="PostalAddress">
+  <div property="address" typeof="PostalAddress">
     <span property="streetAddress">
       20341 Whitworth Institute
       405 N. Whitworth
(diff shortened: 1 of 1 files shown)


scor:
@Jonadabe could you reference where you saw that the content attribute was not allowed in the span element? this is perfectly fine in HTML5 at least. https://validator.w3.org/nu/ doesn't complain.

Jonadabe:
@scor Thank you. I did a check (https://validator.nu/) and it sended me to https://html.spec.whatwg.org/#global-attributes and I didn't saw a content tag, but you are right with the o https://validator.w3.org/nu/ its ok!
Where can I see some more information about this in HTML5?

unor:
@Jonadabe: RDFa [extends](http://www.w3.org/TR/2015/REC-html-rdfa-20150317/#extensions-to-the-html5-syntax) HTML5:

> For the avoidance of doubt, the following RDFa attributes are allowed on all elements in the HTML5 content model: […], `@content`, […].

danbri:
The vast bulk of this pull request seems to be about minimising whitespace between attributes - makes it quite hard to focus on the substantial issue. Is the main concern that we have used @content on <span>?

gkellogg:
The @content tag is valid on the span element, not because it is defined in HTML5, but because it is defined in RDFa, and is considered valid by validator.nu, which took a bit of work to make happen.

Indeed, whitespace is insignificant, but properly self-closing, or adding an end-tag may be, depending on how good your HTML parser can otherwise detect when a tag should be closed. As a best practice, self-close elements without a content model, and use a closing end-tag for those having a content model.

Jonadabe:
@unor Thanks, now I understand it.
@danbri Dan, yes it was.
@gkellogg Gregg, thank you so mutch for your clear and direct answer.

RichardWallis:

PR #320, opened 2015-02-05
[CODE] Handle display of Classes with multiple super-classes

Dataliberate:
1. Make Thing appear at end of ‘Properties from’ listing
2. Remove Inheritance display (breadcrums) from label display, replace with separate listing below for all inheritance path(s)

PR #567, opened 2015-06-04
[Code] Multi inheritance breadcrumbs

Dataliberate:
* Remove Inheritance display (breadcrums) from label display, replace with separate listing below for all inheritance path(s). eg:
```
LocalBusiness
Thing > Organization > LocalBusiness
Thing > Place > LocalBusiness
```

* Make Thing appear at end of ‘Properties from’ listing

Pull #320 updated for sdo-ganymede


danbri:
Thanks Richard! Just tested it locally, looks a lot better :)

Issue #565, opened 2015-06-03
Schema.org RDFa may need to replace self-closing tags to proper end tags.


boanuge:
I was trying to analyze the RDFa file from the schema.org.
https://github.com/schemaorg/schemaorg/blob/master/data/releases/2.0/schema.rdfa

I found an error when I try to validate the RDFa file with the W3C markup validation service(https://validator.w3.org/) and html parser library(htmlparser-1.4.jar by nu.validator). One of reason could be the self-closing tags.

          e.g. <span xxxxxx></span> is ok, but <span xxxxxx /> is not ok.

The error given by the W3C markup validation service: Self-closing syntax (/>) used on a non-void HTML element. Ignoring the slash and treating as a start tag.

Reference: http://dev.w3.org/html5/spec-author-view/syntax.html#syntax-start-tag

gkellogg:
This is an HTML requirement, not RDFa. Span has a content model, and is not self-closing. It will not validate. If parsed as XML, an RDFa parser won't care, but an HTML parser might, although it will typically work okay.

boanuge:
I think it would be good to make RDFa page to meet the HTML(5) requirement,
because one of the reason to use RDFa is to embed metadata in HTML page without violation.
BTW, thank you for the comment. :)

unor:
I created a pull request that fixes the `span`+`href` cases (by using `link` instead): https://github.com/schemaorg/schemaorg/pull/568

The only missing HTML errors seem to be 9 cases where a `p` is used inside of `span`, e.g.:

    <span property="rdfs:comment">Describes a reservation for travel, dining or an event. Some reservations require tickets.<p>Note: This type is for information about actual reservations, e.g. in confirmation emails or HTML pages with individual confirmations of reservations. For offers of tickets, restaurant reservations, flights, or rental cars, use http://schema.org/Offer.</p></span>

I think the best solution would be to use a `div` element instead of a `span` element for all the comments (`rdfs:comment`). (I did not fix these because I’m not sure if the current CSS would handle it.)

danbri:
I've merged in @unor 's fix #568 after testing that our pseudo-RDFa hack-parser can read the improved syntax variation.

PR #568, opened 2015-06-04
<link> instead of <span> for URI values

unor:
Reported by @boanuge in https://github.com/schemaorg/schemaorg/issues/565

(Older, already merged PR that fixed the same kind of error: https://github.com/schemaorg/schemaorg/pull/437)


danbri:
Thanks -just tested this locally to confirm we can still parse it ok, seems fine!

PR #570, opened 2015-06-06
WatchAction example in HTML/Microdata/RDFa

unor:


unor:
@danbri: I’m not sure about the process here … should I create an issue for reviewing/discussing the markup for examples, or is the comment function here in the pull request sufficient?

danbri:
I think in general it is better to make sure each Pull Request addresses a separate real issue, otherwise we have tended to get solutions to under-documented problems. But in this case it looks like progress, I think fine to merge it in... cheers:)

PR #575, opened 2015-06-06
Potential Actions doc: HTML fixed, IDs added

unor:
Changes to http://schema.org/docs/actions.html

The [first commit](https://github.com/schemaorg/schemaorg/commit/1499926fab54338f5e1dc8306fe2cf459057fbd5) fixes all HTML errors (I did *not* test if this commit changes the design!):

* duplicated `DOCTYPE`, `head`, and `body` (I removed the duplicates, and I moved both inline styles into one `style` element)
* `pre` inside `p` (I replaced `p` with `div`)

The [second commit](https://github.com/schemaorg/schemaorg/commit/9e9932cb9b4c3f08445810f4153ac916135d6f03) adds `id` attributes to the four parts (`h2`), so that they can be linked.

### DIFF
--- a/docs/actions.html
+++ b/docs/actions.html
@@ -108,7 +108,7 @@
     background. Earlier drafts and alternate formats are available from the <a href="https://www.w3.org/wiki/WebSchemas/ActivityActions">W3C Wiki</a>.
 </p>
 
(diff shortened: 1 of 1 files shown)


danbri:
Thanks, that looks better!

PR #590, opened 2015-06-11
BreadcrumbList example: HTML fix

unor:
The `ol` element can’t have text (like `›`) as content, only `li` elements. (If a list is used, authors should use CSS to insert visual separators.)


RichardWallis:
This fix applied eleswhere

PR #523, opened 2015-05-21
Issue #518: Added the new enumeration http://schema.org/PaymentStatusType as an expected type for http://schema.org/paymentStatus.

vholland:

PR #591, opened 2015-06-11
Invoice examples: HTML fixes

unor:
Using `link` instead of `meta` for example change from https://github.com/schemaorg/schemaorg/issues/523

**Update**: I made some more changes to the examples in the `sdo-invoice-examples.txt` file:

* removed Microdata attributes from RDFa examples
* it seemed that the URL for the "Invoice PDF" could be a visible `a` instead of an invisible `link`
* some RDFa examples used full Schema.org URIs for types although `vocab` was specified
* removed a property that was only specified in the RDFa example, but no in the pre-markup, Microdata, or JSON-LD
* used the `time` element for the visible dates
* an RDFa example used `itemprop` instead of `property`


RichardWallis:

PR #593, opened 2015-06-11
InteractionCounter: HTML fixes

unor:
* `meta` can’t have a URI value: using `link` instead

* `span` can’t have a `href` attribute: using `link` instead

* canonical URLs for the mentioned third-party websites

* `span` can’t have a `content` attribute in Microdata; to keep the examples consistent, I removed the `content` attributes from both, Microdata and RDFa, and used numbers without delimiter instead 

 (if we want to keep the delimiters, we could use the `data` element instead, with `@value` in Microdata and `@value`+`@content` in RDFa; or the `meta` element for both)


RichardWallis:

PR #597, opened 2015-06-12
Releases doc: HTML fixes

unor:
* the end tag for an `a` element was missing
* `span` elements were self-closed (which is not allowed)

(The document is still not valid, i.e., there’s a `style` element in the `body`, and the `span` elements are children of `tr` rather than `td`.)

Related PR: https://github.com/schemaorg/schemaorg/pull/586


RichardWallis:

PR #599, opened 2015-06-13
Social media examples: HTML corrections

unor:
* `meta` was used for URI value (and had a closing tag): used `link` instead
* plain HTML examples had a Microdata attribute: removed it
* `p` can’t contain `div`: used `span` instead
* `p` closing tag missed a `/`


RichardWallis:

PR #600, opened 2015-06-14
WebSite example: HTML fix

unor:
* using `link` instead of `meta` for `url` property
* RDFa: making use of `vocab` (i.e., `typeof="SearchAction` instead of `typeof="http://schema.org/SearchAction"`)

(I’m not sure if the value of the `target` URI should be a URL or not if no `EntryPoint` type is used, so I left it untouched.)


RichardWallis:

PR #638, opened 2015-07-09
Publisher includes person

### LINKED ISSUE #611: A publisher can be a person.
`schema:publisher` defines itself with "The publisher of the creative work." Its range is `schema:Organization`. Is there a particular reason why `schema:Person` is not one of the possible ranges?

The definition of _publisher_ from the English dictionary generally states it to be a person or business/organization with such role. I'd like to propose the addition of:

`schema:publisher schema:rangeIncludes schema:Person .`

chaals:
to resolve #611


danbri:
Seems sensible to me!

PR #646, opened 2015-07-15
Moved breadcrum property under WebPage for JSON-LD example for Book

RichardWallis:
Also mapped & to &amp; in same example


danbri:
Merged, although see comment in issue - considering removing 'breadcrumb' from this example

Issue #347, opened 2015-02-18
Keywords property of CreativeWork


lazaruscorporation:
Hi

The current instruction on http://schema.org/keywords is "Keywords or tags used to describe this content. Multiple entries in a keywords list are typically delimited by commas."

I'd like to suggest that we also allow the keywords property to be used multiple times within a Type when describing something (and update the informative text on http://schema.org/keywords to describe this), as this will allow easier parsing by a client. For example:

```
<div vocab="http://schema.org/" typeof="Movie">
  ...
  <ul>
    <li property="keywords">Ghostbusters</li>
    <li property="keywords">Movies</li>
    <li property="keywords">etc</li>
  </ul>
```

This would also allow for better LOD implementation (as proposed in #346 ) as follows:

```
<div vocab="http://schema.org/" typeof="Movie">
  ...
  <ul>
    <li property="keywords" resource="http://example.org/lod/uri/ghostbusters">Ghostbusters</li>
    <li property="keywords" resource="http://example.org/lod/uri/movie">Movies</li>
    <li property="keywords" resource="http://example.org/lod/uri/etc">etc</li>
  </ul>
```

Paul

danbri:
While I appreciate the general motivation, in this case wouldn't 'about' work? http://schema.org/about
i.e. "The subject matter of the content."? Do the URLs you give correspond to a specific set of keywords, or to the underlying concept?

lazaruscorporation:
Hi @danbri 

I'm not sure and it's probably one of those grey areas open to interpretation (at least for me!). 'about' expects an instance of schema.org/Thing, not a URL, so it's not ideal for use for linked data. Yes, you could mark-up an instance of 'Thing' and then use that instance's 'sameAs' property to implement LOD, but that starts to produce messy & extraneous markup (example below):

```
<div vocab="http://schema.org/" typeof="Movie">
  ...
  <ul>
    <li property="about" typeof="Thing"><span property="sameAs" resource="http://example.org/lod/uri/ghostbusters">Ghostbusters</span></li>
    <li property="about" typeof="Thing"><span property="sameAs" resource="http://example.org/lod/uri/movie">Movies</span></li>
    <li property="about" typeof="Thing"><span property="sameAs" resource="http://example.org/lod/uri/etc">Etc</span></li>
  </ul>
</div>
```

[...]

Issue #346, opened 2015-02-18
Linked Open Data: extend rangeIncludes of 'genre' + 'keywords' properties of CreativeWork to include URL


lazaruscorporation:
Hi

I'd like to suggest extending the rangeIncludes of the 'genre' ~~and 'keywords'~~ property of schema.org/CreativeWork (currently just "Text") to also include URL.

This will allow advanced users to implement more Linked Open Data (LOD) in schema.org, by replacing a free-typed text string in 'genre' ~~or 'keywords'~~ with a URI from a LOD vocabulary.

For example, the "genre" property for a Late Renaissance painting (marked up using VisualArtwork, a sub-Type of CreativeWork) could be set to the Getty AAT LOD URI  "http://vocab.getty.edu/aat/300021143" instead of the text value "Late Renaissance". 

~~Similar LOD URIs (e.g. from Wikidata etc.) could be used in the "keywords" property.~~

There may be other properties that can be included in this enhancement to incorporate LOD, but these two seemed the most obvious.

Thanks,

Paul

lazaruscorporation:
Edit: following the resolution of #347 (thanks to @danbri and @Dataliberate) this request to extend the rangeIncludes to URL (in addition to text) should only apply to the "genre" property, not the "keywords" property

realworldobject:
I agree this would be very helpful for improving library linked data because we could leverage multilingual labels and descriptions such as with http://www.wikidata.org/entity/Q24925.

niklasl:
+1

This [has come up before](http://lists.w3.org/Archives/Public/public-vocabs/2013Nov/0126.html), during the SKOS discussion now tracked in issue #251.

(There is a larger question about whether rangeIncludes URL implies Thing, and perhaps whether genre should be narrowed to e.g. Intangible. But since URL is already used like this elsewhere, this specific case might not be the place to resolve that.)

danbri:
Looking at this again, I see that we implemented it for /genre (with an accidental rollback fixed in #1203 and about to be republished in v3.1, see #1212). But we didn't get there for /keywords. I see also @lazaruscorporation retracted the proposal to do this for /keywords, and that anyway we're discussing that kind of functionality elsewhere including #251. I'm therefore going to close this issue. Thanks everyone!

Issue #207, opened 2014-12-25
Missing /Bridge as subclassOf /CivicStructure


redaktor:
http://schema.org/CivicStructure

danbri:
Added in @RichardWallis 's merge - https://github.com/schemaorg/schemaorg/pull/648/files#diff-200232fafee461b4217385c536cb093eR2331

danbri:
This is queued up for review in the next release (sdo-ganymede). Thanks!

Issue #374, opened 2015-03-07
Add Report under Creative Work  


jkreft-usgs:
Reports generated by governmental and non-governmental organizations are currently difficult to describe using schema.org, as they aren't a scholarly article, but don't really fit in anywhere else under CreativeWork either.  Report as a type under CreativeWork would give this common publication type an accurate description.  Properties for Report could be similar to that of Article, but would also include the report number, a key piece of metadata to describe reports.
- Report
  - reportBody (Text) The actual body of the report.
  - pagination (Text or Integer) Any description of pages that is not separated into pageStart and pageEnd; for example, "1-6, 9, 55" or "10-12, 46-49".
  - pageStart (Text or Integer) The page on which the work starts; for example "138" or "xvi".
  - pageEnd  (Text or Integer) The page on which the work ends; for example "138" or "xvi".
  - wordCount  (Integer) The number of words in the text of the Report.
  - reportNumber (Text or Integer) The number or other unique designator assigned to the Report by the publishing organization.

This is an example of a report that I am currently struggling to properly describe with schema.org- a Report CreativeWork type would immediately be put to use here:  http://pubs.er.usgs.gov/publication/sir20145235

Dataliberate:
Just an observation but would a Report not be just as well served by being a subtype of Article with one new property ‘reportNumber’.

Thing > CreativeWork > Article > Report

~Richard

On 7 Mar 2015, at 00:35, jkreft-usgs <notifications@github.com<mailto:notifications@github.com>> wrote:


Reports generated by governmental and non-governmental organizations are currently difficult to describe using schema.org<http://schema.org>, as they aren't a scholarly article, but don't really fit in anywhere else under CreativeWork either. Report as a type under CreativeWork would give this common publication type an accurate description. Properties for Report could be similar to that of Article, but would also include the report number, a key piece of metadata to describe reports.

  *   Report
     *   reportBody (Text) The actual body of the report.
     *   pagination (Text or Integer) Any description of pages that is not separated into pageStart and pageEnd; for example, "1-6, 9, 55" or "10-12, 46-49".
     *   pageStart (Text or Integer) The page on which the work starts; for example "138" or "xvi".
     *   pageEnd (Text or Integer) The page on which the work ends; for example "138" or "xvi".
     *   wordCount (Integer) The number of words in the text of the Report.
     *   reportNumber (Text or Integer) The number or other unique designator assigned to the Report by the publishing organization.

[...]

PR #648, opened 2015-07-17
Assorted fixes

RichardWallis:
Issue (#207) Added Bridge type
Issue (#346) Added URL to range of genre
Issue (#347) Added Report type & reportNumber property
Issue (#382) Removed examples that don't reference WebPage from WebPage
Issue (#552) Fixed Book examples


danbri:
This text,

ssue (#207) Added Bridge type
Issue (#346) Added URL to range of genre
Issue (#347) Added Report type & reportNumber property
Issue (#382) Removed examples that don't reference WebPage from WebPage
Issue (#552) Fixed Book examples

... is almost what we need for docs/releases.html  - can you adapt?

danbri:
re report type, I think intent was to write #374 not #347

Issue #429, opened 2015-04-16
Implement the new extensions model 


danbri:
Meta bug. See discussion thread https://lists.w3.org/Archives/Public/public-vocabs/2015Mar/0117.html

Related: draft bib: extension #431 https://lists.w3.org/Archives/Public/public-schemabibex/2015Apr/0000.html

First cut code was merged from sdo-scripts into sdo-gozer, April 27th.

# TODOs

## Site navigation

For any running installation of the schema.org site, ... considering a base domain such as 'schema.org', 'sdo-gozer.appspot.com', 'webschemas.org' etc., here written BASE_DOMAIN:

- [x] foo.BASE_DOMAIN/Person (since Person is a term defined in the core, and 'foo' is a hypothetical extension) should display as normal, while indicating that we are viewing the 'foo' extension.
- [x] xyz.BASE_DOMAIN/Person etc. should completely fail, e.g. 404 or browser resolution error, if 'xyz' is not enabled as an extension name on that installation.
- [x] foo.BASE_DOMAIN/XyzAbc should fail by 404 if XyzAbc is not a known term in any enabled extension running on that installation.
- [x] BASE_DOMAIN/Globe should fail with a helpful 404, with links to each extension e.g. 'foo' which defines/mentions the term 'Globe' (or any term that is known to an extension enabled in that installation).
- [x] foo.BASE_DOMAIN/Globe should succeed, rendering a per-term page describing Globe, assuming Globe is a term defined in an extension 'foo' enabled on that installation. Navigation to related terms from core and from within 'foo' by normal means, i.e. Place, Event, CreativeWork, name etc from core would be listed and navigable "within the extension". STATUS: http://bib.webschemas.org/Globe?ext=bib works but needs the ext=bib still. 
- [x] foo.BASE_DOMAIN/ should present a simple homepage for the extension, including a short description of its purpose/scope, and a link to github for development. It should also list all terms defined by the extension, including their (e.g. supertype, subtype) connection to the core. This assumes content-type requested was text/html. 
- [ ] foo.BASE_DOMAIN/docs/datamodel.html and other generic documentation should redirect transparently to BASE_DOMAIN/docs/datamodel.html to avoid proliferation of multiple URLs for the same static content.
[...]

PR #658, opened 2015-07-23
Introduce domain (extension) specific data caching

RichardWallis:
Work in support of #429 
Needed so that different views of same Type eg. bib.schema.org/Book & schema.org/Book could be served up (and cached) by the same instance of the site.


danbri:
Thanks for figuring this out - it was driving me nuts ;)

Issue #684, opened 2015-07-30
Extension defined term links display in different colour(s)


RichardWallis:
To provide visual clue that term is in an extension

Issue #681, opened 2015-07-30
Full listing extension display capability


RichardWallis:
docs/full  needs capability to display at least 3 states:
1. Core terms only
2. Core plus currently displayed extension
3. Core plus all extensions

Ideally an option to display only terms from the current extension (plus their supertypes) would be useful.

RichardWallis:
Fixed ages ago.

PR #685, opened 2015-07-30
Extension full display and link css

RichardWallis:
Added CSS for extension defined term links bib / auto (#684)
Added selectable full listings for current view (core or core + extension) and core plus all extensions (#681)
* If in an extension default list will be for core plus the extension, if in core it will be for core only.
* Selectable option is for core plus all extensions.


danbri:
Thanks!

PR #697, opened 2015-08-05
removed vin and changed status note in homepage.tpl

mfhepp:
These are the last minor issues before auto.schema.org could go live.


danbri:
Thanks. I've removed vin separately, and reworded the homepage following our discussion.

PR #724, opened 2015-08-07
Fix references to www.schema.org to point to standardized non-www URLs

scor:


danbri:
Thanks :) There were some quick fixes via master after sdo-phobos was created, I'll sort out that admin first...

scor:
Happy to rebase this pull request on any branch. just when you're ready...

danbri:
If you can point it at 'master' that would be great, thanks!

danbri:
Thanks!

Issue #725, opened 2015-08-07
Populate meta description with item description


Aaranged:
Currently all schema.org pages carry the same `<meta>` description.

Person - schema.org
http://schema.org/Person
`<meta name="description" content="Schema.org is a set of extensible schemas that enables webmasters to embed structured data on their web pages for use by search engines and other applications." />`

Thesis - bib.schema.org
https://bib.schema.org/Thesis
`<meta name="description" content="Schema.org is a set of extensible schemas that enables webmasters to embed structured data on their web pages for use by search engines and other applications." />`

This is not helpful for any data consumer that uses the `<meta>` description to provide users with a description in things like snippets (most notably search engines and social media networks), as the current boilerplate description does not describe any page specifically.

A better - and probably relatively easy - method of populating the meta description would be to use the item description.

Person - schema.org
http://schema.org/Person
`<meta name="description" content="A person (alive, dead, undead, or fictional)." />`

inSupportOf - bib.schema.org
https://bib.schema.org/inSupportOf
`<meta name="description" content="Qualification, candidature, degree, application that Thesis supports." />`

A slightly more sophisticated method of populating the value for `content` would prefix this with basic item information.

Person - schema.org
http://schema.org/Person
`<meta name="description" content="The schema.org type Person. A person (alive, dead, undead, or fictional)." />`

inSupportOf - bib.schema.org
https://bib.schema.org/inSupportOf
`<meta name="description" content="The bib.schema.org property inSupportOf. Qualification, candidature, degree, application that Thesis supports." />`

InStock - schema.org
https://schema.org/InStock
`<meta name="description" content="The schema.org enumeration inStock. Indicates that the item is in stock." />`

danbri:
Yeah, this has been a quiet embarrassment for a while. I think a precursor to fixing it is either introducing substructure to our definitions (since some are large, and have hypertext etc.), or at least having a quick python function that pulls out and cleans up the first sentence or paragraph.  Another usecase for this is http://schema.org/docs/tree.jsonld - a JSON view of the type tree - which currently uses crude truncation to get a brief description of each type term.

danbri:
@RichardWallis maybe this sits alongside all the other little TODOs around in-page RDFa?

RichardWallis:
Long descriptions should probably be truncated to say the first paragraph?

Aaranged:
[...]

PR #753, opened 2015-08-27
(#725) HTML <Meta> 

RichardWallis:
Fix (#725) HTML <Meta> built from term name, type, and comment (truncated on a sentence boundary to a sensible length - currently 200 chars)

### DIFF
--- a/api.py
+++ b/api.py
@@ -788,7 +788,24 @@ def read_examples(files):
         parser = parsers.ParseExampleFile(None)
         parser.parse(example_contents)
 
-
+def StripHtmlTags(source):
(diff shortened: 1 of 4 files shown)


danbri:
See also https://github.com/schemaorg/schemaorg/blob/sdo-phobos/sdoapp.py#L189 for similar truncation - we should integrate these into one function _sometime_

danbri:
Should also strip any markup you find too

danbri:
getJsonFriendlyQuoteFreeMarkupFreeInitialSentenceOrSo() ? :)

Issue #188, opened 2014-11-26
Make the telephone property more structured (not just Text)


pierrefar:
http://schema.org/telephone

http://schema.org has machine-readable DataType values for dates, prices, etc. The schema.org/telephone property has an expected value of Text, which is freeform and ambiguous.

It would probably better to have the expected value be stricter, following one of the standard phone formats like E.123 or similar:

http://en.wikipedia.org/wiki/E.123

cp5w:
+1 on this suggestion and surprised its not already part of schema.org

@danbri any movement on this? The thread seems to have gone quiet.

Would `Telephone` be a sublcass of `Thing` > `Intangible` > `StructuredValue` or a `DataType` in its own right (like `URL` is)?

dbs:
@cp5w I don't think we need a schema:Telephone type, we would just want to add "URL" to rangeIncludes for the schema:telephone property and adjust existing examples / add some examples.

+1 for adding URL to the range of schema:telephone so that tel: URIs are explicitly supported and encouraged. (I'm already using them, and occasionally even naughty ones that include a space which then raises an exception in Python rdflib... such is the dirty, dirty nature of the web.)

mfhepp:
+1 for adding URL to the range of schema:telephone so that tel: URIs are explicitly supported

RichardWallis:
+1

If applied, recommend adjusting at least one of the current examples to show this.
eg.

From:
```html
  <span property="telephone">(425) 123-4567</span>
  <a href="mailto:jane-doe@xyz.edu" property="email">
```
To:
```html
  <a href="tel:+1-425-123-4567" property="telephone">(425) 123-4567</a>
  <a href="mailto:jane-doe@xyz.edu" property="email">
```

scor:
+1 on adjusting the example and allowing URL as range for telephone. Created pull request https://github.com/schemaorg/schemaorg/pull/757

unor:
The suggestion to add `URL` to the expected range (for using `tel` URIs) got a few +1s and no objections here. The PR #757 contains the needed changes.

Is there anything that needs to be done before this could be committed?

danbri:
@unor this seems sensible - but I think my comment in the associated issue remains underexplored:

> Do we know what browsers (desktop, mobile) do with phone numbers? let's check as rdfa/microdata markup on links is relevant there...

(also note to self, don't lazily merge the PR as it'll change an old branch)

unor:
[...]

PR #757, opened 2015-09-03
Fix for #188 - Make the telephone property more structured (not just Text)

scor:


danbri:
Do we know what browsers (desktop, mobile) do with phone numbers? let's check as rdfa/microdata markup on links is relevant there...

danbri:
is there a corresponding issue?

Issue #662, opened 2015-07-23
Add SPARQL unittest: values of domainIncludes + rangeIncludes must carry a type


danbri:
This should help us catch easy slips like "rangeIncludes: InteractionCount" instead of "InteractionCounter", since the latter is defined in the schema and is therefore typed. /cc @RichardWallis @vholland 

https://github.com/schemaorg/schemaorg/blob/sdo-ganymede/tests/test_graphs.py#L134

PR #760, opened 2015-09-07
Unit test for domain & range for valid name and type…

RichardWallis:
(#662)

Also fixed couple of domain assersions on Vehicle that replicated from parent class


danbri:
(Waiting for your revisions...)

RichardWallis:
Revised - now ready to be applied

danbri:
Merged!

Issue #687, opened 2015-07-31
New type for barcode images [final task: add an example]


vholland:
It would be useful if there were a child type of http://schema.org/ImageObject for barcodes (including QR codes) so that readers of the data can handle them appropriately.

danbri:
Good idea. Is there any obvious structure here, or just pass it to the barcode tools to figure out?

vholland:
At a minimum, it lets barcode tools do the Right Thing(tm). One thought is to add all of the gtin properties, so you don't even have to scan the barcode.

RLRichardson:
The vast majority of barcodes don't include GTINs, and there's a potentially complicated structure associated with those that carry more than just the GTIN. QR's can be even more complicated as there aren't many structured approaches in place. Dan's suggestion to just pass it along is a good one IMHO

danbri:
(and as I'm writing this, RLRichardson comments before I could even look up his GitHub ID :)

yes - thinking of gtin - https://github.com/zxing/zxing has a list ... probably the main thing to do is figure out how barcode images relate to the various /Product codes. I hear from the GS1 team that the numerous gtin* properties may converge on a single unified representation btw. So hypothetically, a 'gtin' property would apply directly on the image, for example? Our definitions for gtin13 and gtin8 mention EAN (e.g. gtin13 "The GTIN-13 code of the product, or the product to which the offer refers. This is equivalent to 13-digit ISBN codes and EAN UCC-13.").

See also:

 * http://barcode4j.sourceforge.net/
 * https://github.com/woo-j/OkapiBarcode
 * http://zbar.sourceforge.net/
 * http://barcode4j.sourceforge.net/
 * https://github.com/zxing/zxing

vholland:
I added a new type, but no properties in pull request #766.

danbri:
Merged for review/discussion. 

http://sdo-phobos.appspot.com/Barcode

Also note that we have 

<http://schema.org/ticketToken> <http://www.w3.org/2000/01/rdf-schema#comment> "Reference to an asset (e.g., Barcode, QR code image or PDF) usable for entrance." .

... in the schemas nearby. It seems complementary. We could add Barcode or CreativeWork to the range values for ticketToken but I don't think that is essential.

kevinp1:
This would be useful. +1

danbri:
[...]

PR #766, opened 2015-09-11
Issue #687: Added Barcode as a subtype of ImageObject.

vholland:


RichardWallis:
Is there need for a QR Code equivalent?

~Richard

mfhepp:
I would define this property in a way that it is a compatible with QR Codes. The Wikipedia entry for barcode seems to use such a broader notion.

https://en.wikipedia.org/wiki/Barcode

chaals:
tentative +1 to @mfhepp - it's a "visual machine-readable code". 

On the other hand, if there are really only two subtypes and it is often important to distinguish, maybe we should make them explicit.

danbri:
Even qr codes are used in many ways - let's avoid a hardcode taxonomy that will date quickly

vholland:
I updated the definition to "visual machine-readable code such as a barcode or QR code".

danbri:
Merged for review/discussion. Also note that we have 

<http://schema.org/ticketToken> <http://www.w3.org/2000/01/rdf-schema#comment> "Reference to an asset (e.g., Barcode, QR code image or PDF) usable for entrance." .

... in the schemas nearby. It seems complementary. We could add Barcode or CreativeWork to the range values for ticketToken but I don't think that is essential.

Issue #773, opened 2015-09-14
Add a simpler mechanism to specify the platform for EntryPoint targets


vholland:
When authors use http://schema.org/Action, they have two ways to specify the platform where the Action should be executed: http://schema.org/actionApplication or http://schema.org/target with the special prefix "android-app" or "ios-app" as appropriate. (See the examples in http://schema.org/EntryPoint.)

These mechanisms are insufficient as:
- They don't support mobile web apps very well. The author should not have to specify the specific browser app.
- There is no way to distinguish mobile web URLs from URLs intended for a desktop experience.
- Authors often want the abstract notion of a particular OS, not a specific version (e.g. iOS, not iOS 9).
- With [Universal Links in iOS 9](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/AppSearch/UniversalLinks.html), the Action will have a single URL regardless of platform.

To make it easier for developers to specify which platforms support their Action URLs, I propose the following:
- A new property on http://schema.org/EntryPoint called **targetPlatform** which takes one of the following enumeration values:
  - **IOSPlatform**
  - **AndroidPlatform**
  - **WindowsMobilePlatform**
  - **MobileWebPlatform**
  - **DesktopWebPlatform**

Authors who prefer to specify a specific application or OS version can continue to use actionApplication as they do today.

/cc @shankarnat Is there a better term for Windows Mobile?

vholland:
Created pull request #778. Some things to note, which I would like feedback on:

1. http://schema.org/targetPlatform already exists for http://schema.org/APIReference. It does not seem entirely compatible with this proposal, so I added **actionPlatform** instead.

2. Should this be unified with http://schema.org/gamePlatform? If so, I would name the type **ComputingPlatformType** rather than **ActionPlatformType**.

/cc @chaals @ajax-als @tilid @pmika @mfhepp @shankarnat @rvguha @danbri

danbri:
Note: if we go for this, http://schema.org/docs/actions.html will likely need an update too

vholland:
I'd like to squeeze this into sdo-phobos if possible. Comments?

/cc @chaals @ajax-als @tilid @pmika @mfhepp @shankarnat @scor @rvguha @danbri

danbri:
where are we with this folks? PTAL!

tilid:
Hi
Sorry for the late answer
I think it is a good idea to unify this with http://schema.org/gamePlatform. 
[...]

PR #778, opened 2015-09-15
Issue #773: Added actionPlatform to EntryPoint and supporting enumerations.

vholland:


danbri:
How about we add actionPlatform (Text, URL) initially, and work out the enum later?

danbri:
I have added the simplest possible version of this to sdo-phobos for discussion.

Note that I omit:

* changes to examples and docs/actions.html
* enumeration of possible values

... as those would need more discussion than we've seen so far. I'll close the PR but it's linked from the larger issue.

Issue #323, opened 2015-02-05
New property for location a CreativeWork was created/captured


vholland:
Adding a property like recordingLocation would allow us to capture where a CreativeWork was created. Particularly for images and video, it is useful to know the geolocation of the image.

danbri:
See also https://github.com/schemaorg/schemaorg/issues/249 on the vagueness of http://schema.org/contentLocation

vholland:
I had read this as the place where the Thing is. For example, my local library may be the contentLocation of a manifestation of Macbeth. Am I reading that wrong? If so, we can just tighten the definition.

danbri:
@vholland your guess is as good as anyone's here re contentLocation. The ancients have not left us adequate records of their initial vision here, unfortunately. From an unscientifically anecdotal look at a few pages that use it, country seem to be amongst common values for the property. But that makes a kind of sense even if to mean "where is the copy now", in that the bigger a place is always going to have more items associated to it than a smaller place (UK versus Dan's Desk). 

I'm curious if @dbs is seeing any particular interpretation of this property for holdings data in library systems - https://coffeecode.net/archives/271-RDFa-and-schema.org-all-the-library-things.html doesn't mention it explicitly. Dan, any thoughts?

lazaruscorporation:
We already have a "recordedAt" property for CreativeWork which would seem to fulfil the requirement for where the CreativeWork *depicts* (currently "recordedAt" expects schema.org/Event, but I'd suggest that schema.org/Place be added as well). 

So what's missing is a property to specify where an instance of CreativeWork *resides* - of course that would only be of use where the CreativeWork is a unique object (e.g. a piece of VisualArtwork). contentLocation would fit that bill, but I can imagine people easily getting confused between recordedAt and contentLocation.

vholland:
I've been looking at uses of contentLocation and authors use it for either the location of the item or where the item was created, depending on the author's needs and interpretation of the vague description.

My only concern with using "locationRecorded" is people are already using "recordedAt" for the same idea. I hate to break their markup without a compelling reason.

sesuncedu:
There is metaphysics involved here-
A "creative work" does not necessarily reside anywhere.

There are certain stances you can take that would allow you to assign a
spatial region to the creative work "Moby Dick", which would be the same as
the space occupied by every single copy of the work (including copies in
people's heads). Every time a copy was moved, this region would change.
This is rarely useful unless required by other commitments.
[...]

PR #788, opened 2015-09-21
Issue #323: Added locationCreated to CreativeWork.

vholland:


danbri:
Thanks. Fixed merge conflicts (just docs/releases.html) locally.

PR #795, opened 2015-09-25
Replace javascript [more..] option with a CSS only solution

RichardWallis:


danbri:
Thanks, though it would be great if we could figure out a CSS animation for slow reveal too...

Issue #796, opened 2015-09-25
Add an offeredBy property to Offer


RichardWallis:
Follow on from #764 

Inverse of makesOffer

danbri:
Merged - PTAL: http://sdo-phobos.appspot.com/docs/releases.html

danbri:
http://sdo-phobos.appspot.com/offeredBy

danbri:
Fixed in http://schema.org/docs/releases.html#v2.2  - thanks all

Issue #764, opened 2015-09-11
Can we define 'seller' as a subproperty of 'provider' ? (cf noncommercial Offers)


danbri:
* http://schema.org/provider "The service provider, service operator, or service performer; the goods producer. Another party (a seller) may offer those services or goods on behalf of the provider. A provider may also serve as the seller."
* http://schema.org/seller "An entity which offers (sells / leases / lends / loans) the services / goods. A seller may also be a provider."
* seller is currently marked as a subproperty of participant (an Action-oriented property)
* http://schema.org/participant "Other co-agents that participated in the action indirectly. e.g. John wrote a book with *Steve*."

My motivation here is that I still cringe every time I see 'seller' in non-commercial Offer descriptions, notably library holdings data (#733). Using 'provider' would seem more natural. I'd suggest revising 'seller' description to note that non-commercial offers can be described via 'provider'.

See also #424 #620

/cc @RichardWallis @dbs @vholland @chaals @rvguha @pmika  @mfhepp @shankarnat

vholland:
+1 I think authors often do this anyway for the case where seller and provider are the same entity.

akuckartz:
:+1:

RichardWallis:
+1

danbri:
I'd love to know what @dbs thinks about this, as an implementor. The 'seller' variant would remain valid but imho this gives us a path to something that isn't needlessly commercial in its tone.

erimille:
+1

dbs:
@danbri So the included domains of schema:seller currently are:

* BuyAction 
* Demand 
* Flight 
* Offer 
* Order

And the included domains for schema:provider currently are:

* BusTrip 
* CreativeWork 
* Flight 
* Invoice 
* ParcelDelivery 
* Reservation 
* Service 
* TrainTrip

There's not much overlap (just schema:Flight), and schema:provider isn't even valid on schema:Offer (for some definition of "valid").

Are you suggesting that schema:Offer _replace_ schema:seller with schema:provider, which would then have the option of either taking schema:seller as a more commercially defined subproperty when appropriate? This could lead to a subproperty hierarchy of participant -> provider -> seller, given than seller is already a subproperty of participant, but provider is not.

[...]

PR #797, opened 2015-09-25
OfferedBy property & examples

RichardWallis:
Added offeredBy property plus examples and examples for makesOffer.
#796 #764


danbri:
I think we should do this. Would you mind addressing the conflict, @RichardWallis ?

danbri:
thanks!

Issue #818, opened 2015-10-01
Create subtype of ItemList for OfferCatalog


vholland:
When http://schema.org/ItemList was revamped, one of the use cases was a catalog. The second example is for one such use case.

As this is a common use case and seeing adoption, it would be nice to add a separate type for it:

***OfferCatalog***: An OfferCatalog is an ItemList that contains Offers and/or further OfferCatalogs.
***hasOfferCatalog***: Indicates an OfferCatalog listing for this Service or Organization.

vholland:
Created pull request #821.

jvandriel:
Just to make sure I get this right @vholland, do you mean this type would be the preferred type for a product listing page?

vholland:
@jvandriel to be pendantic, it would be the preferred type for offers. For example, Amazon's catalog could be a giant OfferCatalog.

Something strange like all the Products I own would be an odd use of this type.

jvandriel:
_"All the products I own"_ would indeed be a bit strange.   :)

I was more thinking along the lines of product listing pages on ecommerce sites, which generally are modelled from a ```schema.org/Product``` POV and not from an ```schema.org/Offer``` POV.

danbri:
I'll merge this into sdo-phobos to aid with reviewing it - easier to understand in context vs reading config file markup I think.

danbri:
FYI: http://sdo-phobos.appspot.com/OfferCatalog - queued for final review

jvandriel:
I'm struggling to understand what type of page you imagine using ```schema.org/OfferCatalog``` @vholland. Do you something like this: http://www.amazon.com/Catalogs-Directories-Reference-Books/b ?

And why the sudden change from modeling such data from a ```schema.org/Product``` POV to an ```schema.org/Offer``` POV? 

After all, it's goes against what's been promoted over the years and thus isn't something most webmasters are accustomed to do.

danbri:
I've opened an issue https://github.com/schemaorg/schemaorg/issues/827 for tracking final review comments, and have stopped making active changes to the schemas/examples/docs. Keep discussing by all means but we'll use #827 to track final review comments and make a hopefully final round of edits when things settle down (I'd guess we need about a week, but let's see :)

vholland:
There are multiple threads across various pull requests and issues. Trying to tie together:

* Re: Pull request #828: I am fine with using _makesOffer_ to link an Organization to an OfferCatalog.
[...]

Issue #812, opened 2015-09-30
Consider specific textual changes to /Product and /Service to explain their relationship


danbri:
This is a sub-issue of #801 (and related to #411) which led to a number of small clarifying changes around services. 

The purpose of this issue is more specific: to collect, discuss and potentially implement tweaks to the _wording_ of /Service and /Product that better present schema.org's approach and the intended meaning of the relevant terms.

Improved or additional examples are welcomed too.

danbri:
Suggestion from @jvandriel: remove  "a haircut" from /Product (carried over from #801)

danbri:
I've left this issue open after the [v2.2 release](http://schema.org/docs/releases.html). Does anyone have concrete textual changes to propose? I see only the 'remove haircut' suggestion.

jvandriel:
Well, I guess with the silent upgrade of adding ```offers``` to ```schema.org/Service```most of my issues with the overlap between ```schema.org/Product``` & ```schema.org/Service``` have been resolved as one no longer needs to use an MTE (```[Product,Service]```) any longer to describe commercially offered services even though ```schema.org/Product``` _can_ be used for this as well. 

So removing 'a haircut;' suffices for me.

Issue #801, opened 2015-09-27
ProfessionalService and Service have no stated relationship - clarify and better integrate Service into core schemas


danbri:
Currently we have two similarly named types relating to service provision, and no stated relationship between them.

See also sub-issues: #812 (specific wording changes), #411 (serviceArea range)

1.) http://schema.org/ProfessionalService "Provider of professional services."

* Thing > Place > LocalBusiness > ProfessionalService 
* Thing > Organization > LocalBusiness > ProfessionalService

Notes on ProfessionalService:

* This type and its subtypes have no properties defined especially for them (incoming or outgoing)
* Subtypes of ProfessionalService: AccountingService, Attorney, Dentist, Electrician, GeneralContractor, HousePainter, Locksmith, Notary, Plumber, RoofingContractor.
 * These are all subtypes of LocalBusiness, and so (like ProfessionalService itself) have multiple supertypes: Organization and Place

2.) http://schema.org/Service  "A service provided by an organization, e.g. delivery service, print services, etc."

* Thing > Intangible > Service

Notes on Service: 

* uses availableChannel to point to a ServiceChannel (which in turn has properties for online, local business, phone etc. channels of a service)
* uses provider to reference Person or Organization, (alongside 'review' for service reviews, aggregateRating for service ratings etc.) 
* service-specific outgoing properties are: serviceArea (an AdministrativeArea), serviceOutput (a Thing), serviceType (Text).
* has one example, which is of an Order that has an 'orderedItem' which is a Service whose 'description' is "furnace installation".
 * related minor issue: orderedItem property should anticipate Service values.
* Service has 2 incoming properties: issuedThrough (on Permit), and providesService on ServiceChannel.
* Service type was added later for Civic Services. See http://blog.schema.org/2013/08/vocabulary-for-describing-civic-services.html (blog post), http://www.w3.org/wiki/WebSchemas/CivicServices (earlier design discussions).
* http://schema.org/Permit was also part of the Civic Services model. A Permit ("A permit issued by an organization, e.g. a parking pass.") is issuedThrough a Service, and can have temporal and geographic validity, as well as an audience.
 * http://schema.org/GovernmentPermit is a subtype of Permit: "A permit issued by a government agency.", see link for example which expressed that a GovernmentPermit was issuedBy a GovernmentOrganization named "Department of Health and Mental Hygiene", and issuedThrough a GovernmentService named "NYC Food Service Establishment Permit Service"
* Subtypes of Service:
[...]

Issue #134, opened 2014-09-20
properly relate Product and Service (ProductOrService)


elf-pavlik:
http://schema.org/Product
http://schema.org/Service

from reply on public-vocabs thread _http://lists.w3.org/Archives/Public/public-vocabs/2014Sep/0229.html_ by @mfhepp

> I suppose that "Service" was once added for some specific, but maybe only hypothetical use-case by one of the sponsors of schema.org.
> 
> In order to keep the vocabulary clean, however, I suggest to use Product (in the sense of ProductOrService) as long as the distinction between product and service does not really matter for processing the data. Many "products" are nowadays bundles of tangible objects and services anyway, so unless you need specific properties that do not make sense for schema:Product, I would minimize the use of schema:Service.
> 
> Note: In principle, we could make Service a subclass of Product, and/or include a superclass ProductOrService (which was the original name in GoodRelations anyway):
> 
> schema:ProductOrService
>     -> schema:Product
>         -> schema:SomeProducts
>         -> schema:IndividualProduct
>         -> schema:ProductModel
>     -> schema:Service
> 
> This has two caveats, though: First, the more special types SomeProducts, IndividualProduct, and ProductModel are then not available for services any longer. But on the contrary, the notion of individuals, models, and bags of products does not really fit to pure services anyway (or they do not matter that much). So this seems acceptable to me.
> 
> Second, if we want to add specializations for very important types of products and services (like Vehicle/Car; HotelRoom, etc.), we will have to make a strict decision on whether that type is a product or a service. That can lead to inconsistency in the vocabulary over time. For instance, I think that "Taxi" should be better placed beneath schema:Product etc. For me, this is the reason for rather sticking to the "ProductOrService" notion of schema:Product.
### TODO
- [ ] http://schema.org/Product
- [ ] http://schema.org/Service
- [ ] http://schema.org/BroadcastService

danbri:
I see a lot of discussion here. Would any of you care to summarize where you think we are on this issue?

elf-pavlik:
@danbri to lower confusion we can start with adding short notes to Service and Product [as you suggested](http://lists.w3.org/Archives/Public/public-vocabs/2014Sep/0273.html)

>A service provided by an organization, e.g. delivery service,
print services, etc. (Note that a particular delivery or print job can
be modeled as a Product.)

I would suggest adding something matching it to Product

[...]

PR #821, opened 2015-10-01
Issue #818: Added OfferCatalog and hasOfferCatalog.

vholland:


mfhepp:
In general, I think the approach makes sense, but I would suggest to expand the range of makesOffer to include OfferCatalog (and of seeks, too). So makesOffer can either point to a single Offer or to an OfferCatalog.

Also, I think that naming OfferCatalog OfferList would be more consistent and make clear that it is a subtype of ItemList.

Lastly, note that we do only need OfferCatalog if we need to preserve the order of the items, i.e. nextItem, previousItem, or position properties from ListItem. Otherwise, I see no difference to simply  attaching multiple offers to the Organization via makesOffer. In JSON-LD and Microdata, this would be an ordered list in its own right. In RDFa, it would be an unordered list.

But again - we only need this if specifying the order is critical, or if we want to attach a name to the list.

By the way, in both cases I think we could simply combine ItemList with Offer as it stands.

rvguha:
There may be properties that apply to all the offers in an OfferCatalog
that can be specified at a Catalog level.

The term Catalog makes sense, because it corresponds to the Catalogs that
are provided by shops.

guha

realworldobject:
I term Catalog makes sense to me. It could take some of the pressure off the term Collection.

Jeff

jvandriel:
I don't want to nitpick but do you have an example of a possible use case?

First off I have doubts we need this types as there are plenty of methods for chain recurring data to offers, products and services.

Secondly I think @mfhepp has a valid point with: _"we only need this if specifying the order is critical, or if we want to attach a name to the list"_

Most off all because _"I see no difference to simply attaching multiple offers to the Organization via makesOffer."_

And lastly, I tend to disagree with @rvguha's statement: _"The term Catalog makes sense, because it corresponds to the Catalogs that
are provided by shops."_

In all my years of working on ecommerce sites I've never heard any website stakeholder speak about a catalog nor about a catalog of offers. Stakeholders simply speak about product listings and/or service listings.

A catalog reminds me of those things that were popular during print era. In digital environments most simply speak about 'listings', which - at least for me - using  ```schema.org/XXXList``` a much more logical name.

vholland:
An important part is the Catalog may have sections (or sub-catalogs) with Offers as the leaf nodes. Perhaps I should improve the example to be something more like:

```
"@type": "OfferCatalog",
  "itemListElement": [
      {
        "@type": "OfferCatalog",
        "name": "Home cleaning",
        "itemListElement": [
          {
            "@type": "Offer",
            "itemOffered": {
            "@type": "Service",
            "name": "Apartment light cleaning"
            }
          },
         {
            "@type": "Offer",
            "itemOffered": {
            "@type": "Service",
            "name": "House light cleaning up to 2 bedrooms"
         }
       }
       ]
     },
      {
        "@type": "OfferCatalog",
        "name": "Miscellaneous",
        "itemListElement": [
          {
            "@type": "Offer",
            "itemOffered": {
            "@type": "Service",
            "name": "Window washing"
            }
         },
        {
            "@type": "Offer",
            "itemOffered": { 
               "@type": "Service", 
              "name": "Carpet cleaning"
            }
         },
         {
           "@type": "Offer",
           "itemOffered": {
              "@type": "Service",
              "name": "Move in/out cleaning"
         }
      }
    ]
  }
}
```

danbri:
I think that showing the hierarchical aspect does help, yes.

jvandriel:
Thanks for the example, that helps indeed.   :)

Any chance you also have some webpage examples you think this might be interesting for? 

(NOT to be stubborn but simply because I'm curious as in my 6/7 years of working with structured data I've never come across a site/page where I thought: _"mmm, I wish there was a catalog type"_ (not even one that comes close to your newest example).

danbri:
@jvandriel there are probably various pages this kind of data could be stuffed inside, but looking to the future it isn't so far fetched to imagine that the HTML wrapper isn't always entirely necessary. Whether or not this happens I would encourage you not to worry so much about which page it would have to live on, but rather think about whether the vocabulary lines up reasonably well with the kind of information available within a site.

jvandriel:
Thanks for explaining that @danbri, and I agree with _"looking to the future..."_ 

Last question, keeping the here and now in mind, do you think it would be best practice if ecommerce sites start using ```schema.org/OfferCatalog``` for product listing pages or do you consider ```CollectionPage > mainEntity > ItemList``` to be valid markup for this as well?

(and no, not from a Google POV but conceptually. I know I'm the SEO here but I'm not fishing for any handouts. I'm just curious what your view on this is, and what the others participating here think)

danbri:
My belief is that OfferCatalog adds clear value, and that its relationship with other possible idioms will shake out over time.

elf-pavlik:
> In general, I think the approach makes sense, but I would suggest to expand the range of makesOffer to include OfferCatalog (and of seeks, too). So makesOffer can either point to a single Offer or to an OfferCatalog.

@mfhepp what do you think about adding **hasOfferCatalog**? it also looks like now one needs to check two properties *makesOffer* and *hasOfferCatalog* to find possible offers. It looks like OfferCatalog can nest as items more OfferCatalog which may require recursive search for offers...

> An important part is the Catalog may have sections (or sub-catalogs) with Offers as the leaf nodes.


Did someone consider using pattern specified in [LDP Direct Container](http://www.w3.org/TR/ldp/#ldpdc) and similar one discussed in [Hydra Collection Design](https://www.w3.org/community/hydra/wiki/Collection_Design)?

Since none of properties uses rdfs:range and many accept schema:Role, why not also allow in similar way schema:ItemList for all the properties?

elf-pavlik:
I also have impression that still some confusion may exist around Product vs. Service. Looking at @vholland example in https://github.com/schemaorg/schemaorg/pull/821#issuecomment-145062059

See also: https://github.com/schemaorg/schemaorg/issues/812 && https://github.com/schemaorg/schemaorg/issues/134 && http://lists.w3.org/Archives/Public/public-vocabs/2014Sep/0273.html

mfhepp:
I can live with the name OfferCatalog and the general pattern, but I strongly suggest to rather expand the range of makesOffer to include OfferCatalog rather than introducing hasOfferCatalog, or , ate least, recommend hasOfferCatalog for Service only and use makesOffer for Organizations -> makesOffer -> OfferCatalog.

Otherwise, linking from the Organization or Person who makes all those offers is cumbersome - you would have to include many offeredBy statements with a link to the offering entity.

makesOffer would be more intuitive:

So here is my proposal:

1. Expand the range of makesOffer to include OfferCatalog
2. Remove Organization from the domain of hasOfferCatalog and update the text to reflect that this is just for services and that for organizations, makesOffer is the property to use.
3. Update the examples.

Minor change, but I think worth the effort

By the way, I now see value in the OfferCatalog type, because it may pave the way for a new JSON-LD-based product feed format.

Very minor: We should have a clear understanding of the difference between OfferCatalog and AggregateOffer. My assumption is that AggregateOffer is for multiple offers of the same product (or very similar ones) by different vendors , while OfferCatalog is for different items offered by the sane vendor.

Last point: Do we also need ProductCatalog for lists of products (ProductModels) without offers?

elf-pavlik:
> ate least, recommend hasOfferCatalog for Service only 
...
Last point: Do we also need ProductCatalog for lists of products (ProductModels) without offers?

Why OfferCatalog than and not ServiceCatalog?

This would also address https://github.com/schemaorg/schemaorg/issues/801#issuecomment-143840249

LocalBusiness - hasServiceCatalog -> ServiceCatalog

mfhepp:
@danbri @vholland I just submitted a PR for my changes as proposed above:

https://github.com/schemaorg/schemaorg/pull/828

I think that would do the trick.

As for @elf-pavlik: The OfferCatalog holds offers of services, not services. So OfferCatalog is fine.

We might need a ProductCatalog for ProductModels without offers (e.g. on manufacturers' pages), but not ServiceCatalog, since we do not use the ProductModel notion for services (it does not fit there very well - in practice).

danbri:
Thanks @mfhepp, as I mentioned in #818 let's hold off on further active changes to give people time to review a stable target. But it helps to have this tweak as a pull request - cheers :)

vholland:
@mfhepp re: AggregateOffer, yes, exactly. AggregateOffer is for multiple offers of the same product. It was the inability to express multiple products/services being offered.

vholland:
Ooops I missed a piece.

I am fine with expanding _makesOffer_ instead of adding _hasOfferCatalog_.

I will be traveling soon, but can create a pull request Monday if necessary.

danbri:
+1 for recycling makesOffer - I'll note this in #827

mfhepp:
Thanks  but note that we should not use makesOffer for linking from Service to OfferCatalog. If you want to maintain that pattern, we need hasOfferCatalog.

But you have to decide whether that route is needed or whether makesOffer or offeredBy with a link to the Organization or Person are sufficient.

mfhepp:
@vholland The pull request is already there. If agreed, we could add an example for the Organization -> makesOffer -> OfferCatalog route, but I would want to wait with that until the model is finalized.

danbri:
Ah ok, misread you. Sounds like we're converging - let's pick this up after the weekend!

mfhepp:
Yes, on a more abstract level, I accept that services may require different modeling patterns in schema.org in some cases - as long as we agree that as soon as we are talking of commercial aspects like prices, Offer remains the core type.

twamarc:
any reason we prefer ```hasOfferCatalog``` rather than ```offerCatalog```?

danbri:
Try saying "offerCatalog" and "OfferCatalog" on the phone to someone - it can be very confusing since we have a type of that name - sdo-phobos.appspot.com/OfferCatalog

Schema.org has a few cases of a type + property with same name (except for case) but we are trying not to add more...

elf-pavlik:
I still find it confusing in *ACME Home Cleaning offers a variety of services in Massachusetts* example to have:
Service - hasOfferCatalog -> OfferCatalog - itemListElement -> Offer -> itemOffered -> Service

In case where *OfferCatalog* would appear as root element how does one specify the provider? hasOfferCatalog doesn't have a reverse property.

Why not to just start from Organization

Organization - makesOffer -> OfferCatalog - itemListElement -> Offer -> itemOffered -> Service

Or make OfferCatalog or root element and go from here to Organization following offeredBy property

Can even in ACME Home Cleaning example, the top level Service `"serviceType": "Home cleaning"` appear as value of itemOffered in the Offer? If not, I have impression that Service type here gets used in two different ways and some confusion around it still exists.

mfhepp:
@elf-pavlik we may argue about the name hasOfferCatalog, but I think it makes sense to have a property that links from a generic Service description to the individual services. We could name it hasServicesCatalog or servicesDetails:

Service -> servicesDetails ->OfferCatalog - itemListElement -> Offer -> itemOffered -> Service

I assume the hasOfferCatalog name was inspired by the original usage for product and services offers.

elf-pavlik:
If *makesOffer* and *offeredBy* will accept OfferCatalog, why not also extend rangeIncludes of [schema:offers](http://sdo-phobos.appspot.com/offers) - which already happen to have plural name :question: Otherwise one will need to use *offers* to link to individual offers from Service and *hasOfferCatalog* to link to the collection - catalog. 

Once again I can imagine often a case where OfferCatalog might appear as root on the page, if *hasOfferCatalog* goes through #828 helps a lot by hinting use of *offeredBy*.

radusi:
@elf-pavlik : If OfferCatalog would be accepted by makesOffer, this could allow mark-up for products with variants, on the same Product page. Is this one benefit you were thinking ? Because it would be wonderful, IMHO.

elf-pavlik:
@radusi I just don't like idea of creating a new properties with the same meaning but which accept ItemList instead of simply multiple Things. Can you maybe write an example where you need OfferCatalog to include multiple product variants on the same page, and can't do it with multiple Offers?

vholland:
@elf-pavlik listing multiple Offers loses any sense of hierarchy or categorization of the Offers. For example, it is useful to know a cleaning service offers home vs commercial cleaning without knowing the specific prices of building size or whether carpet cleaning is included.

We could achieve the same result with serviceType, but authors are often bad at making that a reliable signal. If their services database already includes some hierarchy, it is easier to let them express that as a seperate sub-catalog.

elf-pavlik:
Does the whole OfferCatalog have to appear on a single web page or it also provides a way to split it across multiple web pages?

mfhepp:
@elf-pavlik @vholland I am against including OfferCatalog to the range of schema:offers. schema:offers links from products to offers for this product and has a very specific meaning. Its inverse is itemOffered. This is a different branch of schema.org. makesOffer and offeredBy link between an agent and its Offers. itemOffered and offers link between Offers and Products (or Services). Let's not dilute that.

@radusi I would at this point not recommend OfferCatalog for linking from products to their variants.

jvandriel:
_"makesOffer and offeredBy link between an agent and its Offers. itemOffered and offers link between Offers and Products (or Services). Let's not dilute that."_

+1

_"I would at this point not recommend OfferCatalog for linking from products to their variants."_

Is there a method for 'grouping' nearly similar products?
(eg, products of which only the color differs, or products that are the same but that have different gtin's due to the fact they come from different countries/suppliers).

The reason I ask is because many ecommerce sites have a lot of nearly identical products for which the difference often is limited to a single property value (eg, iPhone 6 Gold vs iPhone 6 Silver). 

Reason for a lot of sites to publish things like reviews multiple times for different products, causing there to be tons of duplicate content (which is bad for search) yet providing a lot of value to visitors of a site. After all, the color of a product is hardly a reason why reviews shouldn't be the same for color variants, and makes it easier for people to find relevant reviews without having to look at a ton of different product variants.

A duplicate content problem I think could be resolved if we introduce something like a ```schema.org/ProductFamily```, which could be a list of nearly identical products that share the same reviews (or questions, or descriptions, etc).

radusi:
I think @jvandriel thought at a page similar to this one http://www.amazon.co.uk/All-New-Kindle-Paperwhite-Resolution-Display/dp/B00QJDO0QC/.

elf-pavlik:
> I am against including OfferCatalog to the range of schema:offers. schema:offers links from products to offers for this product and has a very specific meaning. Its inverse is itemOffered. This is a different branch of schema.org. makesOffer and offeredBy link between an agent and its Offers. itemOffered and offers link between Offers and Products (or Services). Let's not dilute that.

I have impression that currently [schema:Service](http://sdo-phobos.appspot.com/Service) gets used in two different ways:

1) as Agent/Organization (Agency?) which proposed [schema:hasOfferCatalog](http://sdo-phobos.appspot.com/hasOfferCatalog) suggests
(Organization | Service) - hasOfferCatalog ->OfferCatalog

2) as Product which the other end of the chain suggests
OfferCatalog - itemListElement -> Offer -> itemOffered -> Service

To my understanding aims at first way of using it, as sort of Agent(Agency?), which IMO could get handled by Organization and makesOffer.

> listing multiple Offers loses any sense of hierarchy or categorization of the Offers. For example, it is useful to know a cleaning service offers home vs commercial cleaning without knowing the specific prices of building size or whether carpet cleaning is included.

Does need for such hierarchy only makes sense for Service (used as Product or Agent), or also to Product?

I really think we currently have issue here with Service sometimes wearing Organization hat and sometimes Product hat :disappointed:

Issue #819, opened 2015-10-01
Provide a way to specify a location is temporarily open/closed


vholland:
For seasonal businesses, it is difficult to use http://schema.org/OpeningHoursSpecification to state the location is open/closed for an extended period of time. For example, a campground may be closed in Winter or a restaurant may be closed for renovations.

I propose adding a new property to http://schema.org/OpeningHoursSpecification, **openingHoursStatus**. The possible values are:
- OpeningHoursClosed
- OpeningHoursOpen

```
<script type="application/ld+json">
{
  "@context": "http://schema.org",
  "@type": "Campground",
  "name": "Middle of Nowhere Campground",
  "openingHoursSpecification":
  {
    "@type": "OpeningHoursSpecification",
    "validFrom": "2013-10-01",
    "validThrough": "2014-05-01",
    "openingHoursStatus": "http://schema.org/OpeningHoursClosed"
  }
}
</script>
```

vholland:
Created pull request #824.

mfhepp:
I am against this quick fix. First, the opening hours issue is complex, so adding quick fixes to immediate problems may not last very long. Second, we could achieve the same effect with less elements. Two suggestions:

1. We use the convention from GoodRelations that opens = 00:00:00 + closes = 00:00:00 means closed all day while opens = 00:00:00 and closes 23:59:59 or closes 24:00:00 means open all day. Then, a longer closing period would simply be modeled as follows:

    <div itemscope itemtype="http://schema.org/Store">
      <h1 itemprop="name">Middle of Nowhere Foods</h1>
      <h2>Opening hours</h2>
        <p>Normally open <time itemprop="openingHours" datetime="Mo,Tu,We,Th,Fr,Sa,Su 09:00-14:00">daily 9am-2pm</time></p>
        <p>Closed from </p>
        <div itemprop="openingHoursSpecification" itemscope itemtype="http://schema.org/OpeningHoursSpecification">
          <span itemprop="validFrom" content="2015-12-01">December 1, 2015</span> through
          <span itemprop="validThrough" content="2015-12-31">December 31, 2015</span>
          <meta itemprop="opens" content="00:00">
          <meta itemprop="closes" content="00:00">
        </div>
    </div>

I think that is the best approach, for it simply means amending the documentation by this example.

[...]

PR #824, opened 2015-10-01
Issue #819: Added openingHoursStatus and corresponding enumeration.

vholland:


mfhepp:
I am against this pull, see #819 for my arguments.

Issue #704, opened 2015-08-06
auto.schema.org review tracking 


danbri:
This issue tracks final review comments on the auto.schema.org extension (see parent issue, #509).

Known unresolved issues at time of publication:
- consider renaming specialUsage as vehicleSpecialUsage, or broaden its definition - #705
- consider renaming seatingCapacity as vehicleSeatingCapacity or broaden its definition - #706

mfhepp:
Solved via https://github.com/schemaorg/schemaorg/pull/825

seatingCapacity and specialUsage are now defined in a way that fits other types of objects than vehicles.
seatingCapacity may also be useful for buildings, boats, classrooms, etc.
specialUsage may also be useful for other types of products, namely real estate objects.

We may want to move the definition of these elements to schema.org core once they are added to other types. For the moment, I suggest to keep them in the auto extension and expand their range in a second step.

-----------------------------------
martin hepp  http://www.heppnetz.de
mhepp@computer.org          @mfhepp

mfhepp:
It seems that the current status is that specialUsage is defined in the extension and seatingCapacity has been renamed vehicleSeatingCapacity and is defined in the core.

So what is left to be done is

* https://lists.w3.org/Archives/Public/public-gao/2015Nov/0006.html
* https://github.com/schemaorg/schemaorg/issues/899
* https://github.com/schemaorg/schemaorg/issues/898

Then we should be all set for the final release of the auto extension and this issue should be closed.

danbri:
specialUsage needs some attention - see https://github.com/schemaorg/schemaorg/issues/705 (and #898)

PR #825, opened 2015-10-02
Auto fix: Broaded definition of seatingCapacity and specialUsage

mfhepp:
This solves #704.

seatingCapacity and specialUsage are now defined in a way that fits other types of objects than vehicles.
seatingCapacity may also be useful for buildings, boats, classrooms, etc.
specialUsage may also be useful for other types of products, namely real estate objects.

We may want to move the definition of these elements to schema.org core once they are added to other types. For the moment, I suggest to keep them in the auto extension and expand their range in a second step.


danbri:
Thanks - I agree

Issue #733, opened 2015-08-12
Add an example showing how availability of a Book at a Library is modelled


danbri:
Raw materials:

* https://www.w3.org/community/schemabibex/wiki/Holdings_via_Offer
* http://catalog.mvlc.org/eg/opac/record/1003645?query=linear%20algebra;qtype=keyword;locg=1
* http://acorn.biblio.org/eg/opac/record/1680040?query=linear%20algebra;qtype=keyword;locg=1

/cc @dbs @RichardWallis @erimille @vholland 

Consider illustrating use of http://schema.org/parentOrganization http://schema.org/department

dbs:
I'm (perhaps obviously) happy to take responsibility for this task...

danbri:
@dbs :) I've just talked Eric Miller through installing AppEngine on his laptop, but if you can take a lead here that would be great. 

AFAIK "seller" is still the best practice for relating the Offer to the Library, rather than say "provider". Nobody has shouted about this being terrible but I'm open to trying to find less $-oriented language if it is proving an obstacle. 

Let's work out some ideal examples here in the issue tracker to make sure they're capturing the use cases everyone cares about, rather than getting distracted by redesigning vocab.

danbri:
@dbs - did you get a chance to look at this yet?

dbs:
@danbri My apologies, this slipped off my radar; I'll pull some examples together tonight.

danbri:
Thanks @dbs!

dbs:
So let's start with a simplified example of the offer made of a DVD from the Music Resource Centre (MRC), which is part of the Laurentian University system. This is based on https://laurentian.concat.ca/eg/opac/record/2845929:

    <https://laurentian.concat.ca/eg/opac/record/2845929#schemarecord> a schema:Movie,
            schema:Product;
        schema:about "Composers > Austria > Biography."@en-ca,
            "Documentary films."@en-ca,
            "Haydn, Joseph, 1732-1809."@en-ca;
        schema:datePublished "[2011]"@en-ca;
        schema:name "In search of Haydn [videorecording] / Phil Grabsky Films.com & Seventh Art Productions in association with Sky Art HD ; filmed, written and directed by Phil Grabsky."@en-ca;
        schema:offers [ a schema:Offer;
                schema:availability schema:InStock;
                schema:availableAtOrFrom "Circulation"@en-ca;
                schema:businessFunction <http://purl.org/goodrelations/v1#LeaseOut>;
[...]

PR #826, opened 2015-10-02
733 examples

RichardWallis:
Fix to #733 - library examples


RichardWallis:
Examples based on those provided by @dbs

danbri:
Thanks @dbs , @RichardWallis 

I have tweaked the examples a bit, and merging. Could do with a final checkover (and the library example converted from Turtle to JSON).

PR #828, opened 2015-10-02
Allow makesOffer for OfferCatalog

mfhepp:
As suggested, I think we should use makesOffer for linking an Organization to a OfferCatalog.
This pull request implements that proposal.
We would then use hasOfferCatalog only for linking from a Service to its OfferCatalog and makesOffer for Organizations and Persons linking to their offers.

Advantages:

1. We support Person and Organization.
2. makesOffer and offeredBy are then symmetrical, while in the current proposal, you can use offeredBy with OfferCatalog and Offer and makesOffer only with Offer.
3. You need only one property to link from the Organization to the OfferCatalog and implicitly add the makesOffer information,

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -1503,8 +1503,7 @@
 
     <div typeof="rdf:Property" resource="http://schema.org/hasOfferCatalog">
       <span class="h" property="rdfs:label">hasOfferCatalog</span>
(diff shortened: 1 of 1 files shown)


jvandriel:
+1

RichardWallis:
+1

Issue #829, opened 2015-10-02
Mechanism for Compound Prices


mfhepp:
In GoodRelations and schema.org, we miss a mechanism for modeling prices that consist of multiple components, like

car: 100 USD per week + 0.10 USD per mile
apartment: 100 USD per week + 50 $ cleaning fee per stay

The current model leaves open whether multiple UnitPriceSpecifications are alternatives or meant in combination.

Pull-request #830 addresses this issue.

danbri:
Can you give some example sites (ideally using schema.org) where this vocab could improve things?

brewerdigital:
@danbri @mfhepp here are two examples where a mechanism for compound prices could be helpful; rental fees are a standard charge at nearly all vacation rental websites. 
http://www.vrbo.com/324460#rates
http://www.homeaway.com/vacation-rental/p333759#ratesAnchor
Additional information about rental rates > Cleaning Fee $200

danbri:
@mfhepp et al - do you have a sense for how sites have dealt with the absence of this construction so far? Have they been attempting to partially mark it up somehow, or just omitting?

thadguidry:
@danbri @mfhepp so... since this is particular issue has some work merged, does the discussion on this issue then flow further into possibly us adding a concrete creditCardFees property on PriceSpecification #908 ?  Its a common thing and the upcoming W3C Payment Request API work has this at a different level at the Details (line items in an order) which is equivalent to PriceSpecification with Composite pricing.

@RichardWallis does FIBO have an opinion on potential creditCardFees property ?

RichardWallis:
@thadguidry I have not been involved in FIBO discussions recently, but I don't think CreditCards using Schema.org are a current focus for them.

Looking at the current [CreditCard](http://schema.org/CreditCard) spec, the [feesAndCommissionsSpecification](http://schema.org/feesAndCommissionsSpecification) property inherited from FinancialProduct comes close to what is being looked for here.

Not sure how this could be related via a payment type option to a PriceSpecification.

Seems a bit ironic to be discussing this at the time the European Union have made the applying of surcharges for credit card transactions illegal. ;-)

PR #830, opened 2015-10-02
added CompoundPriceSpecification

mfhepp:
This pull requests addresses #829.

It fixes a major inconsistency in the GoodRelations pricing model and is needed for hotels and rental cars.


rvguha:
This might break some of the applications that currently use schema.org
price specification.

Lets please get their ok before making these changes

guha

mfhepp:
Dear Guha:
Of course, it is good to check that it won't break existing apps, but I am pretty sure that there is almost no markup in the wild to far that properly uses multiple price components in the existing form. So the focus of a cross-check should be that an app that currently understands a single UnitPriceSpecification will also be able to consider multiple components attached via the new CompoundPriceSpecification type.

Martin

rvguha:
Breaking usage of existing markup is not the only problem. Another problem
that occurs is that new (and some old) publishers start switching to the
new way of expressing things, which old apps cannot (yet) consume.

guha

vholland:
Do you have examples of this new schema? It is complex, so authors are probably not going to use it correctly with only the definition to go by.

mfhepp:
I have closed and updated this pull request, see https://github.com/schemaorg/schemaorg/pull/994 for the correct one.

PR #835, opened 2015-10-06
Fix the Microdata in the example translation

dsalo:
property="itemprop" shows up in the Microdata example for both author and translator, instead of itemprop="author" and itemprop="translator" respectively


danbri:
Something's up with this pull request; I think it is comparing all of sdo-phobos with our master branch, which is something I'm doing now in release prep. I'll close this PR, but have noted the examples glitch as #871 - thanks.

PR #836, opened 2015-10-06
PropertyValue example fixes (pre-markup ex. had Microdata; alt attributes were missing)

unor:
* removed Microdata from two PRE-MARKUP examples
* added empty `alt` attribute to all `img` elements

(@danbri: Should I create an issue for this? I assumed not because these are only example markup changes.)

### DIFF
--- a/data/sdo-property-value-examples.txt
+++ b/data/sdo-property-value-examples.txt
@@ -3,7 +3,7 @@ TYPES: #exif ImageObject, PropertyValue
 PRE-MARKUP:
 
 <h2>Beach in Mexico</h2>
-<img src="mexico-beach.jpg" />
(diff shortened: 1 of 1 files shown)


mfhepp:
+1 !
Thanks for spotting and fixing this. Should go into sdo-phobos if posssible.

danbri:
@unor - I see that I just merged this by mistake into the old sdo-phobos branch. Would you mind re-targetting it at sdo-deimos? Otherwise I can figure it out from my side but I believe Github makes such things easier from the PR proposer.

unor:
@danbri: The new one is https://github.com/schemaorg/schemaorg/pull/1037

PR #839, opened 2015-10-11
Typo in sharedContent description

unor:
"posting" instead of "psting"


danbri:
I fixed this manually for some reason, but thanks for the pull request - much appreciated.

PR #843, opened 2015-10-13
Issue #818: Changed hasOfferCatalog to serviceDetails and added OfferCatalog to the range for makesOffer.

vholland:


mfhepp:
+1 looks good to me.

danbri:
@rvguha dissents - see #818 for discussion (and let's try to keep high level discussion in the issues not the pull requests).

PR #850, opened 2015-10-24
fix a typo: is 'psting', should be 'posting'

michalroszka:
Context: _A CreativeWork such as an image, video, or audio clip shared as part of this psting._

Changed files:

 * data/releases/2.1/all-layers.nq
 * data/releases/2.1/core.nq
 * data/releases/2.1/schema-all.html
 * data/releases/2.1/schema.nt
 * data/releases/2.1/schema.rdfa
 * data/schema.rdfa


danbri:
Thanks for catching this, and @chaals for merging. 

In general btw we should fix things in the live version but not re-write our history of timestamped releases. For such a simple typo this isn't a problem but generally everything under data/releases/x.y/* should be frozen upon publication. Otherwise it wouldn't be clear where to stop w.r.t. updating mistakes found in these, and they lose their value as a record of the evolution of these schemas.

PR #855, opened 2015-10-29
Added fibo extension files 

Dataliberate:
Addid fibo as a possible extension plus fibo1.0.rdfa including a placeholder testProperty and TestType


RichardWallis:
Added to wrong repository in error :-(

Issue #636, opened 2015-07-08
JSON-LD example for EventCancelled uses invalid date format


Aaranged:
On:
http://schema.org/EventCancelled

The JSON-LD example contains this declaration:
`"startDate": "Sat Sep 14"`

`startDate`, of course, requires a date in ISO 8601 format, which this is not.

Note that both the RDFa and microdata examples provide the correct value:
`2013-09-14T21:30`

mfhepp:
Will handle that

RichardWallis:
Now part of #865

PR #865, opened 2015-11-03
Fix to invalid date format in JSON-LD example for Event, EventCancelled, & SoldOut

RichardWallis:
Fix to invalid date format in JSON-LD example for Event, EventCancelled, & SoldOut includes fix for #636


danbri:
Thanks, this is a screamingly obvious and useful fix so let's not stand on ceremony. Merging. 

/cc FYI @pmika @scor  @chaals @tilid @ajax-als @shankarnat @mfhepp @vholland @rvguha

Issue #866, opened 2015-11-03
bug in https://schema.org/RentalCarReservation examples


mfhepp:
The JSON-LD example for https://schema.org/RentalCarReservation uses a non-existent property operatingCompany. This should be replaced by schema:provider and attached to the schema:RentalCarReservation, not the schema:Car.

I attach a fixed example.

[fixed-reservation-example.txt](https://github.com/schemaorg/schemaorg/files/25985/fixed-reservation-example.txt)

mfhepp:
For proper acknowledgment: This was spotted by Jimmy Phan, see

https://lists.w3.org/Archives/Public/public-vocabs/2015Nov/0002.html

danbri:
Thanks Jimm, Martin, Richard :) queued for release asap.

PR #867, opened 2015-11-03
Fix to RentalCarReservation Example

RichardWallis:
Fix to RentalCarReservation Example - Issue (#866)

(_The JSON-LD example for https://schema.org/RentalCarReservation uses a non-existent property operatingCompany. This should be replaced by schema:provider and attached to the schema:RentalCarReservation, not the schema:Car._)

Thanks to @mfhepp and Jimmy Phan.


danbri:
Silly bug! Merging. Thanks!

PR #877, opened 2015-11-06
Release cleanup.

danbri:
Last notes before we realised that there are problems with
the extensions, and that these notes deal inadequately with that topic.
TODO:
 * testing extensions
 * dealing with screwups: update master and merge back into
   the release branch, for things < 24h after publication.


danbri:
Tidying some notes into the release.

PR #890, opened 2015-11-13
Fix wrong inverseOf for containsPlace

teohhanhui:


unor:
+1 (same fix like in https://github.com/schemaorg/schemaorg/pull/983)

Issue #899, opened 2015-11-18
http://schema.org/vehicleSeatingCapacity returns a 500 status code


mfhepp:
See below. The same problem exists for http://auto.schema.org/vehicleSeatingCapacity.

```
$ curl -i http://schema.org/vehicleSeatingCapacity
HTTP/1.1 500 Internal Server Error
Content-Type: text/html; charset=UTF-8
Date: Wed, 18 Nov 2015 07:51:53 GMT
Server: Google Frontend
Content-Length: 228

<html>
 <head>
  <title>500 Internal Server Error</title>
 </head>
 <body>
  <h1>500 Internal Server Error</h1>
  The server has either erred or is incapable of performing the requested operation.<br /><br />
```

PR #901, opened 2015-11-19
Fix899

RichardWallis:
Fix to #899 - Fix to ShortenOnSentence() which failed on comment string over max length which did not end with '.'.  For completeness, added the missing '.' to definition of vehicleSeatingCapacity property.

Also  Disabled _siteDebug call on host schema.org unless PRODSITEDEBUG environment variable set to 'True'. Still operates as before on all other hosts.

### DIFF
--- a/api.py
+++ b/api.py
@@ -862,15 +862,27 @@ def StripHtmlTags(source):
 
 def ShortenOnSentence(source,lengthHint=250):
     if len(source) > lengthHint:
+        source = source.strip()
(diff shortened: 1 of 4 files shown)


danbri:
Ok let's get this released 'early access'...

RichardWallis:
Yup - you going to do that - and update deimos to bring it into step?

~Richard.

danbri:
I will! Thanks for the fix.

Issue #881, opened 2015-11-07
Schema Review microdata example doesn't use reviewBody


inetbiz:
I was looking at https://schema.org/Review today. I noticed the microdata example uses description instead of reviewBody.

unor:
I think you’re right, `reviewBody` should be used there instead of `description`.

→ https://github.com/schemaorg/schemaorg/issues/909

dandv:
:+1:, can we update this please?

PR #909, opened 2015-11-28
Review example: use 'reviewBody' instead of 'description'

unor:
Reported by @inetbiz in https://github.com/schemaorg/schemaorg/issues/881

(Also fixed two occurences of a `span` element having a `content` attribute in Microdata, which is not valid; replaced with `meta`.)


inetbiz:
@danbri can this PR be fast-tracked? I've been talking to a developer. They won't believe me because the example microdata is still flawed using description instead of **reviewBody**.

PR #910, opened 2015-11-28
ExhibitionEvent example: fixes and cleanup

unor:
* `property` instead of `itempropertyprop`
* RDFa missed `datetime` value in the `time` element with the `endDate` property

* added `datetime` attributes to `time` elements in PRE-MARKUP (invalid otherwise)
* removed `class` attributes (the class `nowrap` doesn’t seem to add anything needed for understanding the markup)


frankieroberto:
:+1: looks good to me, thanks.

danbri:
Thanks again :)

Issue #898, opened 2015-11-18
https://schema.org/CarUsageType and instances should go to the auto.schema.org extension


mfhepp:
I think that 

https://schema.org/CarUsageType and its predefined instances should be moved from schema.org core to the auto.schema.org extension, because the only property that uses this value set type,

https://auto.schema.org/specialUsage

is also defined in the extension, not in the core.

danbri:
Thanks. However please see #705 "auto.schema.org extension: consider renaming specialUsage as vehicleSpecialUsage, or broaden its definition"; closing in favour of that existing issue.

mfhepp:
This is IMO a different issue. If we decide to rename and broaden specialUsage, then https://schema.org/CarUsageType and its predefined instances should be moved from schema.org core to the auto.schema.org extension AND specialUsage should be added to the core. 

Currently, it is the other way round.

mfhepp:
Since we also have itemCondition for generic product information about special usage, I think that simply  moving https://schema.org/CarUsageType and its predefined instances  to the auto.schema.org extension and maybe renaming http://auto.schema.org/specialUsage to vehicleSpecialUsage (tbd) will be the better route.

danbri:
Let me rephase: the current approach, which uses a generic phrase for a very car-specific concept, isn't viable. Either the definition or the name has to change. Once we have resolved that question, then we are in a sensible position to decide whether terms are core or extension. I think we agree on where the terms should best live (i.e. in auto extension, with the property renamed as above).

mfhepp:
@danbri I agree, so let's

1. rename http://auto.schema.org/specialUsage to http://auto.schema.org/vehicleSpecialUsage in the extension AND
2. move  http://schema.org/CarUsageType and its predefined instances to auto.schema.org AND
3. double-check that specialUsage is changed to vehicleSpecialUsage in all examples.

danbri:
Sounds good to me. We can always collect xyzSpecialUsage properties into a common superproperty in the future, if some appear.

BTW from many years ago we also have http://schema.org/specialCommitments whose name is similarly too generic for its job-centric definition; maybe we can address that in future, but I'm glad we're not compounding the issue.

twamarc:
@mfhepp @danbri this is an interesting and my solve some issues for other extensions as well. I will try to how it looks like once applied to vague/too general terms used fro medical/health extension.

trypuz:
[...]

Issue #705, opened 2015-08-06
auto.schema.org extension: consider renaming specialUsage as vehicleSpecialUsage, or broaden its definition


danbri:

danbri:
Implemented in 3.0

PR #922, opened 2015-12-08
Auto final fixes

W3C-GAO:
The following issues are addressed by that pull request:
- #704 (master issue)
- #899
- #898
- https://lists.w3.org/Archives/Public/public-gao/2015Nov/0006.html

### DIFF
--- a/data/ext/auto/auto.rdfa
+++ b/data/ext/auto/auto.rdfa
@@ -204,9 +204,9 @@
         <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/QuantitativeValue">QuantitativeValue</a></span>
         <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#Automotive_Ontology_Working_Group">GAO</a></span>
     </div>
-    <div typeof="rdf:Property" resource="http://schema.org/specialUsage">
+    <div typeof="rdf:Property" resource="http://schema.org/vehicleSpecialUsage">
         <link property="http://schema.org/isPartOf" href="http://auto.schema.org" />
-        <span class="h" property="rdfs:label">specialUsage</span>
+        <span class="h" property="rdfs:label">vehicleSpecialUsage</span>
         <span property="rdfs:comment">Indicates whether the object has been used for special purposes, like vehicles that have been used for commercial rental, driving school, or as a taxi. The legislation in some countries may require this information to be revealed when offering the item for sale.
     </span>
         <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Vehicle">Vehicle</a></span>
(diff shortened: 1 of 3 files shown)


mfhepp:
Thanks - a few questions:

1. Is this version somewhere online on a demo site (on appspot)?
2. From the diff, I think that vehicleSpecialUsage lacks a comment (not 100 % sure).
3. Did you double-check the examples for mentions of the elements removed from core or renamed in the extension?

W3C-GAO:
@mfhepp
Ad 1. No
Ad 2. vehicleSpecialUsage in the core has a comment. vehicleSpecialUsage in auto extension does not have a comment.
Ad 3. Yes

mfhepp:
@RichardWallis @danbri I think this extension is the first to modify the range of a property from core; can you have a look whether this is fine? 

If I understand correctly, the proposal is to have just Text as the range in core and Text OR http://auto.schema.org/CarUsageType.

It might be simpler to move vehicleSpecialUsage to the extension completely and removing it from core.

A counter argument is that in the past, we wanted all properties that are useful for typical car listing sites in the core and externalize only those relevant mostly for manufacturer pages.

mfhepp:
@W3C-GAO It would be nice to have an appspot version available for review. It should not be a big issue to set up and will be much easier for reviewers than looking at the pure code.

RichardWallis:
If specialUsage is defined as in the core with a range of Text,  you should
only need to define a range of CarUsageType in its definition in the
extension to achieve that objective.  The site will sum them together at
run time.

~Richard

trypuz:
@RichardWallis, there is no specialUsage in the core (we had it in the auto extension before but it was changed to vehicleSpecialUsage). So now, we have vehicleSpecialUsage both in the core and in the extension. vehicleSpecialUsage in the core has just Text as the range; in the extension we only added a range of CarUsageType in its definition.

RichardWallis:
OK, sorry I was looking at the pre name change version.

Principle is the same however.

If vehicleSpecialUsage is defined as in the core with a range of Text,  you
should only need to define a range of CarUsageType in its definition in the
extension to achieve that objective.  The site will sum them together at
run time.


Which as you say is what is in pull request #922
<https://github.com/schemaorg/schemaorg/pull/922>

~Richard.

trypuz:
We created a demo site on appspot of this version of auto extension. Here it is: http://sdo-gao.appspot.com

danbri:
Note that this update wipes out specialUsage; normally we leave the old definition in place and indicate the newer with 'supersededBy'. In this case, since it was in the stated 'preview' release of the extension, I think we can get away with it as usage levels of specialUsage are likely very low. If it turns out to have been used significantly anywhere we can always add the old name back into the schemas for the record.

mfhepp:
Ooops - I did not remember that we agreed to remove specialUsage. Can anybody point me to the discussion? For rental cars, it is a pretty important information (and mandatory in many countries). Didn't we probably agree to move it to Product?

sdml:
@mfhepp We changed the name of specialUsage to vehicleSpecialUsage. See #705 and #898.

danbri:
Yes, just a name change - sorry should have been clearer

danbri:
http://webschemas.org/docs/releases.html#g922

Issue #921, opened 2015-12-07
Members of DayOfWeek should belong to the schema.org ontology


betehess:
For some reason, it's still relying on http://purl.org/goodrelations/v1#.

Would the following snippet something we want to support?

```json
"openingHoursSpecification": [
  {
    "@type":"OpeningHoursSpecification",
    "dayOfWeek": [
      "Monday",
      "Tuesday",
      "Wednesday",
      "Thursday",
      "Friday",
      "Saturday"
    ],
    "opens":"09:00",
    "closes":"21:00"
  }
]
```

mfhepp:
The pattern is supported.
When we integrated GoodRelations into schema.org, we decided to leave the enumerated values in their original namespace. IMO, they work fine as they are and are properly listed in http://schema.org/DayOfWeek.

I would rather not include all enumerations from GR into schema.org; it just expands the size of the vocabulary.

betehess:
> The pattern is supported.

What do you mean? We can use Text instead of GoogleRelations' URIs?
It's really not clear to me which values are actually accepted/acceptable when I look at the documentation.

mfhepp:
No, you should use GoodRelations URIs.
But even if we added the enumeration to schema.org, you would have to use the full URLs, afaik.

Aaranged:
FWIW Google's initially-released specifications for use of this property on their [Local Business overview page](https://developers.google.com/structured-data/local-businesses/) specified the expected value as text, and provided an example which reflected this.

I say "intially-released" as the page, which initially returned a 404 from the link that appeared in the [Google Developers Structured Data site](https://developers.google.com/structured-data/), is now again returning a 404, so the screenshot displayed below is derived from the [Internet Archive copy](http://wayback.archive.org/web/20151203223607/https://developers.google.com/structured-data/local-businesses/) of Dec. 3, 2015.  It, of course, remains to be seen whether they'll used the plain text enumerations as pictured when they reactivate the pages or modify this to use the GR URIs.

![standard-hours](https://cloud.githubusercontent.com/assets/1033730/11644556/e38867ba-9d01-11e5-9419-3e87ad06b572.png)

betehess:
@Aaranged Google's Local Business is actually what triggered this issue :-) (and that's where the snippet was coming from)

[...]

Issue #245, opened 2015-01-21
schema.org has two ways to represent openingHours - need unified documentation and cornercase tweak


danbri:
(e.g. how to say closed all day xmas)

We added a complex opening hours construct from Good Relations without documenting how it fits with the existing single property openingHours micro-notation. That was a mistake.

Migrating issue from https://www.w3.org/2011/webschema/track/issues/22

        http://schema.org/CivicStructure has openingHours, whose value is said to be a Duration,
        "The opening hours for a business. Opening hours can be specified as a weekly time range, starting with days, then times per day. Multiple days can be listed with commas ',' separating each day. Day or time ranges are specified using a hyphen '-'.
        - Days are specified using the following two-letter combinations: Mo, Tu, We, Th, Fr, Sa, Su.
        - Times are specified using 24:00 time. For example, 3pm is specified as 15:00. 
        - Here is an example: <time itemprop="openingHours" datetime="Tu,Th 16:00-20:00">Tuesdays and Thursdays 4-8pm</time>. 
        - If a business is open 7 days a week, then it can be specified as <time itemprop="openingHours" datetime="Mo-Su">Monday through Sunday, all day</time>."

        As part of the GoodRelations addition, there is also openingHoursSpecification which takes http://schema.org/OpeningHoursSpecification structures as a value.

        The differences should be explained, and perhaps converged.

danbri:
There are subtleties, e.g. how multiple entries interact, esp. when combining the two different approaches.

danbri:
See also https://github.com/schemaorg/schemaorg/issues/372

unor:
See also https://github.com/schemaorg/schemaorg/issues/534 for past midnight opening hours

betehess:
Is there any plan to revisit this issue?

`OpeningHoursSpecification`'s intent seems to be exclusively capturing the [intended opening hours](https://github.com/schemaorg/schemaorg/issues/372#issuecomment-77530832):

> the reasoning about when a shop will be open will become extremely complex. When I designed the OpeningHoursSpecification in GoodRelations, my assumption was that points of sale typically promise when they open, but not, when they close.

From what I've seen, this is **not** how a business actually advertises its hours of operations. It would rather follow that pattern:

1) we are **generally open** on those days of the week
2) but we'll have **special opening hours** on those days
3) and we'll be **exceptionally closed** on those days

One the one hand, `openingHours` already captures 1 pretty well but doesn't help with 2 and 3.

[...]

Issue #534, opened 2015-05-23
OpeningHoursSpecification for a place that is open past midnight


unor:
On Stack Overflow, [devOp asked](http://stackoverflow.com/q/30400998/1591669) how to specify opening hours for something that is open past midnight, e.g., a place that is open from 22:00 to 03:00.

[I assumed](http://stackoverflow.com/a/30401896/1591669) that using something like the following snippet should be fine, as the definitions for [`closes`](http://schema.org/closes) says it’s the "closing hour of the place or service on the given day(s) of the week" (i.e., _yes_, it is the closing hour on that day, but it closes what was opened on the previous day):

```
<div typeof="schema:OpeningHoursSpecification">
  <link property="schema:dayOfWeek" href="http://purl.org/goodrelations/v1#Monday">
  <meta property="schema:closes" content="03:00:00">
  <meta property="schema:opens" content="22:00:00">
</div>
```

But then I found a [mailing list post](http://ebusiness-unibw.org/pipermail/goodrelations/2012-September/000475.html) from Martin Hepp (@mfhepp), where he explains for `gr:OpeningHoursSpecification`:

> opening hours that cross midnight must be broken into two chunks, one opening hour specification for the first day, then closing at 23:59:59, and one for the next day of the week, opening ant 00:00:00
> 
> […]
> 
> A consuming client should assume that if the shop closes for only 1 second, it is actually open without interruption.

As `schema:OpeningHoursSpecification` is derived from this, **I assume this advice still holds?** Or may the above snippet be used, too?

Either way, I think we should explicitly mention in the description how to handle past midnight opening hours, and add an example for that, too.

chaals:
I note that in Japan at least, the standard approach is to write times like "22-27" for the case you mention above. Which takes a bit of getting used to but makes sense…

…although I imagine it would upset processors if we tried moving to it :(

danbri:
We're getting to the stage where we need a custom issue label here for date/time/event issues - there are a lot! Will do that now.

odraencoded:
Has there been any progress on this?

>opening hours that cross midnight must be broken into two chunks, one opening hour specification for the first day, then closing at 23:59:59, and one for the next day of the week, opening ant 00:00:00

[...]

Issue #372, opened 2015-03-04
+1 ClosingDaysSpecification Type


ronanguilloux:
`ClosingDaysSpecification` Type is an additional StructuredValue, counterpart of `OpeningHoursSpecification`.

It scope is to ease the definition of opening/closing periods, as in an annual point of view.

__Proposal / RFC__:

Property | Expected Type | Description | Example
----------- | -------------------- | --------------- | ----------- 
closes | Date | a `Date`or a (collection of) `Date` |  `["2017-05-01", "2017-12-25", "2018-01-01"]`
closeDatesType | Text | a (collection of) HolidayType | `["BankHoliday", "PostalHoliday", "LocalHoliday"]`

This proposal also proposes to create an _HolidayType_, as in http://en.wikipedia.org/wiki/Public_holiday:
- `LocalHoliday` means the legal holiday list that apply in a certain place, as determined by a national/local administration.
- Other holiday types may be defined, as the [Bank Holiday](http://en.wikipedia.org/wiki/Bank_holiday)

mfhepp:
I am not in favor of this approach, because the reasoning about when a shop will be open will become extremely complex. When I designed the OpeningHoursSpecification in GoodRelations, my assumption was that points of sale typically promise when they open, but not, when they close. Already now, the reasoning of whether a point of sale is open or closed can be fairly complex, because you can have seasonal opening hours (via vaidFrom and validThrough), can computing which one is given priority is non-trivial.

If you now add negation, which is pretty much absent from other parts of schema.org so far, the reasoning gets way more complicated than it already is.

ronanguilloux:
Thank you very much for your feedback!

> my assumption was that points of sale typically promise when they open, but not, when they close

Well, this pretty much depend from the country & culture you see it.

Points of sale are far from being the only places to have open/close hours. But, as points of sale typically promise when they open but not when they close, most public buildings do exactly the opposite.

Just think about public buildings (schools, etc.) or public places (parks), civic structures, business centers, residences, landmarks, etc. : In French & Italian contexts (the ones I know the best), most of this places display time panels that are not strict timetables or schedules, but are rather a simple phrase, as "open on working days, [closed on holidays](https://www.google.fr/webhp?q=%22closed+on+holidays%22)" 

Such "holidays" concept may vary from local (regional) school holidays (FR) to Bank Holidays (UK), to Post Office holidays (India), etc.

[...]

PR #923, opened 2015-12-08
Days of week and special/closing OpeningHoursSpecification

betehess:
This is an actual proposal to tackle #245 and #921 (and possibly other issues). Please keep in mind that this is my first PR against Schema.org so I have probably missed a few things.

This adds 7 instances for each day of the week. References to GoodRelations in the docs and the examples should have been updated.

I tried to keep the current semantics as intact as possible. The "closing day" and "late night" use-cases are addressed by refining the definition of `OpeningHoursSpecification`. Special opening hours overriding typical opening hours are possible through the new `specialOpeningHoursSpecification` property.

A few remarks:
* I didn't find a way to order Enumeration Members in the produced HTML, so DayOfWeek members don't follow the natural order
* I don't understand how the JSON-LD context is managed so one example has an explicit `"dayOfWeek": { "@type": "@id" }` in its context while that should be the default

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -2339,21 +2339,9 @@
 
     <div typeof="rdfs:Class" resource="http://schema.org/DayOfWeek">
       <span class="h" property="rdfs:label">DayOfWeek</span>
-      <span property="rdfs:comment">The day of the week, e.g. used to specify to which day the opening hours of an OpeningHoursSpecification refer.
-&lt;br /&gt;
-    Commonly used values:&lt;br /&gt;
-&lt;br /&gt;
-    http://purl.org/goodrelations/v1#Monday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#Tuesday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#Wednesday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#Thursday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#Friday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#Saturday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#Sunday &lt;br /&gt;
-    http://purl.org/goodrelations/v1#PublicHolidays &lt;br /&gt;
-        </span>
-       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Enumeration">Enumeration</a></span>
-       <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#source_GoodRelationsClass">GoodRelationsClass</a></span>
+      <span property="rdfs:comment">The day of the week, e.g. used to specify to which day the opening hours of an OpeningHoursSpecification refer.</span>
+      <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Enumeration">Enumeration</a></span>
+      <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#source_GoodRelationsClass">GoodRelationsClass</a></span>
     </div>
 
     <div typeof="rdfs:Class" resource="http://schema.org/PriceSpecification">
@@ -2405,9 +2393,14 @@
 
     <div typeof="rdfs:Class" resource="http://schema.org/OpeningHoursSpecification">
       <span class="h" property="rdfs:label">OpeningHoursSpecification</span>
-      <span property="rdfs:comment">A structured value providing information about the opening hours of a place or a certain service inside a place.</span>
-       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/StructuredValue">StructuredValue</a></span>
-       <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#source_GoodRelationsClass">GoodRelationsClass</a></span>
+      <span property="rdfs:comment">A structured value providing information about the opening hours of a place or a certain service inside a place.
+&lt;br /&gt;
+The place is &lt;b&gt;open&lt;/b&gt; if the &lt;a href=&quot;/opens&quot;&gt;opens&lt;/a&gt; property is specified, and &lt;b&gt;closed&lt;/b&gt; otherwise.
+&lt;br /&gt;
+If the value for the &lt;a href=&quot;/closes&quot;&gt;closes&lt;/a&gt; property is less than the value for the &lt;a href=&quot;/opens&quot;&gt;opens&lt;/a&gt; property then the hour range is assumed to span over the next day.
+      </span>
+      <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/StructuredValue">StructuredValue</a></span>
+      <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#source_GoodRelationsClass">GoodRelationsClass</a></span>
     </div>
 
     <div typeof="rdfs:Class" resource="http://schema.org/OwnershipInfo">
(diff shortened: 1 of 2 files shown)


betehess:
Another remark: I am not sure why but the range for `openingHours` is `LocalBusiness` while the range for `openingHoursSpecification` is `Place`. So `specialOpeningHoursSpecification` follows the same range as for `openingHoursSpecification`. If there is no objection, I could have all 3 properties having `Place` as their range.

mfhepp:
Thanks for the proposal! However, I think the "opening hours" branch of schema.org is a non-trivial challenge and I suggest that we try to first work on a comprehensive solution to the pending challenges instead of an incremental approach with quick fixes, because quick fixed will likely constrain our redesign.

betehess:
@mfhepp I understand the concern re: quick fixes.

I have tracked down the issues/use-cases around opening hours and found the following:
* late night hours #534
* closing days #819 #824 #372
* special opening hours #245
* syntax for `openingHours` wrt HTML5 #143 #245
* day of the week instances #921 #894
* `openingHours` vs `openingHoursSpecification` #245  #186
* [explicit opening day approach](https://github.com/schemaorg/schemaorg/issues/372#issuecomment-77530832) vs "[typical/general opening hours, to allow for exceptions](https://github.com/schemaorg/schemaorg/issues/245#issuecomment-159063399)"

The reported usage statistics on the website say
* "Between 100,000 and 250,000 domains" for `openingHours`
* "Between 10,000 and 50,000 domains" for `OpeningHoursSpecification`
* "Between 1000 and 10,000 domains" for `DayOfWeek`

So I guess the questions are:
* do we really need a "redesign" here?
* what are the guidelines for revising and deprecating existing parts of the ontology?
  * what can be deprecated in the context of "opening hours"? Anything?
* is this very proposal addressing the issues identified at this time?
* who is working on a proposal?

betehess:
Not sure what to do with that... Does anybody really care? I mean, there's been open issues around `openingHours` for quite some time now and all proposals have been shut down and abandoned. In the meantime, folks are [shipping an implementation](https://developers.google.com/structured-data/local-businesses/) that is not aligned with the current state of the ontology, without really discussing with the community (or I don't know where conversations are happening).

So we have several options from here:
* we let those guys [experiment on their own](https://web.archive.org/web/*/https://developers.google.com/structured-data/local-businesses/), we wait for them do be done, then people start deploying data, and we eventually fix the ontology to reflect what was done
* or we try to fix the issues now by working together, if possible by inviting the people actually shipping products to the party

In my opinion, Google's approach is simple, elegant, and answers most of the issues listed above. No surprise this PR mostly reflects that. The only difference is about `DayOfWeek`, which could be adapted to Google's approach by making it a `http://schema.org/DataType` and by updating the JSON-LD context so that `Text` will be correctly coerced to `DayOfWeek`. I am more than happy to continue working on the proposal but only if I have a sense that there is some interest. Otherwise, I'll just stick with Google's implementation and stop worrying about standardizing in Schema.org.

danbri:
Nudging this thread. See also #245 and #921 comments.

@mfhepp - w.r.t. comprehensive solution, which issues would you seek to address beyond those identified by @betehess ?

Opening Hours are worth a bit of polish without overloading that task with defining an entirely general purpose scheduling mechanism. There may be other related usecases we can address later more comprehensively for other kinds of repeating schedule...

danbri:
Reviewing this i.e. http://schema.org/Monday   rdfs:label: Monday, rdfs:comment: The first day of the week.

... can we come up with a better description of "Monday"? It is not a strong convention to count from Monday = 1. 

E.g. in http://www.timeanddate.com/calendar/days/  
    Monday or Sunday first day of the week?
        The first day of the week varies all over the world. In most cultures, Sunday is regarded 
        as the first day of the week although many observe Monday as the first day of the week. 
        According to the Bible, the Sabbath or Saturday is the last day of the week which marks 
        Sunday as the first day of the week for many Jewish and Christian faiths, while many 
        countries regard Monday as the first day of the week.

        According to the international standard ISO 8601, Monday is the first day of the week 
        ending with Sunday as the seventh day of the week. Although this is the international 
        standard, countries such as the United States still have their calendars refer to 
        Sunday as the start of the seven-day week.

I quite like the way Wikipedia ground https://en.wikipedia.org/wiki/Monday - i.e. as being between Sunday and Tuesday. I would also like (in the machine-readable representation) to associate Monday with https://www.wikidata.org/wiki/Q105 and so on, both to build up our links with Wikidata and to get multilingual labels for all these items.

betehess:
> I quite like the way Wikipedia ground https://en.wikipedia.org/wiki/Monday

Very good remark. I have updated the PR accordingly.

> I would also like (in the machine-readable representation) to associate Monday with https://www.wikidata.org/wiki/Q105 and so on

I have done something like that:

```html
<span>Same as: <a property="schema:sameAs" href="http://www.wikidata.org/entity/Q105">Monday</a></span>
```

Is that the right approach?

danbri:
re sameAs, we haven't made a mapping to wikidata URIs before, but that looks plausible. Perhaps better to use the https variant, though I'm not sure which they consider the most canonical URI. - i.e. https://www.wikidata.org/wiki/Q105

betehess:
> I'm not sure which they consider the most canonical URI.

That's the Concept URI (bottom left link of the Wikidata page), which gets redirected to the page URL when in a browser. Note the entity vs wiki in the path.

danbri:
I meant http versus https versions of the entity URI

betehess:
@danbri the entity URI really is the http version, even if the https one resolves.

The statement `<https://www.wikidata.org/entity/Q105> ?p ?o` won't match anything on https://query.wikidata.org.

betehess:
That being said, I am still not happy with my PR because of https://github.com/schemaorg/schemaorg/pull/923/files#diff-473245f5cdaf7ff4e5dd3462e1a62bbcR180 . I still don't understand how/where I can modify the JSON-LD context :-/

betehess:
@danbri @mfhepp I have rebased the PR, it's ready for another look.

Note that I still don't understand where the JSON-LD context for schema.org is managed so this is not part of the PR (yet?).

jaygray0919:
Have the same concern. For our applications, it's important to know version changes. We don't use @vocab, but instead use the precise term in our @context. Therefore, also would like to know how JSON-LD context for schema.org is managed.

danbri:
Re JSON-LD context, the trick we use currently is that properties that have a domainIncludes of URL (alongside potentially others) get treated as expecting "@id" from a JSON-LD perspective. That catches the common case of relative URI references, where you don't want a dangling "../news/more.html" text value to end up in the parsed output instead of a full URI. The context is entirely generated from the master schemas in data/ but the exact structure has been evolving as we figure things out. The most recent change (this week) #990 should be both harmless and useful: we just take care to associate all schema.org terms with the schema.org vocabulary explicitly, so that our context file can be mixed in alongside others without confusion. This came out of collaboration with GS1 towards http://gs1.org/voc/ and the issues were not apparent when the context file implementation was first put together.

danbri:
I have merged this (thanks @betehess !) and made some modest tweaks to the description. Also carried over the public holiday type from GR.

What happened to the idea of specialClosingDay - should that be spec'd and added too?

danbri:
Closing per http://webschemas.org/docs/releases.html#g923 (but ping re specialClosingDay)

betehess:
> Also carried over the public holiday type from GR.

@danbri I fail to understand how `PublicHolidays` actually works, and why it was even needed. I failed to find discussions asking for it to be added in here. It makes http://webschemas.org/DayOfWeek look really weird.

In practice, I am not sure a `OpeningHoursSpecification` will ever be able to be applied to all `PublicHolidays`, which are highly locale-dependent. A schema.org consumer would always lack that important information: which are the public holidays for the context in which `PublicHolidays` was used?

Maybe an example would help?

But until this is figured out, can we remove it from sdo-deimos? I don't think `PublicHolidays` was addressing any urgent use-case anyway.

betehess:
> What happened to the idea of specialClosingDay - should that be spec'd and added too?

This is already handled with the combination of `OpeningHoursSpecification` and `specialOpeningHoursSpecification`.

* http://webschemas.org/OpeningHoursSpecification says "The place is open if the opens property is specified, and closed otherwise."
* and http://webschemas.org/specialOpeningHoursSpecification lets you override typical hours

danbri:
Ok, great, just wanted to make sure. If you've time to look over the changes that have been merged in, that would be great. Extra sanity checks are always useful!

betehess:
@danbri I did (not just my contributions), and everything else looks good!

danbri:
Thanks for checking!

Re PublicHolidays, indeed it seemed under-discussed, but on the basis that we are replacing a structure that has been in place since 2012 (http://blog.schema.org/2012/11/good-relations-and-schemaorg.html). I  wanted to make the new schema.org URLs a drop-in replacement for the GoodRelations ones. If we want to deprecate that construction we can work out a design for doing so and make that clear in the /PublicHolidays documentation. But to just abruptly omit it without explanation would I think be even more confusing. I've filed an issue to track this at https://github.com/schemaorg/schemaorg/issues/1139

PR #924, opened 2015-12-09
RDFA & Examples for ExternalEnumeration Proposal

RichardWallis:
New definitions for EnumerationValue, EnumerationValueSet, partOfEnumerationValueSet, enumerationValueCode, hasEnumerationValue

Plus examples.

See (#894) for full proposal


stuartasutton:
What more do we need to move this ExternalEnumeration proposal along?

danbri:
I'll try to draw wider attention to it in the next release cycle. It's the kind of thing that needs serious and focussed attention as it has impact across the entire project.

PR #925, opened 2015-12-09
Remove @value clauses from Periodical JSON-LD examples

### LINKED ISSUE #904: Remove @value from datePublished in Periodical JSON-LD examples
As noted by @Aaranged and related to #903, the JSON-LD examples for Periodical / ScholarlyArticle examples currently contain unnecessary `@value` stanzas that complicate the otherwise straightforward datePublished attributes. These should be removed.

dbs:
The JSON-LD examples for ScholarlyArticle included unnecessary @value
clauses due to a naive conversion from the RDFa source. Remove them, so
instead of:

      "datePublished": {
        "@value": "2006-10",
        "@type": "xsd:gYearMonth"
      },

we'll have the much simpler, easier to publish and consume:

      "datePublished": "2006-10",

Fixes #904.

### DIFF
--- a/data/sdo-periodical-examples.txt
+++ b/data/sdo-periodical-examples.txt
@@ -88,10 +88,7 @@ JSON:
       {
         "@id": "issue9735", 
         "@type": "PublicationIssue", 
-        "datePublished": {
-          "@value": "2010-07-03", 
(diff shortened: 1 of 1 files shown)


danbri:
thanks! and for the git tips. I'd forgotten about cherry-pick...

Issue #879, opened 2015-11-06
Collect some useful/interesting visualizations of the schema.org vocabularies


danbri:

danbri:
See also ./data/2015-04-vocab_counts.txt in the repo for source of the per-term stats we mention in each page. These are in broad buckets (1-10, 10-100, etc etc sites).

fabiovalse:
I want to show you this http://wafi.iit.cnr.it/webvis/lab/public/90a352775be6496cbb27/. It's the equivalent visualization you have in your schemas html page (http://schema.org/docs/full.html) but a bit more graphical and interactive. It's not finished yet but if you want to use it I can work on it and improve it.

Actually, I think that it is not the best visualization for your hierarchy. The sunburst is more compact while the tree uses a lot of space. In general, if you want to build a good visualization for your hierarchy I think that the best thing to do is to understand which are the need of the users. Which are the information users are interested in? Some examples:
- understand how big is the hierarchy
- understand how deep is the hierarchy
- understand which is the father of a certain type
- understand how many children has a certain type
- consult the properties of a certain type
- "incoming / outgoing term-count per type"

Anyway, we could visualize the same data you are working on, using a set of different visualizations (tree, sunburst, ...) and than include them in your website.

I think that the evolution diagram would be very interesting. I updated the one you saw about DBpedia improving the transitions between the different versions so that we can use it for your data too. I included all the past versions of the DBpedia hierarchy and now more insights come up (http://wafi.iit.cnr.it/webvis/lab/preview.php?gist_id=0dfe7280086553c4a233).

rajido:
In [BioSchemas](https://github.com/schemaorg/schemaorg/issues/1028) we are interested in visualisation on top of schema.org properties, types and content. We are looking to create several widgets using [BioJS](http://biojs.net) that could be easily reused for different purposes. Happy to count with your feedback and share our ideas of what it could be useful.

Visualisation with BioSchemas, Schema.org and BioJS is one the [project ideas proposed in GSoC 2016](https://biojsnet.herokuapp.com/gsoc/). Though the focus is on the bio domain we would like to create widgets that could work not just for us but for any other domains and schema.org information in general. We hope to get good proposals and push forward some interesting ideas.

fabiovalse:
@rajido: I'm interested in seeing the round visualization you expose on your website, the one on which you superimpose "Explore the possibilites of biological data representation today". I didn't find it on the website, may I see it?

Issue #718, opened 2015-08-07
Schema.org site markup has accessibility issues that need fixing


danbri:
1) full.html lists

The giant hierarchy of nested lists in http://schema.org/docs/full.html (where list items contain sub-lists) is poorly presented in (some? all?) screen readers. Reported by @LjWatson:

> The trouble with the full schema hierarchy is that screen readers don’t have an efficient means of 
> navigating through complex nested lists. Specifically, they have no way of moving from one li in a 
>  parent list to the next, without traversing all the child/grand-child/great-grand-child lists (and all their li) > in between.
> 
> The full schema hierarchy has lists nested up to five deep. So for example, to get from Action to
> CreativeWork, means navigating down through multiple nested lists – which collectively contain over 
> 100 individual list items. That’s a lot of key presses!   It would be good if the hierarchy could be made 
> interactive perhaps – so the branches could be expanded/collapsed. An example of the kind of thing 
> I’m thinking of is: http://hanshillen.github.io/jqtest/#goto_tree

2) tables in per-term pages

"Unhappy that the "tab-panel" on http://Schema.org  isn't accessible :( https://schema.org/Product  because no hrefs on the anchors. sadpanda", reported in https://twitter.com/stommepoes/status/625928819619307520

3.) selection CSS issue - see https://github.com/schemaorg/schemaorg/issues/1381

/cc @RichardWallis

danbri:
Nearby: https://github.com/schemaorg/schemaorg/issues/177 (this will let us compose URLs that point to specific examples, and then per-format versions of examples).

LjWatson:
For the tabs, there is a whole lot that could be done to improve accessibility for keyboard users and screen reader users. The basic design pattern goes something like this:

```<ul role="tablist" class="selectors">
<li role="presentation">
<a href="#html-panel" id="html-tab" role="tab" aria-selected="true" aria-controls="html-panel">Without markup</a>
</li>

<li role="presentation">
<a href="#microdata-panel" id="microdata-tab" aria-selected="false" aria-controls="microdata-panel" tabindex="-1">Microdata</a>
</li>

...
</ul>

<div id="html-panel" role="tabpanel" aria-hidden="false" aria-labelledby="html-tab"><pre>...</pre></div>
<div id="microdata-panel" role="tabpanel" aria-hidden="true" aria-labelledby="microdata-tab"><pre>...</pre></div>```

The basic overview of what's going on is...

[...]

PR #968, opened 2016-01-20
D3.js visualization of schema.org hierarchy

fabiovalse:
This pull request consists of a D3.js sunburst visualization for substituting the HTML indented tree of the schema.org hierarchy displayed [here](http://schema.org/docs/full.html).


Dataliberate:
Do you have a test instance of this that we can view?

fabiovalse:
Unfortunately it was running on my laptop. If you can't run it, I can try to put it on a server in the next days.

Dataliberate:
Managed to get it going locally by cloning from your [fabiovalse] fork.

Looks nice, I could imagine it being added as an optional view, but I could not see it being a default view that would be used to navigate around the hierarchy of Types. 

Couple of implementation points:  
* The tool-tip texts do not include the Type name so are not very helpful. 
* The display does not reflect the view selection buttons - try it at http://bib.localhost:8080/docs/full.html and select "Extension 'bib'"
* The link, when clicking on a segment in the sunburst, is hard coded to go to http://schema.org...  the rest of the application uses relative URLs to enable testing, operation within extensions, etc.

~Richard.

fabiovalse:
Ok, so do you have some solution in mind for the optional view? Something you prefer? A tab menu where you can switch between the main and the optional view?

Dataliberate:
Good question that ties in with (#879). 
 
We have the need to enable browsing through the Type hierarchy of the core vocabulary and each and all extensions for several uses. To answer the _what are the subtypes of CreativeWork_ style questions; Identify what core terms are super-types for an extension; as well serendipitous discovery aided by a visual representation.

We also need to be aware of (#718) accessibility issues which still needs some efforts to resolve.

I have no firm preferences as to how that page is remodelled. My current thoughts are that you should be able to choose what to display - core, extension(s), data types, combinations thereof - and then variations  of display (two or three) defaulting to the simplest/most accessible first.

danbri:
I'm sorry, I'd missed this one.  Let's consider it as addressing https://github.com/schemaorg/schemaorg/issues/879

Just tried viewing it (http://localhost:8080/docs/full.html is entry point?) via "git clone https://github.com/fabiovalse/schemaorg" ...  however the page only displays up until "Select vocabulary view: Core vocabulary   Core plus all extensions" and there's an error in the console:

>    sunburst.js:58 Uncaught TypeError: Cannot read property 'size' of null(anonymous function) @     sunburst.js:58(anonymous function) @ sunburst.js:204
58sunburst.js:38 Uncaught TypeError: ".semantic_zoom".attr is not a function(anonymous function) @ sunburst.js:38t @ d3.v3.min.js:1(anonymous function) @ d3.v3.min.js:1c @ d3.v3.min.js:3p @ d3.v3.min.js:3(anonymous function) @ d3.v3.min.js:1

danbri:
Ping @fabiovalse

mik-laj:
What's up with this?

danbri:
@fabiovalse did you continue to explore this?

danbri:
closing for now...

PR #984, opened 2016-02-09
bug fixes for bib.schema.org/Chapter example

dauwhe:
The JSON example wouldn't parse due to a stray comma, and all examples used startPage/endPage instead of the correct pageStart/pageEnd.


unor:
+1

danbri:
Thanks! :)

PR #993, opened 2016-02-18
Compound price pull request updated

mfhepp:
This supersedes #830 (which was designed against sdo-phobos. It also includes examples.


mfhepp:
This was in error, ignore

PR #1014, opened 2016-03-03
Seems like a typo

jasonkhanlar:
https://schema.org/OpeningHoursSpecification shows wrong html attribute and property name for first Microdata example


unor:
+1

danbri:
Thanks! :)

Issue #358, opened 2015-02-24
InformAction from an Agent


zeroware:
I have an `Agent` (a machine) that will send mail notifications to some `Person` when an `Event` occurs.

I found that `InformAction` is appropriate but i'm not quite sure about the usage.

Here is what I have so far : 

``` json
{
"@type":"http://schema.org/InformAction",
"http://schema.org/recipient": {
  "@type": "http://schema.org/Person",
  "http://schema.org/email": "user@example.com"
},
"http://schema.org/event": {
  "@type": "http://schema.org/Event",
  "http://schema.org/startDate":"..."
},
"http://schema.org/object": {
  "@type":"http://schema.org/EmailMessage",
  "http://schema.org/headline": "email subject...",
  "http://schema.org/text": "email body..."
}
}
```

Is this a correct way to describe a mail notification for an event ? 
Should I use `object` or `instrument` to include the `EmailMessage` ?

vholland:
The "instrument" property would seem a better fit.

I'll update the example to illustrate this.

vholland:
See pull request #1018

vholland:
Included in schema.org 3.0.

PR #1018, opened 2016-03-07
Issue #358: Added instrument to InformAction example.

vholland:


danbri:
bad idea? https://github.com/schemaorg/schemaorg/issues/358 is still open...

Issue #1004, opened 2016-02-22
Add broadcastFrequency to BroadcastService to specify the over the air frequency


vholland:
Authors have the ability to specify the channel a radio/TV broadcast uses on a particular service, but there is no way to specify the over the air frequency (e.g. 107.5 MHz for a radio station or 66 MHz for TV).

BroadcastService should have a "broadcastFrequency" property allowing either TEXT or a QuantitativeValue.

vholland:
See pull request #1019

danbri:
Sounds plausible to me. All your examples are unitCode "MHZ", looking at https://github.com/schemaorg/schemaorg/pull/1019/files ... what other codes are potentially applicable? 

I was wondering - would the mixed-case "MHz" be more conventional? But according to /unitCode and the subset/excerpt at http://wiki.goodrelations-vocabulary.org/Documentation/UN/CEFACT_Common_Codes ... it seems 'MHZ' is correct. I'd suggest including "MHZ" (and any others) in the property definition, to avoid people chasing around downloading UN/CEFACT spreadsheets.

danbri:
nearby, but this stuff always confuses me, http://www.vlf.it/frequency/bands.html

danbri:
I pinged Yves Raimond @moustaki on Twitter https://twitter.com/moustaki/status/710151533279576064 who responds:

     ":-D That looks great! I wonder if we might need something extra for e.g. DVB-S - @njh do you know?"
 
... where @njh conveniently has the same nick on Github; hi :)

If this is to cover TV too, we ought to get another example added for that (later if not immediately)

njh:
Hi!

Yes, digital broadcast systems are a bit more complex than analogue - need various broadcast parameters, not just the frequency. For TV, the Local Channel Number / LCN could be useful but you also have to describe the service that it is being broadcast on.

Describing this kind of thing for Radio has been done in this ETSI specification:
http://www.etsi.org/deliver/etsi_ts/102800_102899/102818/03.01.01_60/ts_102818v030101p.pdf

Which the BBC publishes its FM and DAB services here:
http://radio-service-information.api.bbci.co.uk/radiodns/spi/3.1/SI.xml
(I would love to convert this to JSON-LD)

This is something that also isn't handled very well in the BBC's Programme Ontology but it is something that Tom Grahame is looking into. So not a simple answer but I would say that ideally there would be a lot more to describe this type of thing with schema.org.

[...]

PR #1019, opened 2016-03-07
Issue #1004: Added broadcastFrequency to BroadcastService.

vholland:


danbri:
Manually integrated into data/ext/pending -> http://pending.webschemas.org/broadcastFrequency

PR #1025, opened 2016-03-10
Align deimos with Master

RichardWallis:
Synchronise sd-demos with running state of Master - applying urgent fixes shipped post 2.2


danbri:
Thanks Richard, hopefully we're back in sane setup again now. All tests pass. I've pushed to appspot  for tire-kicking: http://sdo-deimos.appspot.com/docs/qa.html

PR #1037, opened 2016-03-18
PropertyValue example fixes (pre-markup ex. had Microdata; alt attributes were missing) [supersedes #836]

unor:
Same as https://github.com/schemaorg/schemaorg/pull/836, but now against the new base branch.


danbri:
Thanks @unor ! :)

PR #994, opened 2016-02-18
Composite pricing, updated (supersedes #830)

mfhepp:
This pull request updates #830. It adds examples for two domains and addresses the issue described in #829.

CompoundPriceSpecification is a needed enhancement from GoodRelations for cases in which a price consists of multiple components (e.g. 50 USD per day and 0.10 USD per mile for a rental car).

PR #965, opened 2016-01-19
Financial (FIBO) proposals for core vocabulary

Dataliberate:
This Pull Request constitutes a small proposal for enhancements to the core of Schema.org from the [Financial Industry Business Ontology Community Group](https://www.w3.org/community/fibo/).  

The proposal offered by the Group is based on the work of the [Financial Industry Business Ontology project](http://www.fibo.org/schema) to extend the capabilities of Schema.org aligned with the Financial Industry Business Ontology (FIBO).  This has been a team effort utilising the expert knowledge and know-how of members of the FIBO Content Team involved in creating the Financial Industry Business Ontology, working with others familiar with Schema.org including Mirek Sopek, Richard Wallis and Martin Hepp.

The proposal consists of the following elements in the schema.rdfa file and some examples in sdo-fibo-examples.txt.  
* __New Types:__ MonetaryAmount, FinancialProduct, BankAccount, DepositAccount, LoanOrCredit, InvestmentOrDeposit, CurrencyConversionService.
* __New Properties:__ annualPercentageRate, feesAndCommissionsSpecification, interestRate, leiCode, loanTerm, maxAmount, minAmount, requiredCollateral.
* __Updated Type:__ CreditCard
* __Updated Properties:__ amount, acceptedPaymentMethod, currency.

More detail and links to the definitions in a test version of schema.org (<http://sdo-fibo.appspot.com/>) are available in the [Group Wiki](https://www.w3.org/community/fibo/wiki/Main_Page).

The proposal contains Types and Properties that will generally enhance capability in describing things of a financial nature but without introducing too much financial industry specific detail.  We also took the opportunity to extend the domain of several properties such as logo, brand, and isRelatedTo to help the description of Services as well as Products.  The future intension is to explore more detailed proposals that would form a hosted extension that builds upon these generally useful ones.

We look forward to comments, suggestions and questions that will help these enhancements become part of the vocabulary.

Richard Wallis.

PR #1038, opened 2016-03-18
Composite pricing examples: various fixes

unor:
I fixed the points mentioned in https://github.com/schemaorg/schemaorg/pull/994 (`itemtyp` → `itemtype`; absolute URL value for `itemtype` attribute; missing `</span>`) and:

* the Microdata examples had an additional `</div>`, which lead to parsing errors
* the JSON-LD examples used a duplicated `priceComponent` property instead of using an array, which lead to parsing errors (please review if I used the array correctly, as I don’t know JSON-LD very well)

### DIFF
--- a/data/sdo-examples-goodrelations.txt
+++ b/data/sdo-examples-goodrelations.txt
@@ -8,23 +8,22 @@ MICRODATA:
 
 <div itemscope itemtype="http://schema.org/Offer">
   <h1 itemprop="description">Our Private Bank Account Package: Low monthly fees and fair per-use charges.</h1>
-    <b itemprop="name">Private Bank Account Package</b>
-    <span itemprop="itemOffered" itemscope itemtype="http://schema.org/BankAccount">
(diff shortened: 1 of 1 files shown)


unor:
By the way, the used type `http://schema.org/BankAccount` is not part of schema.rdfa. Will it be added for the next version, or should another type be used?

mfhepp:
This is part of the FIBO proposal https://github.com/schemaorg/schemaorg/pull/965. We could replace the type in the example in order to remove this dependency.

danbri:
Thanks @unor :) Merged, but let's change that example while the FIBO stuff awaits a bit more discussion - but what can we change it to?

unor:
`BankAccount` seems to have this type hierarchy ([demo](http://sdo-fibo.appspot.com/BankAccount)):

> `Thing` > `Intangible` > `Service` > `FinancialProduct` > `BankAccount`

As `FinancialProduct` is new, too, `Service` would be the closest parent type.

Issue #1030, opened 2016-03-17
schema:encodesCreativeWork should be declared inverse of schema:encoding


tfrancart:
schema:encodesCreativeWork : The CreativeWork encoded by this media object.
schema:encoding : A media object that encodes this CreativeWork.

I don't see why they are not inverse of each other.

unor:
In case this finds agreement: https://github.com/schemaorg/schemaorg/pull/1039

danbri:
You are right, sorry we missed this!

PR #1039, opened 2016-03-18
Make 'encoding' and 'encodesCreativeWork' inverse properties

unor:
https://github.com/schemaorg/schemaorg/issues/1030


danbri:
Thank you for this. It was against an old branch, but I've implemented it against master based on this, thanks!

Issue #1041, opened 2016-03-18
RDFa representation contains music-related TODOs


unor:
The file `schema.rdfa` (live: http://schema.org/docs/schema_org_rdfa.html) contains some TODOs under the _Music_ section:

> TODO: For discussion - If a MusicGroup can have a genre, why not any Organization?
> 
> Deprecating musicGroupMember in favour of member (already done, but site UI is not ideal; TODO: we should update any examples).
> 
> TODO: For discussion - do we need something like CreativeWorkContributionRole ?
> 
> TODO: broaden meaning or add 'album' to name? 
> 
> TODO: Add CreativeWorkReleaseRegion - or decide whether PublicationEvent does the job, and rename properties accordingly.
> 
> TODO: Person birthPlace, deathPlace. (check old historical-info proposal, make a converged design later).

I think, in case they are not obsolete, these should be tracked in this issue tracker here, instead of the RDFa file.

unor:
Related: https://github.com/schemaorg/schemaorg/issues/1044 (`hasReleaseRegion` in HTML comment)

danbri:
Thanks, you are entirely right, the schema configuration file is the wrong place.  So the candidate issues are (or we can just address them here for now...)

* [ ] If a MusicGroup can have a genre, why not any Organization?
* [ ] Check examples for musicGroupMember (should be member)
* [ ] do we need something like CreativeWorkContributionRole ? (for music use case)
* [ ] (regarding which term???) broaden meaning or add 'album' to name?
* [ ] Add CreativeWorkReleaseRegion - or decide whether PublicationEvent does the job, and rename properties accordingly.
* [ ] Use of Person birthPlace, deathPlace for music use cases, or need improved vocabulary (check old historical-info proposal)

PR #1042, opened 2016-03-18
Removed TODOs from RDFa

unor:
https://github.com/schemaorg/schemaorg/issues/1041


danbri:
Thanks! I redid this manually and stored the text from TODOs that may need attention in #1753

Issue #1048, opened 2016-03-21
Add new Document and DocumentPermission types


vholland:
Documents often get shared over email, including access rights (read, write, comment) for the document.

I propose adding a new **DigitalDocument** (originally **Document**) type with the following subtypes:
* **SpreadsheetDigitalDocument**
* **PresentationDigitalDocument**
* **TextDigitalDocument**
* **NoteDigitalDocument**

To describe the permissions, a new **DigitalDocumentPermission** type would have the following properties:
* **grantee**
* **permissionType**

See pull request #1049 for en example.

Note: this description updated by @danbri to reflect naming changes agreed below (SpreadsheetDigitalDocument was SpreadsheetDocument, etc.).

Dataliberate:
Document has much broader generally accepted meaning than a [file] document that this proposal seems to be focused upon.

Libraries, archives, museums, etc. have shelving full of physical Documents.

Potentially there are many more document types than Presentation, Text, Spreadsheet and Note. Would not a documentType property be more flexible than individual subtypes? Types such as Government Report, Pamphlet, Medical Document, Legal Document, etc. could then be accommodated in this basic structure.

As a minimum physical Documents should be recognised in an example.

danbri:
Is /Users/danbri/Desktop/my_spreadsheet1.xls a SpreadsheetDocument ? Could it have DocumentPermission properties?  Would those describe filesystem access to the bytes, file-format related access mechanisms e.g. who has the password (see http://askubuntu.com/questions/223153/decrypting-a-password-protected-libreoffice-calc-ods-file-forgotten-password etc., haven't found the relevant list of standards yet); or they might just describe who 'should' have access, or who has access via some Web service or Cloud API.

I share @RichardWallis's concern about the scope of "Document". Currently people come to schema.org looking for document, poke around then realise that the closest match is really CreativeWork. If we are to subset CreativeWork and use up such a major term ("Document") we will need to be pretty clear what is included and what is excluded. Especially if we are touching on access control issues.

Are the Document types proposed here essentially cloud-hosted online things? or is the intent to cover free-floating standalone office-style files too?

jasondouglas:
We can find better names, but I don't see what's controversial about this proposal. File may be better than Document because in this day and age I think it implies digital more than document does. DigitalFile seems silly and PresentationDigitalFile even sillier.

Dataliberate:
[...]

PR #1049, opened 2016-03-21
Issue #1048: Document and DocumentPermission types

vholland:
Issue #1048.


Dataliberate:
Document has much broader generally accepted meaning than a [file] document that this proposal seems to be focused upon.

Libraries, archives, museums, etc. have shelving full of physical Documents.   

Potentially there are many more document types than Presentation, Text, Spreadsheet and Note.  Would not a documentType property be more flexible than individual subtypes?  Types such as Government Report, Pamphlet, Medical Document, Legal Document, etc. could then be accommodated in this basic structure.

As a minimum physical Documents should be recognised in an example.

danbri:
Let's discuss in the issue (#1048).

danbri:
This is merged already, closing.

Issue #1055, opened 2016-03-24
www.* redirects seem broken


danbri:
Trying http://www.schema.org/Action

curl -v http://www.schema.org/Action

sends headers

    HTTP/1.1 301 Moved Permanently
    Location: http://None:None/Action

and HTML with similar bad link

reported by StatelessCat in #swig IRC, http://chatlogs.planetrdf.com/swig/2016-03-24.html#T12-47-50

StatelessCat:
If specifing the Host Header everything seems fine:
`curl -v http://www.schema.org/Action -H "Host: schema.org"`
```
< HTTP/1.1 200 OK
< Content-Type: text/html; charset=utf-8
< Cache-Control: no-cache
< Access-Control-Allow-Origin: *
< Date: Thu, 24 Mar 2016 15:52:04 GMT
< Server: Google Frontend
< Content-Length: 22260
< 
...
```

This issue occurs only when we ommit the Host Header.

RichardWallis:
Fixed in V3.0

PR #1069, opened 2016-03-31
Fix to 301 Redirect issue (#1055)

RichardWallis:
Fix for (#1055)


danbri:
Looks like this is in already. Not sure how the pull request is still open? I backported to master and published too, hence www.schema.org works fine. Closing...

Issue #1068, opened 2016-03-31
Wording for properties available from extensions is confusing.


RichardWallis:
For example Recipe states:

* From Recipe: totalTime
* From CreativeWork: claimReviewd, contentReferenceTime, publisherimprint, ....

Makes it appear that Recipe is an extension.

Reword and possibly identify which extensions

RichardWallis:
Fixed in V3.0

PR #1070, opened 2016-03-31
Extension specific listing of additional properties on term page

RichardWallis:
Fix for issue (#1068)


danbri:
Thanks! Looks a lot better :)

See for example http://webschemas.org/Recipe

Issue #181, opened 2014-11-20
Consider an rdfs:comment sub-notation such as _Foo_ that can be auto-hyperlinked


danbri:
Currently we have to entity-escape HTML anchor syntax to link from the term definition to another schema.org term. We should consider a notation that makes this easier, less error prone, and readable in the RDFS. Perhaps there are conventions that could be borrowed from Markdown? 

I don't see anything perfect in http://daringfireball.net/projects/markdown/syntax#link although _foo_ is supposed to mean same as an 'em' HTML tag. So we could say if something is _emphasised_ and it is the name of a schema.org term, it should be automatically hyperlinked in the site navigation.

(and yes, this post would have been more readable if github wasn't interpreting it as markdown)

danbri:
The markdown implementation fixes this. Thanks @RichardWallis

PR #1075, opened 2016-04-04
Added markdown format processing of comment fields 

RichardWallis:
See https://daringfireball.net/projects/markdown/syntax

For links within Schema.org site use [[target]] for example [[CreativeWork]] will result in
```<a href="/CreativeWork">Creative Work</a>```

Other options:
```_term_``` becomes ```<em>term</em>```
```__term__``` becomes ```<strong>term</strong>```
```List\n\n* Item\n``` becomes ```List<ul><li>Item</li></ul>```

See issue (#181)

Once applied an issue to address the changing of comments containing escaped html codes in *.rdfa files should be created.


mfhepp:
Strong +1!

danbri:
See #1106 for followup details.

Issue #1024, opened 2016-03-10
The website's server does not accept HEAD requests


raindev:
```
$ curl -i -XHEAD http://schema.org/breadcrumb
HTTP/1.1 405 Method Not Allowed
Allow: GET
Content-Type: text/html; charset=UTF-8
Content-Length: 188
Date: Thu, 10 Mar 2016 08:52:25 GMT
Server: Google Frontend
```

Accordingly to RFC 2616 section 9.4:

> The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request.

danbri:
Thanks for the nudge. We already track this as #256 so I will close this as a duplicate issue, but I want you to know we do feel guilty about this! The reason it is taking a while was the need for some code cleanup so that the Content-Length header could be accurately computed; details in the other bug. /cc @RichardWallis

Issue #256, opened 2015-01-22
HTTP server should support HEAD method against Python-generated content


danbri:
Commandline curl -I flag fails ("HTTP/1.1 405 Method Not Allowed") on all our python-generated URLs. 

http://curl.haxx.se/docs/manpage.html

-i, --include

(HTTP) Include the HTTP-header in the output. The HTTP-header includes things like server-name, date of the document, HTTP-version and more...

-I, --head

(HTTP/FTP/FILE) Fetch the HTTP-header only! HTTP-servers feature the command HEAD which this uses to get nothing but the header of a document. When used on an FTP or FILE file, curl displays the file size and last modification time only.

Appengine automatically responds to HTTP HEAD for static URLs e.g. most of docs/* but for python scripted content we need to implement a head function, see https://cloud.google.com/appengine/docs/python/tools/webapp/requesthandlerclass#RequestHandler_head for details.

See the corresponding 'get' function for details, https://github.com/schemaorg/schemaorg/blob/master/api.py#L865

It is not a huge task but needs some care to deal with the content negotiation (HTML vs JSON-LD) of the homepage.

danbri:
There are two main cases here. The homepage (HTML vs JSON-LD) and the per-term documentation. To be able to get the content-length info needed for HEAD would require some reorg / tidying so that content generation was more nicely separated from sending it via GET. Other cases like favicon, docs/full.html too.

danbri:
From @mfhepp  in a duplicate report:

Also note that the current implementation does not support HTTP HEAD requests, which it should, IMO.

Evidence:

$ curl -i http://www.schema.org/
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Cache-Control: no-cache
Access-Control-Allow-Origin: *
Date: Thu, 21 May 2015 06:18:00 GMT
Server: Google Frontend
Content-Length: 4630
Alternate-Protocol: 80:quic,p=0

$ curl -i http://www.schema.org/Thing
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Cache-Control: no-cache
Access-Control-Allow-Origin: *
Date: Thu, 21 May 2015 06:21:07 GMT
Server: Google Frontend
Content-Length: 20554
Alternate-Protocol: 80:quic,p=0

$ curl -I http://schema.org
HTTP/1.1 405 Method Not Allowed
Allow: GET
Content-Type: text/html; charset=UTF-8
Content-Length: 188
[...]

PR #1080, opened 2016-04-06
Introduced handling of HTTP HEAD Requests

RichardWallis:
Also impemented 304 return codes for ETAG & modified since requests.
Addresses issues (#256) and (#1024)


danbri:
This is merged and live on webschemas.org staging / testing server

Issue #447, opened 2015-04-23
Add EventSeries as subclass of Series


wikier:
This is issue is to discuss the PR #446 about how to add support to event series.

It is related with some other ongoing discussions, such as the repeating events modelling (issue #240) and the new event types (issue #406).

danbri:
Thanks! Let's discuss under #240 for now...

danbri:
Can folk here please take a look at the Courses discussion, which has run directly into this topic: https://www.w3.org/community/schema-course-extend/wiki/Main_Page https://lists.w3.org/Archives/Public/public-schema-course-extend/2016Mar/0021.html

danbri:
I've just added EventSeries into the 'pending.schema.org' extension I've created - intent is to avoid things getting stuck in Github limbo for so long. Right now this is just in the upstream/dev site webschemas.org but I hope we'll agree a workflow where proposals can go more quickly to pending.schema.org without waiting for large releases.

See http://pending.webschemas.org/EventSeries

(files are in data/ext/pending/issue-447* ... could do with an example - maybe someone could convert the Turtle in the PR into JSON-LD?)

wikier:
Good to see this moving forward, @danbri!

Since #446 was based on a pretty old branch, I've just added a new PR #1088 with the same proposal we had, but adding an example (both in Turtle and JSON-LD).

chaals:
Can we have a bit more information in the definition please?

(Admittedly I can't think of a clear explanation immediately either :( )

danbri:
How about: 

"An EventSeries is a Series of related events. The nature of the association between the events can vary, but typical examples could include a thematic event series (e.g. topical meetups or classes), or a series of regular events that share a location, attendee group and/or organizers."

I have mentioned 'classes' here which might merit some consideration w.r.t. the courses work. /cc @philbarker

chaals:
Counter offer that also tests whether I have understood what everyone else has:

An EventSeries is a collection of events that share some unifying characteristic. For example, "The Olympic Games" is a series, which is repeated regularly. The "2012 London Olympics" can be presented both as an Event in the series "Olympic Games", and as an EventSeries that included a number of sporting competitions as Events.

"The Ashes" (cricket contest between Australia and England) is repeated frequently, but not according to any long-term stable schedule. The course "Beowulf: the manuscript and the man" can be an EventSeries consisting of individual seminars with the same participants, and an EventSeries of the course being held repeatedly over a decade.

[...]

PR #1088, opened 2016-04-08
Add Event Series support

wikier:
(supersedes #446, provides an implementation for #447)

Currently `schema.org` misses support for event series. 

The following diagram depicts the proposal:

Event Series desing

The proposal tries to minimize the impact on the current vocabulary by reusing already available terms, for instance `hasPart`/`isPartOf`. The design is based on the assumption that is better to materialize descriptions of each event rather than base the model on using the [recurrence rule](http://www.kanzaki.com/docs/ical/rrule.html).

The [Salzburger Festspiele](http://www.salzburgerfestspiele.at/) could be the typical example where such construction would be necessary to model the different editions every summer.

This proposal has been developed by [Redlink GmbH](http://redlink.co) in the context of [TourPack](http://tourpack.sti2.at/), a research project partially funded by the Austrian Research Promotion Agency ([FFG](https://www.ffg.at/en)), under grant agreement no. 845600.


danbri:
Thanks for this. I think we'll have to tweak it somewhat though. Schema.org does not advocate for general use of Turtle currently (RDFa + Microdata + JSON-LD give plenty of options), so we won't take the Turtle example into the site. For JSON-LD it could probably do with hand editing for publisher simplicity - I'm afraid the current file looks very machine generated and could say the same thing in a much smaller file.

wikier:
Understood, @danbri. Take a look now, please.

danbri:
Sorry, reverting - this changed core not just pending. My bad.

wikier:
@danbri :-?

danbri:
To explain properly: I thought I'd merge this to pick up your example, but it seems 1.) I've already done that manually 2.) the PR also included changes against data/schema.rdfa to define the vocabulary.

For this release (v3.0 in next few days) we'll stick with EventSeries as a pending extension, and I think all the files are in place for that already.  Also within pending let's keep things together under issue-447-* as the main filename, to avoid losing track of which things are where. 

As a next step the role of "Series" deserves some more consideration. It was more or less obsoleted previously by CreativeWorkSeries, not clear if it serves us any value as a common supertype here.

danbri:
How does http://pending.webschemas.org/EventSeries look? Shall we take out the Series supertype for now?

danbri:
I've convinced myself. I'll remove Series as a supertype.

PR #1091, opened 2016-04-11
prefer http over https (see PR comment)

abhisharma2:
We noticed Bing does not properly markup schema when you refer to the `https` version of schema.org for healthcare related microdata. I created a test site where you can use the Bing Markup Validator to see what I'm talking about:

https://bing-schemaorg-test-vhtvoulwqx.now.sh/

In order to prevent this from happening to other folks, I suggest we change the language to prefer the `http` version until we get full `https` support. Thanks!

### DIFF
--- a/docs/faq.html
+++ b/docs/faq.html
@@ -228,7 +228,7 @@
     There is a general trend towards using <code>'https'</code> more widely, and you can already write <code>'https://schema.org'</code> in your structured data.
     Over time we will migrate the schema.org site itself towards using <code>https:</code> as the default version of the site and our preferred form in examples.
     However 'http://schema.org' -based URLs in structured data markup will remain widely understood for the forseeable future and there should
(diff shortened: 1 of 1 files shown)


danbri:
Thanks for the suggestion. Microsoft's representatives on the schema.org steering group were previously OK with this language and the decision behind it, so my suggestion is to report this as a problem over at Bing. If @tmarshbing and @shankarnat think the wording should be changed I'm happy to explore this again.

abhisharma2:
Cool, thanks for the input. I'll ping Bing. And thanks for all the work on this @danbri, much appreciated.

shankarnat:
Hi Folks,

I would be looking into it and would get back shortly on this. Thanks!

Thanks,
Shankar N

abhisharma2:
Hey @shankarnat thanks! let me know if you need any more info/help from me.

PR #1102, opened 2016-04-14
Example fixes: no 'content' attribute for 'span' in Microdata; wrong attribute in Microdata; following recommendation for 'price' content

unor:
This PR only changes examples:

* in Microdata, the `span` element can’t have a `content` attribute (#184): either omitted the element (in case where the value is the same) or used the `meta` element instead
* the currency symbol should not be part of the `price` value
* Microdata: replaced `property` attribute with `itemprop`, and used the same properties from the RDFa example (`author`, `translator`)
* RDFa: used `typeof` instead of `typeOf`


danbri:
@unor - thanks so much! :)

PR #1103, opened 2016-04-15
Issue #1048: Added DigitalDocument.

vholland:


danbri:
Queued for review, please take a look: http://webschemas.org/DigitalDocument

(note there is currently a general software bug currently causing 'supersededBy' to show up as an available extension property, in http://webschemas.org/DigitalDocumentPermissionType ... please ignore that)

PR #1105, opened 2016-04-17
Revert 890 patch 1

abol:


danbri:
Thanks. I have lost track of what exactly happened here, but one way or another I believe the sdo-deimos repo is now in the correct state:

    15:52 $ ./scripts/rdfa2nt ./data/schema.rdfa | grep inverseOf | grep contain
    <http://schema.org/containedInPlace> <http://schema.org/inverseOf> <http://schema.org/containsPlace> .
    <http://schema.org/containsPlace> <http://schema.org/inverseOf> <http://schema.org/containedInPlace> 

I think we can close this one now. Thanks again!

PR #1132, opened 2016-04-28
Restored some changes lost with previous commits 

RichardWallis:


danbri:
Merging (noting that @RichardWallis and I have discussed this at length). Essentially it restores us to where we were on monday then adds back in all non-problematic commits.

Issue #1157, opened 2016-05-11
500 error served to W3C validator


danbri:
Both webschemas.org and schema.org (sdo-deimos / master codebase) give an HTTP 500 error to W3C's validator:

http://validator.w3.org/check?uri=http%3A%2F%2Fwebschemas.org%2F&charset=%28detect+automatically%29&doctype=Inline&group=0

@RichardWallis can you have a look?

RichardWallis:
Fixed in V3.0

PR #1159, opened 2016-05-12
V3.0 fixes1

RichardWallis:
- Fix for (#1157) 500 Error served to W3C Validator
- Potential fix for old version being cached after deployment - set max-age to 10mins
- Added creation of jsonldcontext to warmup processing
- Removed Excessive debug output


mfhepp:
H Richard, 
could you also please repair the manufacturer property which was lost in 3.0?
Thanks!

Martin

danbri:
I've fixed 'manufacturer' in the same update.

danbri:
http://schema.org/manufacturer

PR #1168, opened 2016-05-19
Use sdoapp loaded graphs for tests

Dataliberate:
Load graphs from application into local graph for tests


danbri:
Thanks, this should  catch a few more oopsies

Issue #1171, opened 2016-05-20
Lumpy performance of Schema.org site


RichardWallis:
Issue seems to be around new instances spawned to handle load having to create their own in memory page cache.  If you are unfortunate enough to get a new instance to service your request you have to wait whist it builds and caches pages to answer you.

Suggested investigation of use of NDB Cloud Datastore to create a single page cache to be shared by all instances.

RichardWallis:
Fixed in 3.1

PR #1173, opened 2016-05-20
NDB Pagestore implementation

RichardWallis:
In response to issue (#1171) 
Overview:
* Previously created pages are held in a within instance memory cache that is shared between threads.  This means that each page is recreated for each application instance.  We previously noted that the live site spawned off dozens of instances to support the load,
* This version uses the Cloud Datastore to hold cached pages.
* The single NDB store is shared between all threads of all instances.
* ​_Should_​ result in a lot less page creations, and more importantly you having to wait for a new instance to create you a page.
*  The NDB store is persistent so will keep the page store even when you reload/deploy a new version of the code/data.
* Implemented code that detects this condition and deletes all pages in the store as it starts. 
* For early supporting there is a way to  manually call this caheFlush function.

If successful a similar approach should be considered for the underlying RDF data.

### DIFF
--- a/api.py
+++ b/api.py
@@ -189,7 +189,7 @@ class PageStoreTool():
         
     def setCurrent(self,current):
         self.tlocal.CurrentStoreSet = current
-        log.info("Setting CurrentStoreSet: %s",current)
+        log.info("PageStore setting CurrentStoreSet: %s",current)
         
     def put(self, key, val,cache=None):
         ca = self.getCurrent()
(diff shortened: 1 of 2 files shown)


RichardWallis:
Superseded by #1213

RichardWallis:
Reopening to enable applying to master

danbri:
Ok, 2nd time was successful. I had to hunt for the settings to enable datastore writes in schema.org version of the site. For reference that is found via https://cloud.google.com/appengine/docs/python/console/managing-datastore

PR #1194, opened 2016-06-02
Add schema:Service as range for schema:typeOfGood

wikier:
As discussed at https://lists.w3.org/Archives/Public/public-schemaorg/2016Jun/0003.html

Contribution developed by [Redlink GmbH](http://redlink.co) in the context of [TourPack](http://tourpack.sti2.at/), a research project partially funded by the Austrian Research Promotion Agency ([FFG](https://www.ffg.at/en)), under grant agreement no. 845600.


mfhepp:
+1, makes sense

mfhepp:
Just to avoid duplication: The same fix is included in https://github.com/schemaorg/schemaorg/pull/1224.

danbri:
Thanks for this suggestion, @wikier! It seems we have this now (via #1224 I believe) but forgot to close this pull request.

PR #1200, opened 2016-06-10
Updating description for deliveryLeadTime

gmackenz:
On site pickup is a delivery option. To make deliveryLeadTime clearer, modify the description to read:

"The typical delay between the receipt of the order and the goods either leaving the warehouse or being prepared for pickup, in case the delivery method is on site pickup."


danbri:
@gmackenz mentioned that this PR has gotten messed up. However the basic idea makes sense.

Here's the corresponding issue: https://github.com/schemaorg/schemaorg/issues/1201

This seems modest and useful and uncontroversial. I'd like to squeeze it into 3.1, rather than stand on ceremony. /cc https://github.com/schemaorg/schemaorg/issues/1212

PR #1202, opened 2016-06-11
Removing diff lines

lambda2:
Removing existing diff line on `WebSite` examples from commit 1102989


chuckreynolds:
nice - just saw those - came here to fix em. good job OP

danbri:
Thanks!

PR #1223, opened 2016-06-22
Fix minor typo in extension.html

joshmoore:
`release.html` was apparently renamed to `releases.html`.

I wasn't sure if use of "our" was appropriate, but that would make this read more nicely.


danbri:
Thanks!

Issue #1234, opened 2016-06-30
Missing reference to extension defined Types in 'may appear as values' section


RichardWallis:
If a Type that would appear in the "Instances of XXXX may appear as values for the following properties" listing is defined in an extension that is not being shown, that type is not displayed.

Example compare http://webschemas.org/ContactPoint with http://pending.webschemas.org/ContactPoint

In the latter HealthInsurancePlan is displayed as a type in the listing but not in the former.

danbri:
Published via http://schema.org/docs/releases.html#v3.1
http://blog.schema.org/2016/08/schemaorg-update-hotels-datasets-health.html

PR #1242, opened 2016-07-01
Fix (#1234) Show incoming properties for all layers

RichardWallis:


danbri:
Merged. Can we have an entry for docs/releases.html please? (just type it here if you like)

RichardWallis:
For docs/releases.html:
Enhance incoming properties display (_Instances of XXXX may appear as values for the following properties_) to include properties from extensions.

danbri:
I've updated releases.html - https://github.com/schemaorg/schemaorg/commit/5a7cc17e64a0b68115df8b4c9c0a5f523367ab1b

pmika:
This is an automatically generated message.

pmika@yahoo-inc.com is no longer with Yahoo! Inc.

Your message will not be forwarded.

If you have a sales inquiry, please email yahoosales@yahoo-inc.com and someone will follow up with you shortly.

If you require assistance with a legal matter, please send a message to legal-notices@yahoo-inc.com

Thank you!

Issue #1083, opened 2016-04-06
Improving Dataset descriptions


danbri:
Talking with [Natasha Noy](http://web.stanford.edu/~natalya/) about possible improvements around dataset description. Some things to look into:

* coverageStart and coverageEnd (currently, the datasetTimeInterval has DateTime, not interval, as its expected time, which I think is not correct, or at least doesn't allow us to specify the coverage interval)
timestep (dct:accrualPeriodicity)
* bibliographic reference: many of the dataset refer to the paper that describes it
* Main variables measured -- without necessarily knowing the distinction of which ones are dimensions and which ones are measures qb:dimensionProperty and qb:MeasureProperty)

# Related work

This all starts to get into the business of looking inside the dataset, which was discussed at schema.org previously - e.g. see [Looking inside tables](https://lists.w3.org/Archives/Public/public-vocabs/2013Aug/0033.html) thread from Omar. Subsequently in [W3C CSVW](https://www.w3.org/2013/csvw/wiki/Main_Page) some of these ideas went standards track, in particular a templating mechanism to map tabular data into RDF.

* See also the (SDMX-oriented) [W3C Data cube specification](https://www.w3.org/TR/vocab-data-cube/).
* [W3C DCAT final specification](https://www.w3.org/TR/vocab-dcat/) (earlier drafts inspired the schema.org Dataset design) 
* More recently: [DCAT Application Profile for data portals in Europe (DCAT-AP)](https://joinup.ec.europa.eu/asset/dcat_application_profile/description) - @danbri discussed briefly w/ DCAT-AP team the possibility of using a JSON-LD-based @context for DCAT-AP as an 'external extension'.
* [CSVW RDF examples using data cube](https://github.com/w3c/csvw/blob/gh-pages/examples/simple-weather-observation.md) vocabulary from @6a6d74

darobin:
This is also related to https://github.com/schemaorg/schemaorg/issues/975 for versioning dependencies, particularly on datasets (discussed in more detail in https://research.science.ai/article/web-first-data-citations).

danbri:
See also #1066 for a quick bugfix (spotted by Natasha too)

natashafn:
A couple of follow up comments:
- there is a [citation](https://schema.org/citation) property on CreativeWork that is probably the right property to use for bibliographic reference
[...]

PR #1247, opened 2016-07-15
Sdo datasets2

danbri:
Changes to support datasets (but also touches CreativeWork). Main issue is #1083, releases.html entry covers sub-issues.

Additional notes copied here for context:

CHANGES

1.) for temporal and spatial coverage.

As of v3.0 we have:

Relating to Dataset specifically,

 http://schema.org/spatial (Dataset -> Place),
   "The range of spatial applicability of a dataset, e.g. for a dataset of New York weather, the state of New York."

 http://schema.org/temporal (Dataset -> Datetime),
 "The range of temporal applicability of a dataset, e.g. for a 2011 census dataset, the year 2011 (in ISO 8601 time interval format)."

The temporal property superseded by the awkwardly named http://schema.org/datasetTimeInterval -

 http://schema.org/datasetTimeInterval (Dataset -> Datetime),
 "The range of temporal applicability of a dataset, e.g. for a 2011 census dataset, the year 2011 (in ISO 8601 time interval format)."

Relating to CreativeWork,

http://schema.org/contentLocation (CreativeWork -> Place),
"The location depicted or described in the content. For example, the location in a photograph or painting"

http://schema.org/locationCreated (CreativeWork -> Place),
"The location where the CreativeWork was created, which may not be the same as the location depicted in the CreativeWork."

Note also http://schema.org/releasedEvent which structures things a little differently, grouping place/time within an Event.

PROPOSAL:

1a. a minor detail re releasedEvent, but documenting here:
For works (most typically media broadcasts but potentially e.g. datasets too) whose publication is structured in terms of documented releases, it is reasonable to expect the release information in a http://schema.org/PublicationEvent to match direct contentLocation or spatial[Coverage] properties if the latter are present. A startDate property of the event would match http://schema.org/dateCreated of the published item.

1b. 
Create spatialCoverage and temporalCoverage properties as successors to the (vaguely and/or awkwardly named) datasetTimeInterval, spatial and temporal properties.

1c. 
Broaden spatialCoverage and temporalCoverage so that they apply to CreativeWork rather than just Dataset.

1d. 
Update their textual definitions to accommodate their broader scope, and to address any confusion about related properties.
Proposed text:

spatialCoverage: "The spatialCoverage of a CreativeWork indicates the place(s) which are the focus of some work. It is a subproperty of
contentLocation intended for more technical and specific materials. For example with a Dataset, it indicates 
areas that the dataset describes: a dataset of New York weather would have spatialCoverage which was the place: the state of New York."

temporalCoverage: "The temporalCoverage of a CreativeWork indicates the period that the content applies to, i.e. that it describes. In
the case of a Dataset it will typically indicate the relevant time period in a precise notation (e.g. for a 2011 census dataset, the year 2011 (in ISO 8601 time interval format). Other forms of content e.g. ScholarlyArticle, Book, TVSeries or TVEpisode may indicate their temporalCoverage in broader terms - textually or via well-known URL."

1e. 
Update RDFS assertions.

* spatialCoverage subPropertyOf contentLocation.
* temporal supersededBy temporalCoverage. (rather than by datasetTimeInterval as now)
* datasetTimeInterval supersededBy temporalCoverage.

Add mappings,

* temporalCoverage equivalentProperty http://purl.org/dc/terms/temporal
* spatialCoverage equivalentProperty http://purl.org/dc/terms/spatial

### DIFF
--- a/data/ext/pending/issue-1083.rdfa
+++ b/data/ext/pending/issue-1083.rdfa
@@ -1,0 +1,10 @@
+<div typeof="rdf:Property" resource="http://schema.org/variablesMeasured">
+  <span>Category: <span property="schema:category">issue-1083</span></span>
+  <span class="h" property="rdfs:label">variablesMeasured</span>
+  <span property="rdfs:comment">The variables that are measured in some dataset, either described as text or as pairs of identifier and description using PropertyValue.</span>
+  <span>domainIncludes: <a property="http://schema.org/domainIncludes" href="http://schema.org/Dataset">Dataset</a></span>
+  <span>rangeIncludes: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
+  <span>rangeIncludes: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
+  <link property="http://schema.org/isPartOf" href="http://pending.schema.org" />
+  <span>Source:  <a property="dc:source" href="https://github.com/schemaorg/schemaorg/issues/1083">#1083</a></span>
+</div>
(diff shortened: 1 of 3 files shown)


danbri:
Draft queued at http://webschemas.org/docs/releases.html#g1083

PR #1273, opened 2016-08-01
Update star rating example

gmackenz:
Update the existing example to use 'author' property (instead of 'name') to represent the rating awarding organization.


danbri:
Thanks! I was planning to fold this in, except it needs rejigging slightly to interpose an Organization entity since we don't actively encourage author to take strings. Care to refine the PR?

gmackenz:
Updated the pull request to interpose Organization entity on 'DEHOGA'.

PR #1281, opened 2016-08-06
Versions fix

RichardWallis:
* Introduced /versions/latest mapping to current versions of download files
 * Fixed 500 error on /version/*/ pages
 * Updated method of building source page for ‘schema-all.html’:
     http://localhost:8080/version/build-latest
 * Updated text of fullReleasePage.tpl to link to definition file downloads for latest version.
 * Updated RELEASING.TXT to reflect change to curl call for ‘schema-all.html’

### DIFF
--- a/RELEASING.TXT
+++ b/RELEASING.TXT
@@ -50,8 +50,8 @@ Of these,
 
  * schema.rdfa is a frozen one-time copy of data/schema.rdfa
  * schema-all.html is generated from this via running the Web app and
-   capturing via curl of /version/ (no numbers),
-   e.g. http://localhost:8080/version/
+   capturing via curl of /version/build-latest,
+   e.g. http://localhost:8080/version/build-latest
    when served locally via "dev_appserver ."
    For now, you need to edit this html to remove webschemas.org from heading section.
 
--- a/app.yaml
+++ b/app.yaml
@@ -61,8 +61,21 @@ handlers:
 - url: /search_files
   static_dir: static/search_files
 
-- url: /releases/*/
+- url: /version/latest/.*
+  script: sdoapp.app
+
+- url: /(version/[^/]*/)$
+  script: sdoapp.app
+
+- url: /(version/([^/]*))$
+  script: sdoapp.app
+
+- url: /version/
+  script: sdoapp.app
+
+- url: /version/*/*
   static_dir: data/releases/
+  application_readable: True
 
 - url: /.*
   script: sdoapp.app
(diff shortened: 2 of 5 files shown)


danbri:
this should now be showing up on webschemas.org for testing

AymenLoukil:
+1

danbri:
Yup http://webschemas.org/version/3.0/ works again. Not sure about /latest/ since it redirects to something we didn't set up yet for 3.1...

AymenLoukil:
@danbri  shouldn't we block indexation for http://webschemas.org because it's a duplicate of schema.org ..

RichardWallis:
To see what latest would look like, check out http://webschemas.org/version/build-latest

danbri:
@AymenLoukil I think it's ok that people find it now, since we added the banner saying "Note: you are viewing the webschemas.org development version of schema.org. See How we work for more details.", and explained http://webschemas.org/docs/howwework.html

AymenLoukil:
@danbri Yeah i know but i'm talking from an SEO point of view. Duplicate content is not good.

We can add nonindex or an external conical tag pointing to the original website.

danbri:
I don't think we're very worried about SEO.

Tiggerito:
What may happen is search engines will pick the wrong version, and start sending people to the development pages instead of the official ones. And that's a user/usability issue.

SEO wise, probably using canonical tags to the official pages is the neatest solution. A clear signal on which version is the one you want people to go to. It may even make it possible to still Google search the dev site using the site: operator. 

noindex or even using robots.txt are other methods. noindex does not have the benefit of ranking signals being merged and ensures no way to Google search the dev site. robots.txt is not a recommended method, but it completely stops crawling so helps if bot activity is eating up resources.

AymenLoukil:
+1 @Tiggerito  :) #up_SEO_Voice

danbri:
Thank you for your SEO concern. The real issue w.r.t. "sending to the wrong version" situation is in fact that at least Google has started routing people to https://schema.org/ before we as a project agreed to make that the primary URL (for which we'll want to update all examples etc too). So our use of "canonical" is waiting on that. It is perfectly fine for webschemas.org to be discoverable as it is the latest "bleeding edge" version of the schema.org effort.

Tiggerito:
I was wondering about the switch to https. So people should be using that from now on? 

Google tends to show the https version if both are available, so you will see the https pages are now showing up in the search results in place of the http pages. 301 redirects can force that. It depends if you want to force users to the secure pages now.

Being https will also help keep webschemas out of the general results. So you could leave as is and just keep an eye out to make sure webschemas does not start stealing results.

Issue #1262, opened 2016-07-26
Rename feature and AccommodationFeature to preserve namespace and extend domain


vholland:
http://webschemas.org/feature was introduced as part of the hotels proposal in pull request #1224. Given that schema.org exists in a single namespace and "feature" could be applied to many things, we should consider renaming the property.

This got me to thinking that the style of feature listed for hotels may apply to other types of businesses. For example, a coffee shop may want to specify that it has WiFi.

I propose:
* Renaming "feature" to "locationFeature".
* Renaming "AccommodationFeature" to LocationFeatureSpecification.

mfhepp:
Vicki:
I would have no objections if you want to rename the property to locationFeature or placeFeature. Note that we did not name it accommodationFeature in order to avoid having the same name for property and type.

Same for the type, but I think that PlaceFeature is better.

Also note that you can always use the plain additionalProperty property; the main reason for feature / AccommodationFeature is that the features can have opening / availability times, which is not a generic property of a PropertyValue.

If you think that these rather late name changes quite some time after our conference call are mission-critical, I kindly ask you to implement them by means of a pull request, and to make sure they are also reflected in all parts of the documentation (namely examples and hotels.html). Thank you!

danbri:
@mfhepp - with respect these "late name changes" suggestions are not so late; @vholland raised the matter back in [December last year](https://github.com/schemaorg/schemaorg/issues/915#issuecomment-162013384). Re-reading the dialogue from #915 to figure out where we are and how we got here...

* @mfhepp responded "will reply in detail asap" ([Dec 14 2015](https://github.com/schemaorg/schemaorg/issues/915#issuecomment-164365650))
* @vholland reiterated that proposal (amongst others), and implemented it, [on February 29th 2016](https://github.com/schemaorg/schemaorg/issues/915#issuecomment-190321822).
* @mfhepp detailed response of [March 16th 2016](https://github.com/schemaorg/schemaorg/issues/915#issuecomment-197444349): "I have reviewed @vholland 's proposal and suggest we should stick with our approach and address the individual concerns / aspects individually. ".
 * rejected proposal of AccommodationFeature -> schema.org/AccommodationFeatureSpecification
 * counter-suggestion of *amenity for schema.org/feature -> schema.org/accommodationFeature
  * @mfhepp: "We could use schema:amenity, schema:includedAmenity, schema:optionalAmenity.  Action: Change names and examples and hotels.html"

[...]

PR #1297, opened 2016-08-11
Hotels doc: 'amenityFeature' instead of 'feature'

unor:
I think it was intended to reference the [`amenityFeature` property](http://schema.org/amenityFeature) here, instead of the `feature` property (which doesn’t exist).

(Related issue about the renaming: #1262)


danbri:
Thanks, I missed this edit. I think it's worth fast-tracking onto the site asap...

danbri:
I've quietly made the same edit to master branch and republished. Thanks @unor!

PR #1163, opened 2016-05-16
Fibo.schema.org v1.0

RichardWallis:
This Pull Request constitutes a proposal for a FIBO extension to Schema.org from the [Financial Industry Business Ontology Community Group](https://www.w3.org/community/fibo/).

This proposal builds upon the the new types and properties for the support of financial products and services introduced into Schema.org V3.0.

It introduces a first raft of new and modified Types and Properties to support in more detail the requirements for sharing financial products, services and organisations using Schema.org on the Web. This initial (v1.0) proposal has a focus on the area of loans, plus the addition of some more generally useful terms.

The proposal offered by the Group is based on the work of the [Financial Industry Business Ontology project](http://www.fibo.org/schema) to extend the capabilities of Schema.org aligned with the Financial Industry Business Ontology (FIBO). This has been a team effort utilising the expert knowledge and know-how of members of the FIBO Content Team involved in creating the Financial Industry Business Ontology, working with others familiar with Schema.org including Mirek Sopek, and Richard Wallis.

More detail and links to the definitions in a test version of schema.org (http://fibo.sdo-fibo.appspot.com/) are available in the [Group Wiki](https://www.w3.org/community/fibo/wiki/Fibo.schema.org_v1.0_Proposal).

PR #1300, opened 2016-08-12
Move Fibo.schema.org proposal in to pending

RichardWallis:
Replaces PR (#1254)  (merge into sdo-makemake)


RichardWallis:
See conversations on PR (#1254) & Issue (#1253)
This is an update to the original proposal PR (#1163) - A Pull Request constituting a proposal for a FIBO extension to Schema.org from the [Financial Industry Business Ontology Community Group](https://www.w3.org/community/fibo/).

Changes from original:
- Proposing a prerelease move into pending.schema.org
- Moving ServiceChannel specific proposals into a separate proposal - PR(#1298)
- Update property names to be more focused on the area they are used.

rvguha:
The schema changes look good. 

Examples: Lets use fictitious entities in our examples.

RichardWallis:
No problem on the fictional examples.

Will update these in next couple of days and also resolve the merge conflict.

danbri:
My review notes:

* PR includes a needless (and conflicting) sdoapp.py edit - can you remove that please?
* let's put this into pending.schema.org first, with a view to moving it along towards a hosted extension subdomain a release or two after that.
* I would prefer its eventual home to be under something like "finance.schema.org". We have merged in contributions from BBC, EBU, MusicBrainz, GoodRelations, GS1 and many others in previous releases without creating a project-acronym-based subdomain. I appreciate the desire for fibo.schema.org but I suggest it is better to keep collaboration and derivation acknowledgments elsewhere (in human-facing text and in machine definitions, but not in subdomains).

sopekmir:
Hi Dan,
Sorry for late reply. We will remove the edit and we are also making our examples fictional (after Guha's request).
As for placing it into pending - that's great and this is exactly what we expected to happen. Thanks !

When it comes to the change to the name of the extension, I hope it will not be a problem for the other members of the community which brought the concept of the extension to life. I will start my consultations asap.

sopekmir:
BTW, I just got info from Richard that the requested changes has been made! Thanks Richard !

Dataliberate:
Now issues raised by Guha have been addressed, can we merge this now.

trypuz:
@danbri, @RichardWallis, what is the next step? Is there anything FIBO community can do to move the FIBO extension from "pending" towards a hosted extension?

sopekmir:
I'm aware about reservations to move the extension to the one named by "fibo". Indeed, living between the two worlds, I must admit we had long lasting difficulties with the FIBO team to give us proper, canonical FIBO URIs, so that we could amend the extension documentation and have a "hard link" between the extension and the ontology. However, as it is more problematic for teh extension to stay in "pending" for so long, than having a permanent name, maybe we would indeed go for finance.schema.org for the upcoming edition? Is it possible?

PR #1314, opened 2016-08-18
Revert "adding maximumCapacity and remainingCapacity Closes #1209"

### LINKED ISSUE #1209: Add information about places available at event
There should be some way to identify the number of places that are available for an event. Both in total, as sometimes it matters whether an event in a small gathering of a few people or a coming together of thousands, and also the number of places that remain available, as sometimes it is useful to know that an event is about to sell-out. 

This came up in the context of educational courses as events (see [email discussion](https://lists.w3.org/Archives/Public/public-schema-course-extend/2016Apr/0000.html) for the schema-course-extend group, also mentioned in issue #1092); but seems broader than educational courses.

I suggest two properties:
eventCapacity: integer; the total number of individuals that may attend the event.
placesRemaining: integer; the number of places for the event that remain unallocated.

(Maybe the inventoryLevel property of Offer can be used for places remaining, but it would require a change to the definition)

danbri:
Reverts schemaorg/schemaorg#1279

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -5672,13 +5672,7 @@ The place is __open__ if the [[opens]] property is specified, and __closed__ oth
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Place">Place</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
     </div>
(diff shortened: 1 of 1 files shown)


danbri:
Oops, I did not check which branch this merged into. I will add it to sdo-callisto (likely schema.org v3.2) manually, see #1209 for discussion (I'll change the names too).

Issue #195, opened 2014-12-11
Course, CourseInstance, EducationEvent [Schema.org Course Extension]


westurner:
- [x] Thing > CreativeWork > Course https://schema.org/Course
- [x] Thing > Event > CourseInstance https://schema.org/CourseInstance
- [x] Thing > Event > EducationEvent https://schema.org/EducationEvent

.

- | Project: https://www.w3.org/community/schema-course-extend/
- | MailingList: https://lists.w3.org/Archives/Public/public-schema-course-extend/
- | Wiki: https://www.w3.org/community/schema-course-extend/wiki/
- | Issue: https://github.com/schemaorg/schemaorg/issues/195
- | PR: https://github.com/schemaorg/schemaorg/pull/972
- | RDFa schema & examples: https://github.com/schemaorg/schemaorg/blob/sdo-deimos/data/ext/pending/issue-195.rdfa
- | Web: http://course.schema-course-extend.appspot.com/
- | Web: http://pending.webschemas.org/Course
- | Web: http://pending.webschemas.org/CourseInstance
- | Web: https://schema.org/Course
- | Web: https://schema.org/CourseInstance

Goals:
- schema.org/Course , schema.org/CourseSection , schema.org/EducationEvent 

> SCHEMA COURSE EXTENSION COMMUNITY GROUP
> 
> This mission of this group, initiated by LRMI, is to develop an extension for schema.org concerning the discovery of any type of educational course (online/offline, long/short, scheduled/on-demand). Educational course is defined as "some sequence of events and/or creative works which aims to build the knowledge, competence or ability of learners". (Out of scope: information about students and their progression etc; information needed internally for course management rather than discovery).
> 
> -- https://www.w3.org/community/schema-course-extend/

Products:
- [x] DOC: "Schema.org: Online Courses" [Schema]  (@vholland, et al.)
  https://goo.gl/2Sxxy3
  - Thing > Intangible > Course
  - Thing > Intangible > CourseSection
  - Thing > Event > EducationEvent
  - [UPDATE]
  - [{ }] Thing > Intangible > Course
  - [{ }] Thing > CreativeWork > Course
  - [{ }] Thing > Product > Course
- [x] DOC: "Schema for Courses" [Use Cases] (@philbarker, et al.)
  https://goo.gl/DnTTsb
  - [x] https://www.w3.org/community/schema-course-extend/wiki/Outline_use_cases
[...]

PR #1317, opened 2016-08-18
add examples for Course related terms

philbarker:
adding in examples for Course related terms in new file structure, see  #195


danbri:
Thanks! Let's fill out the @@@@ bits before we wrap this up, but looks good :)

Issue #1329, opened 2016-08-25
Test to catch terms that are subtype/property of an attic term


RichardWallis:
Term should not be sub types or sup properties of terms that have been moved to the attic area.

danbri:
See also http://dydra.com/danbri/schema-org-3-1-sdo-makemake/@query#supersededby for potential terms we might migrate.

danbri:
Looks like we did this. 
```
  def test_superTypeInAttic(self):
    nri1= ('''select ?term ?super where { 
       {
           ?term rdfs:subClassOf ?super.
       }
       UNION
       {
           ?term rdfs:subPropertyOf ?super.
       }
       ?super <http://schema.org/isPartOf> <http://attic.schema.org> .
       FILTER NOT EXISTS { ?term <http://schema.org/isPartOf> <http://attic.schema.org> .}
    }
    ORDER BY ?term  ''')
    nri1_results = self.rdflib_data.query(nri1)
    if len(nri1_results):
        log.info("Super-term in attic errors!!!\n")
        for row in nri1_results:
            log.info("Term '%s' is sub-term of %s a term in attic" % (row["term"],row["super"]))
    self.assertEqual(len(nri1_results), 0, "Super-term in attic  Found: %s" % len(nri1_results))
```

Issue #1205, opened 2016-06-14
Add a unit test warning if a property is inverseOf itself


danbri:
(I guess this could be legit but is much more likely to be a bug...)

danbri:
We did it! Closing.

```
  def test_non_equal_inverseOf(self):
    results = self.rdflib_data.query("select ?x ?y where { ?x <http://schema.org/inverseOf> ?y }")
    for result in results :
      self.assertTrue(result[0] != result[1],
                      "%s should not be equal to %s" % (result[0], result[1]) )
```

... in tests/test_graphs.py

Issue #250, opened 2015-01-21
Add a unit test ensuring sentences end in '.'.


danbri:
Migrated in from https://www.w3.org/2011/webschema/track/issues/29

```
    Schema.org sentences sometimes end '.', and sometimes just stop 

    It would be good to be more consistent. See also this ad-hoc SPARQL database for details:

    http://dydra.com/danbri/schema-org/sparql#comments-not-ending

    SELECT * WHERE { GRAPH ?G { ?p rdfs:comment ?c } FILTER regex(str(?c), "[^.]$") } 

    http://dydra.com/danbri/schema-org/sparql#comments-ending

    SELECT * WHERE { GRAPH ?G { ?p rdfs:comment ?c } FILTER regex(str(?c), ".$") } 
```

We subsequently decided that descriptions end in '.', but labels do not.

SPARQL unit tests live here: https://github.com/schemaorg/schemaorg/blob/master/tests/test_graphs.py

danbri:
I filed a separate issue to deal with the backlog of existing definitions (and a commit fixing most of them), see https://github.com/schemaorg/schemaorg/issues/294

danbri:
This doesn't *really* matter but it is a great starter issue for someone interested in getting to know schema.org infrastructure better...

danbri:
@RichardWallis has added a test for '.' but the data still needs fixing, so I marked it as "expectedFailure" for now.

danbri:
@RichardWallis I'm closing this, but we should work on the "expected failures" :)

PR #1330, opened 2016-08-26
Added several new sparql vocab tests

RichardWallis:
New Tests:

- Test if a property is inverseOf itself - Issue(#1205)
- Test supercededBy & inverseOf don't point to same thing - Issue (#156)
- Test comments end on '.' - Issue (#250)
- Test for terms that are sub-type/property of an attic term - Issue (#1329) 
- Test for terms that (domainIncludes | rangeIncludes | inverseOf | supercededBy) reference an attic term
- Test for nonexistent super-types
- Test for nonexistent super-properties
- Test property name begins with lowercase
- Test type name begins with upper case
- Test Enumerations have defined enum values - Issue (#485)
- Test for terms isPartOf more than one extension

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,6 +1,5 @@
 #application: schemaorgae
-#application: webschemas
-application: sdo-rjwtest1
+application: webschemas
 
 version: 1
 runtime: python27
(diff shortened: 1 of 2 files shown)


RichardWallis:
Before merging it might be worth addressing the following errors identified in testing:
https://gist.github.com/RichardWallis/34709398bcb2160fc03759675bab9297

PR #1331, opened 2016-08-26
Add Subclass of Creative Work to Computer Language

thadguidry:
So that we have access to nice properties like 'creator' and 'dateCreated' among others.

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -1226,6 +1226,7 @@ See also the &lt;a href=&quot;/docs/hotels.html&quot;&gt;dedicated document on t
       <span class="h" property="rdfs:label">ComputerLanguage</span>
       <span property="rdfs:comment">This type covers computer programming languages such as Scheme and Lisp, as well as other language-like computer representations. Natural languages are best represented with the [[Language]] type.</span>
        <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Intangible">Intangible</a></span>
+       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/CreativeWork">CreativeWork</a></span>
     </div>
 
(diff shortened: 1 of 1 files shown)


danbri:
Could equally be considered a Product? How about we look at how many properties are directly applicable and consider adding them directly?

thadguidry:
@danbri if we go the direct route, we might piss off @mfhepp :)  but I would be OK with direct properties as well.

vholland:
Out of curiosity, is there a use case for this? It might help ground the discussion.

danbri:
I can think of use cases in a Wikidata or KG direction e.g. http://rigaux.org/language-study/diagram.html family tree of programming languages. Less obvious for our core Web search-y usecases, but it wouldn't be the first time we've strayed from that zone.

It might be (with extensions etc.) that we're getting to a level of richness where schema.org + extensions are useful for structuring databases rather than just annotating pages. I made a quick demo this morning to show how to populate a SPARQL db with our schemas - https://github.com/schemaorg/schemaorg/wiki/BlazeGraphSPARQLHowto

thadguidry:
@vholland The use case is to 2 parts.

1. align with Wikidata mappings (and Infopanel on Wikimedia) that I am working towards

2. Align with Google's Base Info efforts for Computer Language entities and other entities (as well as Linked Data efforts starting at ACM.org and IEEE.org) ... Base Info is now showing like in this search: [https://www.google.com/#q=D+language](https://www.google.com/#q=D+language)  Notice the DEVELOPER, TYPING DISCIPLNE, STABLE RELEASE, and OS.

In particular, the Developer (creator) property is used as a reconciling qualifier for a computer language along with dateCreated, since funny enough, a few languages have the same name (some are older out of date historical languages that never reached full maturation).

I cannot speak for ACM or IEEE directly, but know of their efforts (helping search engines, by first helping themselves with Linked Data aggregation)

PR #1336, opened 2016-08-31
Fixed typos in JSON-LD example in extensions.html

thadguidry:
@danbri @RichardWallis The JSON-LD context file is updated now with bib from what I saw .... I wonder if this JSON-LD example in extension.html should be cleaned up even more than these typo fixes I did ?


RichardWallis:
extension.html has had major rework that removed this section.

Issue #528, opened 2015-05-22
Product dimensions should be length x width x height


ghost:
Currently the length property is not specified in the Product Schema.
length, width, height are more common than height, width, depth

For example these pallet specifications:
http://www.palletways.com/services/pallet-specifications/

Issue #544, opened 2015-05-26
Feature: Product Schema => Volume [ Vitamin D Milk 1gal ]


inetbiz:
Many products are sold by volume. Lotions, Perfumes, Liquid Consumables, etc. We need to be able to express product volume.

mfhepp:
This is readily supported via 

a) http://schema.org/includesObject ->  http://schema.org/TypeAndQuantityNode (with this you can specify exactly how much of an item is included in an offer) 

b) http://schema.org/priceSpecification -> http://schema.org/UnitPriceSpecification -> http://schema.org/eligibleQuantity -> http://schema.org/QuantitativeValue for indicating the validity of a price by non-overlapping quantities.

If you want to model quantity discounts, simply attach two UnitPriceSpecification entities to the offer with differing prices and non-overlapping quantities for eligibleQuantity.

There are many examples given at http://wiki.goodrelations-vocabulary.org/Cookbook/Pricing.

You only have to replace the GoodRelations identifiers by the ones in schema.org, as listed here:

http://wiki.goodrelations-vocabulary.org/Cookbook/Schema.org#Naming_Differences

inetbiz:
@mfhepp thanks for that. But, if product schema is going to have Height, Width, Depth, Weight. There is no reason not to include volume.

mfhepp:
Ah, I understand. You want a volume property for product.

As of now, you can immediately use additionalProperty with a suitable UN/CEFACT code for indicating the volume of a liquid.

I hesitate to add schema:volume to schema:Product, because 

1. this is a property that applies only to small part of all products (not to electronics, cars, books, etc.) and
2. as long as search engines only need to display that value (but not use it for computations), schema:additionalProperty will serve as well.

Without any data at hand, I would argue that only a minority of products sold in Web shops are liquid.

If we go that route, I would rather propose a schema:productQuantity property with a range of schema:QuantitativeValue. This could then be used for arbitrary units of measurement.

It would also be a nice addition to the existing mechanism for quantities in GoodRelations, because in the current model, you can only specify the amount of an object contained in the offer. But for boxed products, it may indeed make sense to indicate the quantity contained in the boxed product.

@danbri @vholland - what do you think about schema:productQuantity or schema:containedQuantity property for schema:Product? We could also wait for a GS1 extension, which might provide this detail. Practically, schema:additionalProperty should do the trick for search engines for the moment.

DDeering:
[...]

PR #1339, opened 2016-09-02
We don't have length/thickness under Product

thadguidry:
We need to clarify that depth is also known as thickness.
In expanded form of the depth description, we could even say 'the depth or thickness or length of the item'.  But this begs the question of how not to confuse thickness with height

In general terms, its left/right, front/back, top/bottom.

In the States, we typically express those same general terms of a Products' dimensions as length/width/height  .. but for Product we currently have depth/width/height.  I think for developers, that causes more harm than good in trying to align with those general terms.

Proposal:
1. Rename depth to length instead. (or update its description to say 'the depth or length of the item'

Also found in the complaint #528 Product dimensions should be length x width x height

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -4490,7 +4490,7 @@ The place is __open__ if the [[opens]] property is specified, and __closed__ oth
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/depth">
       <span class="h" property="rdfs:label">depth</span>
-      <span property="rdfs:comment">The depth of the item.</span>
+      <span property="rdfs:comment">The depth or thickness of the item.</span>
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Product">Product</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Distance">Distance</a></span>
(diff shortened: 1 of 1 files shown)


mfhepp:
Hmm. We already have properties for three dimensions. The naming might be suboptimal for some cases and straightforward for others. 

I see two options:

1. We revise the model and add a clean yet lightweight model for 3D geometries, because the current model works only for rectangular, solid shapes, and supports only a single shape per product (e.g. no packing vs. product size).

2. We recommend the use of additionalProperty for additional measures by crafting examples.

I tend to recommend #2 because a clean model for 3D shapes will be quite some effort and likely hard to populate from existing databases.

Martin

thadguidry:
+1 on using 2 , the use of additionalProperty with some good examples.

danbri:
ok where are we with this? is the proposal to just updated the description? What about the mention of additionalProperty? 

/cc @vholland @nicolastorzec @scor @rvguha @tilid @tmarshbing

danbri:
(we're also btw now committing changes direct to master branch, so any PR will need to be targetted there, sorry for the administrivia)

vholland:
This and issue #544 about Product quantity seemed to split our collective energy. I am not sure if we should try to merge them under one issue or just be aware there are two.

Issue #1341, opened 2016-09-02
RDFa expression of equivalent term mappings not showing up


danbri:
We collect mappings to other schemas (and these are showing up in the dumps, see https://github.com/schemaorg/schemaorg/issues/1337#issuecomment-244339641 ) but apparently not currently in per-term RDFa pages.

At least types like our Event -> Dublin Core event used to have in-page owl equivalences inline in RDFa. Looking at https://github.com/schemaorg/schemaorg/blob/sdo-callisto/sdoapp.py#L369 we should have properties too. This seems not to be working.

@RichardWallis can you investigate + figure out an approach to unit testing such things so functionality can't silently break?

RichardWallis:
Fixed in PR (#1345)

Need a bigger test issue to identify such changes as this was effectively an unknown unknown.

RichardWallis:
Fixed

PR #1345, opened 2016-09-05
Map owl prefix in load from rdflib graph

RichardWallis:
Fix for issue (#1341)
Missed when integrating rdflib 'owl:' was not recognised as a used prefix.


danbri:
Merged. Curious, what's the thinking behind this idiom:

> l > 42 and (str[:43] == 'http://www.w3.org/1999/02/22-rdf-syntax-ns#')

... it seems longwindly alien to Python, but I didn't poke around the context.

RichardWallis:
A bit of boilerplate pasted from somewhere - could have used startswith and a list of candidate mappings (available in the namespaces list in api.py) - maybe I'll simplify it next time I'm working in that area ;-)

PR #1348, opened 2016-09-06
Missing word in description of object

darobin:
A tiny change, but it was missing.


danbri:
Ladies and gentlemen, we have a *reader* amongst us. And a do-er. Thanks! I wonder how many people read this previously and missed it :)

PR #1366, opened 2016-09-14
Changed RDF/XML download file suffix from .xml to .rdf.

RichardWallis:
Also documented conneg access on developers page.  
Added link to developers page on Documentation page.


danbri:
Merged, although in future let's try to figure out a workflow that is less noisy (w.r.t. dump file versioning).

Issue #1415, opened 2016-10-26
Recognise the support of Legal Entity Identifier Foundation for leiCode property


RichardWallis:

danbri:
Nice to hear that they are supportive!

sopekmir:
If it is possible this PR should be included into the coming publication as it makes an imprtant clarification who is responsible for LEI codes (GLEIF Foundation). I made a note at #1416 about that.

PR #1416, opened 2016-10-26
Added acknowledgement of support from Legal Entity Identifier Foundation

RichardWallis:
… for the leiCode property.

Re issue (#1415)


sopekmir:
Dear All,

If it is possible this PR should be included into the coming publication as it makes an imprtant clarification who is responsible for LEI codes (GLEIF Foundation).

Issue #1288, opened 2016-08-09
Proposal of a Menu/menuItems schema?


dkutcher:
Currently for Restaurants and FoodEstablishments there is a menu item:

http://schema.org/Restaurant
menu
Text  or URL
Either the actual menu or a URL of the menu.

But no schema for the menu(s) and menu items themselves.

For example usage, google a restaurant name + menu such as https://www.google.com/?ion=1&espv=2#q=two%20urban%20licks%20menu

You'll see the data of a menu being used in the rich formatting.

Is anyone developing the menu and menuItem schema? If not, how does one propose?

jvandriel:
What if we added ```schema.org/OfferCatalog``` to ```schema.org/menu```'s name? 

Than we could do something like this:
```
<script type="application/ld+json">
{
  "@context":"http://schema.org/",
  "@type":"Restaurant",
  "name":"TWO urban licks",
  "description":"TWO serves wood-fired meats and fish served in a high-energy, open kitchen featuring fiery American Food.Executive Chef at TWO, Cameron Thompson gets to let his creativity shine as his passion for innovation and desire to exceed diners’ expectation enliven the “Fiery American Cooking” praised by Atlantans year after year. All of the small plates are ideal for sharing or savoring on your own. A few of the menu’s highlights include the Salmon Chips, loaded with short smoked salmon, chipotle cream cheese, capers, and red onion; Lamb Lollipops served with grape chile jam and goat cheese; entrée options include Pork Shoulder with NY baked cheddar macaroni and pork jus; and the Bronzed Sea Scallops served over smoked gouda grits in a tomato broth.",
  "url":"http://places.singleplatform.com/two-urban-licks/menu",
  "menu":
  [
    {
      "@type":"OfferCatalog",
      "name":"Main menu",
      "itemListElement":
      [
        {
          "@type":"Offer",
          "currency":"USD",
          "price":"10",
          "itemOffered":
          {
            "@type":["Product","Recipe"],
            "name":"Salmon Chips",
            "recipeIngredient":["smoked salmon","chipotle cream","cheese","caper","red onion"]
          }
        },
        {
          "@type":"Offer",
          "currency":"USD",
          "price":"12",
          "itemOffered":
          {
            "@type":["Product","Recipe"],
            "name":"Little Necks",
            "recipeIngredient":["steamed clams","mezcal","charred jalapeno","smokey bacon"],
            "nutrition":
            {
              "@type":"NutritionInformation",
              "name":"*",
[...]

PR #1432, opened 2016-11-17
Adding an add-on property to MenuItem (#1288)

gmackenz:
We need to directly connect add-ons such as side dishes for MenuItems, It can link to MenuItem or to MenuSection that contains allowable add-on MenuItems. Issue is #1288

Proposal doc: https://docs.google.com/document/d/1q0M8jdJrYZJHvD_sQXvhkGtG07XmzyPgmjnUad3lGD0/edit#heading=h.l6sgyrdwxqp6

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -11221,7 +11221,7 @@ Typical unit code(s): C62 for person</span>
     </div>
     
     <div typeof="rdf:Property" resource="http://schema.org/menuAddOn">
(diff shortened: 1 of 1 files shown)


gmackenz:
I do wonder if the property should be 'hasMenuAddOn' instead of 'menuAddOn'.

danbri:
Thanks, makes sense. I wouldn't bother with prepending 'has' unless to avoid same term being used for a property and a type.

Issue #1423, opened 2016-11-02
Add "WebAPI" type


IvanGoncharov:
At the moment Schema.org don't have appropriate type to describe Web APIs.
By WebAPI I mean any API which uses HTTP/HTTPS as transport mechanism including REST API, SOAP, GraphQL, etc.
At the moment there is no scalable mechanism to discover publicly available APIs on the internet.
Closest solution so far is API catalogs, and for the last two year, I have been maintaining [my own](https://APIs.guru).
But it's hard to maintain and impossible to scale.

On the other hand, almost all public APIs have dedicated HTML pages where they can use Schema.org types.
Ideally it should have fields like:
  - Title, Description, Logo
  - License, ToS
  - Support contacts
  - Version, e.g. `1.0.0`
  - list of base URLs, e.g. http://api.example.com/v1, https://api.example.com/v1/
  - Type of service: RESTful, SOAP, GraphQL
  - Link to [API Description](https://en.wikipedia.org/wiki/Overview_of_RESTful_API_Description_Languages)

I want to work on a PR adding `WebAPI` type to `schema.org`.
Can you please advice which base type should I derive it from?
And should I submit it as PR or you have some formal procedure?

jaygray0919:
This would complement work done here:
  http://webdatacommons.org/structureddata/

IvanGoncharov:
@jaygray0919 It will benefit entire developers community since a huge amount of data can be accessed only through Web APIs. And for many companies describing their existing Web APIs will be the first step toward Linked Data.

IvanGoncharov:
@danbri @RichardWallis I'm ready to start working on PR, can you please answer my questions?
> Can you please advice which base type should I derive it from?
> And should I submit it as PR or you have some formal procedure?

RichardWallis:
@IvanGoncharov apologies for being a bit slow in responding.

I can see the basic need for something like this - I would like to see some indication of who would apply it to their sites (other than yourself of course).

Most of the properties you describe could be generally available via any Type in Schema (Title (name), description, URLs, etc.) and via the organization Offering the API (Service?) (Logo, Contacts etc.)

As to a base type to enhance or subtype, my inclination is to look at [EntryPoint](http://schema.org/EntryPoint) which is generally describing something similar.  Not sure if we just enhance that type with a few more specific properties, or create a more specific subtype for API (do we need to constrain it to WebAPI ?)

[...]

PR #1445, opened 2016-11-28
Create a new WebAPI class and some useful properties

lanthaler:
This adds a new WebAPI class to Schema.org's core vocabulary. I decided to subclass `Intangible > Service` as a distinct feature of a Web API in contrast to normal *application* programming interfaces is that they are always offered by a `provider`.

Previous discussions in #1423 mentioned a couple of fields, most of them are addressed by this design:

- Title: `Thing > name`
- Description: `Thing > description`
- Logo: `Service > logo`
- License: I don't see how a license applies to a Web API!?
- ToS: newly added `Service > termsOfService`
- Support contacts: available through `Service > provider`
- Version: ~IMO breaking down an API by versions isn't very useful. Either describe multiple WebAPIs or push this into the API documentation~ removed again after discussions; see thread below
- list of base URLs: should be part of API documentation, not the definition of the API IMO
- Type of service: should be part of API documentation
- Link to API Description: newly added `WebAPI > documentation`

@RichardWallis, I saw you suggested to specialize `EntryPoint`. I don't think that would work well as an `EntryPoint` is supposed to be a specific resource (URL), a WebAPI on the other hand is a collection of resources.

Just as @tlrobinson I also had a few concerns about the Web in `WebAPI` but I think what it is supposed to convey here is that it is *somehow* accessible over the Web/Internet. I tried to clarify that in the description of the class.

Last but not least, things like versions, technologies (SOAP, GraphQL), REST maturity level, dependsOn etc. are beyond the scope of this IMO and should be addressed by the `documentation` of the `WebAPI`. I thus made the range of `documentation` a URL. This provides a great extension point for the future and allows people to start pointing to both human-readable and machine-readable docs.

### DIFF
--- a/data/ext/pending/issue-1423-examples.txt
+++ b/data/ext/pending/issue-1423-examples.txt
@@ -1,0 +1,30 @@
+TYPES: WebAPI, documentation, termsOfService
+
+PRE-MARKUP:
+
+See JSON example.
+
+MICRODATA:
+
+TODO
+
+RDFA:
+
+TODO
+
+JSON:
+
+<script type="application/ld+json">
+{
+  "@context": "http://schema.org/",
+  "@type": "WebAPI",
+  "name": "Google Knowledge Graph Search API",
+  "description": "The Knowledge Graph Search API lets you find entities in the Google Knowledge Graph. The API uses standard schema.org types and is compliant with the JSON-LD specification.",
+  "documentation": "https://developers.google.com/knowledge-graph/",
+  "termsOfService": "https://developers.google.com/knowledge-graph/terms",
+  "provider": {
+    "@type": "Organization",
+    "name": "Google Inc."
+  }
+}
+</script>
--- a/data/ext/pending/issue-1423.rdfa
+++ b/data/ext/pending/issue-1423.rdfa
@@ -1,0 +1,26 @@
+<div>
+  <!-- WebAPI, issue 1423 -->
+
+  <div typeof="rdfs:Class" resource="http://schema.org/WebAPI">
+    <span class="h" property="rdfs:label">WebAPI</span>
+    <span property="rdfs:comment">An application programming interface accessible over Web/Internet technologies.</span>
+    <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Service">Service</a></span>
+  </div>
+
+  <div typeof="rdf:Property" resource="http://schema.org/documentation">
+      <span class="h" property="rdfs:label">documentation</span>
+      <span property="rdfs:comment">Further documentation describing the Web API in more detail.</span>
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/WebAPI">WebAPI</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/CreativeWork">CreativeWork</a></span>
+  </div>
+
+  <div typeof="rdf:Property" resource="http://schema.org/termsOfService">
+      <span class="h" property="rdfs:label">termsOfService</span>
+      <span property="rdfs:comment">Rules by which one must agree to abide in order to use the service</span>
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Service">Service</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
+  </div>
+
+</div>
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -11219,7 +11219,7 @@ Typical unit code(s): C62 for person</span>
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/MenuSection">MenuSection</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/MenuSection">MenuSection</a></span>
     </div>
-    
+
     <div typeof="rdf:Property" resource="http://schema.org/menuAddOn">
       <span class="h" property="rdfs:label">menuAddOn</span>
       <span property="rdfs:comment">Additional menu item(s) such as a side dish of salad or side order of fries that can be added to this menu item. Additionally it can be a menu section containing allowed add-on menu items for this menu item.</span>


chaals:
> * Version

…

> I don't think [`EntryPoint`] would work well as an `EntryPoint` is supposed to be a specific resource (URL),

Hmm. I think version information is pretty basic and important for APIs, which might change functions or how you use them from version to version. Similarly, extending `EntryPoint` makes sense to me, but overall Service seems a clearer fit

Nice work - thanks

RichardWallis:
I would add [CreativeWork](http://schema.org/CreativeWork) to the range of _documentation_, this would allow the use of [TechArticle](http://schema.org/TechArticle) or even video instructions to be described.

I also agree that version information could be important - perhaps extending the domain of [softwareVersion](http://schema.org/softwareVersion) to include WebAPI would work.

I'm still not sure if _Service_ is the correct type to extend.  In terms of the english name of the type it makes sense, as to the properties that you inherit from it, it is not as obvious to me.

IvanGoncharov:
@lanthaler Thanks a lot.

> Last but not least, things like versions, technologies (SOAP, GraphQL), REST maturity level, dependsOn etc. are beyond the scope of this IMO and should be addressed by the documentation of the WebAPI. I thus made the range of documentation a URL. This provides a great extension point for the future and allows people to start pointing to both human-readable and machine-readable docs.

I fully agree that technical details should be in a separate entity since many API owners provide multiple versions/protocols to access the same API service. For example SMSglobal supports: [SOAP API](http://smsglobal.com/soap-api/), [REST API v2](http://smsglobal.com/rest-api/), [REST API v1](http://smsglobal.com/rest-api/v1/), etc.

However WebAPI should provide an array of possible ways for calling this API (e.g. RESTv1, RESTv2, SOAP, GraphQL). Maybe we can use [availableChannel](http://schema.org/availableChannel) for that?

lanthaler:
>> I think version information is pretty basic and important for APIs, which might change functions or how you use them from version to version.
>
> I also agree that version information could be important - perhaps extending the domain of softwareVersion to include WebAPI would work.

I can't see how it would be useful but I don't have a strong opinion about it. I extended the domain of `softwareVersion` to include `WebAPI` as suggested by @RichardWallis 

> I would add CreativeWork to the range of documentation, this would allow the use of TechArticle or even video instructions to be described.

Good catch. Changed.

> I'm still not sure if Service is the correct type to extend.  In terms of the english name of the type it makes sense, as to the properties that you inherit from it, it is not as obvious to me.

I agree some of Service's properties such as `hoursAvailable` and `areaServed` won't be very useful for a WebAPI but they are still fine I guess. Typical Web APIs will be available 24/7 world-wide but I can definitely think of use cases/scenarios were that isn't true. Which other properties do you find non-obvious?

> However WebAPI should provide an array of possible ways for calling this API (e.g. RESTv1, RESTv2, SOAP, GraphQL).

@IvanGoncharov, I understand were you are coming from but I strongly suggest to not include this here but push it into the API documentation. We will end up in endless discussions of what those ways are (RESTv1 & RESTv2 are good instances of examples triggering such discussions) and consumers can't make use of this without additional details... which they will find in the documentation anyhow.

If the only use case is the discovery of such APIs (without needing means to verify whether an API is indeed RESTful or not), simple keyword search on the description or title of the API probably brings us a long way.

asbjornu:
>> I also agree that version information could be important
> I can't see how it would be useful but I don't have a strong opinion about it.

I agree with @lanthaler in that I don't see the value of having a version number here.

>> However WebAPI should provide an array of possible ways for calling this API (e.g. RESTv1, RESTv2, SOAP, GraphQL).
> @IvanGoncharov, I understand were you are coming from but I strongly suggest to not include this here but push it into the API documentation.

If we are going to hold to any definition of REST, then Dr. Roy Thomas Fielding wrote that definition 16 years ago. Either we stick to it, or we best avoid the issue altogether.

> We will end up in endless discussions of what those ways are (RESTv1 & RESTv2 are good instances of examples triggering such discussions) and consumers can't make use of this without additional details... which they will find in the documentation anyhow.

Indeed. 👍

IvanGoncharov:
>We will end up in endless discussions of what those ways are (RESTv1 & RESTv2 are good instances of examples triggering such discussions) and consumers can't make use of this without additional details... which they will find in the documentation anyhow.

You missed my main point. It's not about REST, it's about the fact that API owners provide multiple interfaces to access the same service. To keep this discussion in the context I will illustrate the problem without mentioning `REST` word.

Let's use Dropbox as a better example. It has [APIv2](https://www.dropbox.com/developers) and [APIv1](https://www.dropbox.com/developers-v1). Should Dropbox create separate instances for each API version? And what about Facebook which currently supports eight previous API versions(v2.1-v2.8)?

Search results may be polluted with multiple versions of the same API. Moreover, ratings and review will be lost when new API version appears. 

Let's discuss another scenario when you maintain a single instance for each supported API versions.
> I would add CreativeWork to the range of documentation, this would allow the use of TechArticle or even video instructions to be described.

In the case of single instance, the API owner will have to link all the document URLs for every supported version into one huge list.

**Proposal**: To have the main object for API in general and a subobject for every supported version.
Similar to how [Service](http://schema.org/Service) and [ServiceChannel](http://schema.org/ServiceChannel) are organised.

lanthaler:
> Let's use Dropbox as a better example. It has APIv2 and APIv1. Should Dropbox create separate instances for each API version?

Are they the same? I guess not and so they should get separate instances.

> And what about Facebook which currently supports eight previous API versions(v2.1-v2.8)?

These are minor variations... having one v2 should suffice. This is one of the reasons why I think having a version isn't very helpful.

> Search results may be polluted with multiple versions of the same API.

Doesn't seem to be an issue for products. Also, there's `isSimilarTo` if you want to express that two APIs are similar.

> Moreover, ratings and review will be lost when new API version appears.

Which is a good thing IMO because they are different things after all.

> Let's discuss another scenario when you maintain a single instance for each supported API versions.
> 
>> I would add CreativeWork to the range of documentation, this would allow the use of TechArticle or even video instructions to be described.
> 
> In the case of single instance, the API owner will have to link all the document URLs for every supported version into one huge list.

So for 10 versions we would have 1 instance of WebAPI with 10 links under `documentation`, correct? Doesn't sound bad to me. It also doesn't prevent anyone to point to developer centers homepage which then in turn links to more detailed pages.

> Proposal: To have the main object for API in general

What things qualify as general? Will the description be the same for all version? What about the logo?

> and a subobject for every supported version.
> Similar to how Service and ServiceChannel are organised.

I can't see an advantage of "subobjects" over separate instances.

As I already mentioned, I would suggest to keep this as simple as possible for the time being and improve and refine the design as we learn more based on real-world usage.

lanthaler:
Do we need any further changes to the pull request?

IvanGoncharov:
>As I already mentioned, I would suggest to keep this as simple as possible for the time being and improve and refine the design as we learn more based on real-world usage.

Yes, I understand.

But we want to continue working on extending `WebAPI`. So I have initiated W3C `Web API Discovery` community group. Would it be possible to submit group's work as a hosted extensions similar to `bib.schema.org` or `auto.schema.org`?

> These are minor variations... having one v2 should suffice. This is one of the reasons why I think having a version isn't very helpful.

It's enough difference for Facebook to have different base URLs: https://graph.facebook.com/v2.2/,
https://graph.facebook.com/v2.3, etc. So I strongly vote for removing `softwareVersion` since it can lead to a lot of duplications and also be harmful to future extension. 

> Do we need any further changes to the pull request?

After removing `softwareVersion` this PR will be a good starting point 👍

lanthaler:
@chaals @RichardWallis any objections to removing `softwareVersion`?

RichardWallis:
In principle yes.   I am presuming that ```isRelatedTo``` would be used to reference between  https://graph.facebook.com/v2.2/ and https://graph.facebook.com/v2.3/.

I would like to see some of the examples you intend to publish with the proposal.

@IvanGoncharov have you crated the W3C Community Group yet?  I can't seem to find it.

IvanGoncharov:
@RichardWallis W3C editor contacted me and ask us to expand group description.
So I expect it to be added in a few days.

lanthaler:
In case you haven't seen it yet there exists the [Hydra W3C Community Group](https://www.w3.org/community/hydra/) which works in this area as well.

lanthaler:
I removed `WebAPI` from the domain of `softwareVersion`. @RichardWallis, here's how this could look like in real-life :-)

```json
{
  "@context": "http://schema.org/",
  "@type": "WebAPI",
  "name": "Google Knowledge Graph Search API",
  "description": "The Knowledge Graph Search API lets you find entities in the Google Knowledge Graph. The API uses standard schema.org types and is compliant with the JSON-LD specification.",
  "termsOfService": "https://developers.google.com/knowledge-graph/terms",
  "provider": {
    "@type": "Organization",
    "name": "Google Inc."
  },
  "documentation": "https://developers.google.com/knowledge-graph/"
}
```

I will turn that into an example of ACME Corp. and add it to the PR in the various syntaxes right before we agree to merge this :-)

lanthaler:
@RichardWallis, @chaals, @danbri everything else we need to get done before we can merge this PR?

IvanGoncharov:
> have you crated the W3C Community Group yet? I can't seem to find it.

@RichardWallis It's launched 🎆 
https://www.w3.org/community/web-api-discovery/

> In case you haven't seen it yet there exists the Hydra W3C Community Group which works in this area as well.

@lanthaler skimmed through Hydra spec, great work 👍
I really like that Hydra API is self-documenting and how you use `Link` header to discover API itself.
But it's not possible to automatically get a list of public Hydra APIs.
This is where schema.org can be very useful to point existing machine-readable description for API. So our intention is to compliment formats like Hydra, OpenAPI, GraphQL, etc.

danbri:
LGTM, merging into pending for more visibility.

jaw111:
Does it make sense to add `WebAPI` to the range of `distribution` to allow to link from a `Dataset` to the API through which it can be accessed?

lanthaler:
Sounds reasonable. It would require the description of `distribution` to be tweaked a bit though IMO:

> A downloadable form of this dataset, at a specific location, in a specific format.

JervenBolleman:
Would it not be more natural to go from API to Dataset than the other way round? Many tools/api's allow searching through a Dataset but don't offer the original data back in any way.

smrgeoinfo:
I think WebAPI in range of distribution makes good sense; many datasets are accessible via multiple API. @JervenBolleman 's point above is valid, but API's that don't offer original data back would not be distributions.

nickevansuk:
So there's a [revision to this work](https://github.com/schemaorg/schemaorg/pull/2635), set for [inclusion in the next schema.org release](https://github.com/schemaorg/schemaorg/issues/2648), that has one outstanding issue that we're seeking consensus on.

It would be great to get your opinion on https://github.com/webapi-discovery/rfcs/issues/10, in the context of [this draft](https://webapi-discovery.github.io/rfcs/rfc0001.html).

Any other feedback welcome of course!

PR #1459, opened 2016-12-16
Improvements to TouristAttraction

RichardWallis:
Following previous discussions on mailing lists over the last couple of months this is a proposal from the [W3C Tourism Structured Web Data Community Group](https://www.w3.org/community/tourismdata/), working with Angelica Lo Duca and Elisabetta Triolo, to increase the utility and usefulness of the TouristAttraction Type.

A working example, with many examples for several use-cases, can be found here: [TouristAttraction](http://sdo-tourism.appspot.com/TouristAttraction).

By making use of Multi Type Entity (MTE) capabilities, it should be possible to identify any 'Thing' as a TouristAttraction, in addition to its default properties as a museum, mountain, dance festival, or whatever.

To aid that utility, this proposal makes the following additions: 

- Add new property [touristType](http://sdo-tourism.appspot.com/touristType) to TouristAttraction  "_Attraction suitable for type(s) of tourist. eg. Children, visitors from a particular country, etc._"

- Extend the domain of [availableLanguage](http://sdo-tourism.appspot.com/availableLanguage) to TouristAttraction.  "_A language someone may use with or at the item, service or place...._"

- Extend the domain of [isAccessibleForFree](http://sdo-tourism.appspot.com/isAccessibleForFree) to Place  "_A flag to signal that the item, event, place is accessible for free._"

- Add new  property [publicAccess](http://sdo-tourism.appspot.com/publicAccess) to Place.  "_A flag to signal that the Place is accessible by public visitors._"

Although the target type for this proposal is TouristAttraction, it was felt that applying _isAccessibleForFree_ & _publicAccess_ to _Place_ would deliver further utility.

This proposal does not attempt to address individual enhancements to types that could be a TouristAttraction, for example a _touristAuthority_ property, or a _typeOfSand_ property being added to the _Beach_ type.   These may well be approached in later proposals after more thought and discussion.

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -10511,7 +10511,7 @@ Standards bodies should promote a standard prefix for the identifiers of propert
 
 <div typeof="http://schema.org/Organization" resource="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#Tourism">
     <span property="rdfs:label">Tourism</span>
-    <span property="rdfs:comment">This element is based on the work of the [Tourisum Structured Web Data Community Group](https://www.w3.org/community/tourismdata).</span>
+    <span property="rdfs:comment">This element is based on the work of the [Tourism Structured Web Data Community Group](https://www.w3.org/community/tourismdata).</span>
   </div>
 <div typeof="http://schema.org/Organization" resource="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#IIT-CNR.it">
     <span property="rdfs:label">IIT-CNR.it</span>
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -2022,8 +2022,21 @@ See also the &lt;a href=&quot;/docs/hotels.html&quot;&gt;dedicated document on t
 
     <div typeof="rdfs:Class" resource="http://schema.org/TouristAttraction">
       <span class="h" property="rdfs:label">TouristAttraction</span>
-      <span property="rdfs:comment">A tourist attraction.</span>
-       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Place">Place</a></span>
+      <span property="rdfs:comment">A tourist attraction.  In principle any Thing can be a [[TouristAttraction]], from a [[Mountain]] and [[LandmarksOrHistoricalBuildings]] to a [[LocalBusiness]].  This Type can be used on its own to describe a general [[TourstAttraction]], or be used as an [[additionalType]] to add tourist attraction properties to any other type.  (See examples below)</span>
+      <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/Place">Place</a></span>
+      <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#Tourism">Tourism</a></span>
+      <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#IIT-CNR.it">IIT-CNR.it</a></span>
+    </div>
+
+    <div typeof="rdf:Property" resource="http://schema.org/touristType">
+      <span class="h" property="rdfs:label">touristType</span>
+      <span property="rdfs:comment">Attraction suitable for type(s) of tourist. eg. Children, visitors from a particular country, etc. </span>
+      <link property="rdfs:subPropertyOf" href="http://schema.org/audience" />
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/TouristAttraction">TouristAttraction</a></span>
+      <span>rangeIncludes: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Audience">Audience</a></span>
+      <span>rangeIncludes: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Text">Text</a></span>
+      <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#Tourism">Tourism</a></span>
+      <span>Source:  <a property="dc:source" href="http://www.w3.org/wiki/WebSchemas/SchemaDotOrgSources#IIT-CNR.it">IIT-CNR.it</a></span>
     </div>
 
     <div typeof="rdfs:Class" resource="http://schema.org/TouristInformationCenter">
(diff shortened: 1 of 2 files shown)


vholland:
Is touristType different enough than the existing http://schema.org/audience property to warrant a new name?

thadguidry:
@RichardWallis Lets continue to try to use "Simple English" as much as possible.
To that end, please change "A **flag** to signal that" to "An **indicator** to signal that".

sismotur:
Hi @vholland, many thanks for your comment. My name is Felipe Santi, co-chair with Richard Wallis of the [W3C Tourism Structured Web Data Community Group](https://www.w3.org/community/tourismdata/).

Yes indeed I think that the touristType property is necessary to add, separately from the existing [audience](http://schema.org/audience) property. Based on our [experience](http://www.sismotur.com) working in large tourism destinations like Barcelona, this information is key to describe touristic places.

Tourist destinations need to profile and differentiate themselves to cater for a certain type of tourist, e.g. a destination like Ribera del Duero wants to attract "culture and wine tourism" tourists. On the other hand, people that want to plan their trips by themselves do also want to look for places filtered by this type of criteria. You can take the Cemetery example in [TouristAttraction](http://sdo-tourism.appspot.com/TouristAttraction) to see how it makes sense to a certain type of public.

Mixing this with audience (defined as `An intended audience, i.e. a group for whom something was created.`) would in my opinion be misleading, since touristType does not fit the definition, it is a different concept.

Hope this clarifies.
Kind regards,
Felipe

thadguidry:
@sismotur Its not misleading... its just that the definition of audience needs a bit of cleaning up.

+1 for reusing audience and expanding its definiton - An intended audience, group or type of people that some Thing is created for or meant to be attractive to.

RichardWallis:
@thadguidry I agree the 'a flag' wording is a bit odd, it is already part of the _isAccessibleForFree_ description which I lazily copied into the publicAccess description.  We should probably follow your suggestion for both.

@vholland there is obvious connection between _audience_ and _touristType_ as recognised by the sub-typing in the proposal.  However I do believe that there is sufficient difference to justify a _touristType_ (or possibly a _visitorType_) property.   

In my view, audience is about targeting of a thing/event, as per the current description: _a group for whom something was created_   Whereas touristType is more about the group or groups of people that may actively search out the type of attraction.     

There are also situations where the Thing in it's own right will have a target audience (theatre, show, festival, religious venue, etc.) yet as a TouristAttraction have a different profile of visitors (around architecture, aesthetics, history, activities, etc.)  A good example of this would be the Vatican for example.

sismotur:
@thadguidry thanks for your comment, you're right to point that the current definition of audience is not misleading, please let me re-express the argument again.

On top of the arguments just laid out by @RichardWallis, what I meant is that, in my opinion, I would not change the definition of Audience, since the word itself is not intended for that purpose. In my old English dictionary, the definition is "the people listening to or watching a performance, speech, television show, etc".

Therefore the suggestion to broaden it sounds to me a bit like hacking the word Audience to make it encompass more concepts than the ones the word is intended to, which can lead to confusion on both users writing schema.org and consumers of that information, who as Richard pointed, are a different profile of visitors and not a target audience of a performance or show.

thadguidry:
@sismotur @RichardWallis It is hacking (changing) the definition and for VERY good reason.

We want consumers and publishers to be comfortable with Schema.org, no matter which side they are on.  We cannot forgot the intention and mission of Schema.org, which is to help consumers ultimately "filter and find things easier" and in order to do that we have to give good vocabulary for publishers to "structure their data" to describe their Things such as TouristAttractions.

You can view this as the difference between A: "I'm looking for" and B: "I'm offering a" .. where both ultimately have the same Thing in the middle.  Your stance is that A makes more sense (consumer side) and I am proposing that we stick to our Schema.org mission approach of helping web developers describe B.

As a web developer marking up my Tourist website, I come to Schema.org and see 2 types that might work for me to describe who the intended group of people my TouristAttraction is most useful or attractive to...

"history buff" ... is that an audience or a touristType ?
"geek" ... is that an audience or a touristType ?
"gamer" ...
"naturalist" ...
"Doctor" ...
"environmentalist"
"Trump supporters"
"thrill seeker"

See where I am getting at ?
You will make it very confusing for publishers to choose audience or touristType, because there is little to no difference once we fix the definition of audience.

RichardWallis:
As a web developer marking up my tourist site, I would suggest that the answer to all your posed questions would be _touristType_, as that would be the property that would become available by  choosing the TouristAttraction Type.  (if the proposal stands as it is) 

I believe that 'defining' a tourist type as such, explicitly differentiates between the types of groups of people we are discussing here, whereas just relying on audience may well introduce some confusion in those that would not see tourists as audience members.

However, it would be unfortunate to delay this useful proposal by intransigence on any side.   

If the broad consensus is that the domain of _audience_ should be extended to include _TouristAttraction_ instead of creating the _touristType_ sub-property, I am sure the group would accept that as a compromise.   Subject however to getting the description of the audience property appropriately expanded.

The current, and proposed, description enhancements still retain the focus that something has been created with a group of people in mind.  "_some Thing is created for or meant to be attractive to_".  For a show, or event, etc,, or say children's furniture when used on a product, etc., this is perfectly valid.


In the area of tourism however, the intention is often on the part of the groups of tourists, not necessarily those describing or preparing the attraction.

The description we end up with needs to be broad enough to handle this breadth from the traditional theatre & TV program creative work point of view plus products and lodging businesses, yet still be appropriate from a tourism point of view to be useful to categorise a destination as being of interest to cruse-ship passengers, backpackers, military history enthusiasts, families of war cemetery residents, film location fans, nudist bathers, etc.,etc.  Whilst still being suitable when describing things from a glacier or a mountain, to a building, a street festival, or a theme park.

thadguidry:
@RichardWallis 

THING: Mount Everest
TYPE: touristAttraction
AUDIENCE: backpackers

THING: Les Miserables @ The MET New York
TYPE: Theatrical Play
AUDIENCE: general public

but your example on touristType does not describe an audience....
```
  "touristType": [
    "Wine tourism",
    "Cultural tourism"
  ],
```

Now that I actually looked at some of the examples, this is even worse...
```
"touristType": {
    "@type": "Audience",
    "audienceType" : "Memorial Tourism",
```

There is a difference between Tourism (INDUSTRY) and Tourist (PERSON).

I don't want to be the one to choose the definition expansion of audience.  Let @vholland handle that, she's much better than all of us at those things :)

thadguidry:
Ugh... and Event doesn't even have audience .... so we stupidly created tons of SubTypes like http://schema.org/ChildrensEvent just to say its intended audience is for Children ?  ugh... we're so bad. :)  Stop creating subtypes just for the sake of it.... empty bucket types that are more useful as let thought out and planned for properties.

@RichardWallis Add audience to Event while we are at this, for heavens sake. :)

RichardWallis:
```"audienceType" : "Memorial Tourism",``` 
@thadguidry Yes you are right to highlight that - if we do drop back to using audience the examples would have to change to reflect the type of tourist of  as against the form of tourism.

"Memorial Tourist", "Wine Tourist", etc.


... and yes to adding _audience_ to Event - just because it is a sensible thing to do!

RichardWallis:
Reflecting on the comments it is clear that the sub-property relationship between _touristType_ and _audience_ is not as strong as initially proposed.  Removed that link.

danbri:
We didn't get this into 3.2 but it looks like solid work, I'll take care to get it reviewed for next release.

danbri:
Was there a tracking issue, or all github discussion is here?

RichardWallis:
All in this PR - previous discussion was external within the [Tourism Structured Web Data Community Group ](https://www.w3.org/community/tourismdata/)

danbri:
Would you mind tweaking this PR to target the 'master' branch? I merged into into the wrong one in error (now reverted).

danbri:
Don't worry, I'll make a quick new PR. Easily done :)

Dataliberate:
Went to do it to find you had already done it!

PR #1467, opened 2017-01-06
Set up continous integration testing with Travis

lanthaler:
Travis needs to be enabled by one of the repo admins

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+
+python:
+  - "2.7"
+
+before_script:
+  - SDK_URL=https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.49.zip
+  - wget $SDK_URL -nv -o google_appengine.zip
+  - unzip -q google_appengine.zip
+  - rm google_appengine.zip
+  - export PATH=$PATH:DIRECTORY_PATH/google_appengine/
+
+script: python scripts/run_tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ before_script:
   - wget $SDK_URL -nv -O google_appengine.zip
   - unzip -q google_appengine.zip
   - rm google_appengine.zip
-  - export PATH=$PATH:$TRAVIS_BUILD_DIR/google_appengine/
-  - echo $PATH
-  - pwd
+  - export PYTHONPATH=$PYTHONPATH:$TRAVIS_BUILD_DIR/google_appengine/
 
 script: python scripts/run_tests.py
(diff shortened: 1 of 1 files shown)


danbri:
Hmm would you mind re-sending the PR targetting sdo-callisto?

lanthaler:
Sure, here it is: https://github.com/schemaorg/schemaorg/pull/1468

danbri:
Thanks again for this, it was super helpful :)

one more q: is there any way, within the running tests, to see the outer environment i.e. which branch we are being tested against? I'd like to use that to warn if a PR is sent to 'master' branch.

dbs:
I suggest simply adopting the standard git working practice of using master as the development branch against which would-be contributors should send pull requests. schema.org is one of the only projects I've participated in that breaks the "develop against master" pattern; adopting standard practice would reduce the friction required by newcomers to contribute to the project.

thadguidry:
@danbri I agree with dbs on that... make master the development branch.

danbri:
@dbs we have an organizational need for some kind of two-step staging, since schema.org itself doesn't change without its steering group approving each release-candidate bundle of integrated improvements. I was already thinking to ditch the release-named branches and have a single same-name-each-time dev branch, so I guess all we need to do is pick a new name for the "uptight", steering-group-reviewed branch from which schema.org is actually published. Is there an established convention for that? "releases", "published", etc? It is indeed time for a bit of a new year's refresh on these habits...

thadguidry:
@danbri GitFlow-based workflows use "release" and use "develop" for development. Here's a cheatsheet that thousands of devs refer to each day: http://danielkummer.github.io/git-flow-cheatsheet/ - look at finishing up a release ... using GitFlow would probably simplify your lives.

danbri:
"release" or "releases" sounds ok but now I'm getting mixed messages here on "master" vs "develop". @dbs any view?

danbri:
see also https://barro.github.io/2016/02/a-succesful-git-branching-model-considered-harmful/

lanthaler:
Sorry for the late reply, this completely slipped under my radar

> one more q: is there any way, within the running tests, to see the outer environment i.e. which branch we are being tested against? I'd like to use that to warn if a PR is sent to 'master' branch.

Yes, the [`TRAVIS_BRANCH`](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) environment variable tells you exactly that.

> "release" or "releases" sounds ok but now I'm getting mixed messages here on "master" vs "develop". @dbs any view?

To keep it simple and consistent with almost every other project on GitHub, I would use `master` for development and release (with tags) for officially blessed releases.

gkellogg:
The CONTRIBUTING.md file should probably say this, so it shows up when doing a PR.

Issue #1381, opened 2016-09-29
CSS text select accessibility issue with schema.org site


mrshannonyoung:
On the schema item pages themselves, for example:

[https://schema.org/Person](https://schema.org/Person)

Hovering over a property name will produce the hover state, with a burgundy background and white text (easily readable).

```
a:hover {
  color: #fff;
  background-color: #660000;
}
```

However, if I want to select the item property text to copy/paste, the selection background is grey on a grey background making it very hard to know what amount of text if any I have selected.

I suggest the selected state should mimic the hover style

```
.prop-nam a::selection {
  color: #fff;
  background-color: #660000;
}
```

danbri:
Sounds like a problem - thanks for taking the time to report it. @RichardWallis - could you take a look?

chaals:
There's a problem that selection colours are a user setting. Messing with them is therefore pretty tricky. I suggest we actually flip the colour scheme, so the background is white, since in most cases the user will have defined a selection that works for them over white - being the most common background colour.

danbri:
@RichardWallis where are we with this?

RichardWallis:
Fixed with #1477

PR #1477, opened 2017-01-12
CSS Select set to match CSS Hover

RichardWallis:
Issue #1381 In some browsers the select colours are difficult to read


danbri:
@RichardWallis do we still need this? Could you retarget it at 'master'?

RichardWallis:
Yes merge it into master

PR #1491, opened 2017-01-25
Travis linter

gkellogg:
Adds a Rakefile to make examples, and cooked schema.org vocabulary and context (needs fixing). Validates the vocabulary and each example on Travis-CI.
Fixes #53.

PR #1498, opened 2017-01-26
Travis Linter validations

gkellogg:
This replaces #1491 and is now against sdo-callisto (which is well ahead of master). It merges the Ruby tests with Python, so both run. It fails because of errors found in examples, so it is performing as expected.

/cc @lanthaler @RichardWallis


gkellogg:
Other commits are due to merging previous sdo-makemake version to sdo-callisto. If this is an issue, I could re-create cleanly on top of sdo-callisto.

gkellogg:
Replacing with a cleaner version.

PR #1499, opened 2017-01-26
Linter tests on Travis

gkellogg:
This replaces #1491 / #1498 and is now against sdo-callisto (which is well ahead of master). It merges the Ruby tests with Python, so both run. It fails because of errors found in examples, so it is performing as expected.

Ruby tasks to validate vocabulary and examples on Travis-CI.
Allow rspec --next-failure by creating memory file.
Merge Ruby Travis configuration with Python.

/cc @lanthaler @RichardWallis


lanthaler:
This is a great step forward. Is the plan to fix all tests before merging this?

If possible, it would be great to show a file and a line number to know where to look for examples. It took me a while to figure out where to find `Examples VideoObject-MusicGroup-MusicRecording-Event-video-interactionCount-InteractionCounter-duration-interactionStatistic-interactionType-14-jsonld.html`.

Btw. I'm not sure I understand all of the log output. What does

    Failure/Error: specify(example.split('/').last) {expect(example).to lint_cleanly}

or

    # ./spec/examples_spec.rb:8:in `block (3 levels) in <top (required)>'

mean?

gkellogg:
> This is a great step forward. Is the plan to fix all tests before merging this?

I didn't think so, but to keep the build from failing, I can make all of these appear as tolerated failures when run on Travis. When run locally, that would continue to come up as failures.

> If possible, it would be great to show a file and a line number to know where to look for examples.

Good idea, I can probably take the comment block preceding the test and add it as an XML comment, along with file source and line number  to the example file, then extract that when doing error reporting.

> Btw. I'm not sure I understand all of the log output. What does ..
As each failure gets reported using the source line from the test file the RSpec tool (not ReSpec) shows that context. I'm not sure I can silence it. It comes from [here](https://github.com/gkellogg/schemaorg/blob/travis-linter-2/spec/examples_spec.rb#L7-L9) in the spec/example_spec.rb file. In this case `example` is the path to the file containing the extracted example.

danbri:
This is awesome, though doubtless we have a mountain of work caused by it. Maybe we can mark the backlog expected failures for now, just to get it merged?

gkellogg:
@danbri I have some improvements I'd like to make after thinking about @lanthaler's feedback, which I think will make the Travis results more useful.

Until it can build cleanly, I'd like to keep the failures "soft" when running on Travis. At some point later on, we can change (perhaps by just changing an environment variable) the behavior so that errors cause failures, and then it can better be used to find problems in other PRs.

Dataliberate:
+1 to keeping failures soft at this stage.

Approaching a release, adding a raft of issues related to examples syntax I
believe would be counter productive.

Once happy with the running of this in Travis, we should raise a combined
issue to review all the identified errors and fix them as a batch.
Failures in this area should then be identified as hard.


Richard Wallis
Founder, Data Liberate
http://dataliberate.com
Twitter: @rjw

danbri:
Agreed re soft failures (running locally too). While it flags up work that needs doing, right now it drowns out everything else. 

We did btw begin a process of adding short identifiers to each example. That might help here.

Gregg - could you move spec/ down inside the scripts/ folder? (and any of the other top-level files that can live deeper in the filetree). We're primarily a Python-based project so if we can hide the Ruby it may help avoid confusion...

RichardWallis:
@danbri Yes we did btw begin a process of adding short identifiers to each example in code but it wasn't continued out to rewriting the examples files.

That may make sense to be prerequisite to doing a big fix of examples identified here, followed by a check for future identifierless examples.

gkellogg:
> Gregg - could you move spec/ down inside the scripts/ folder? (and any of the other top-level files that can live deeper in the filetree). We're primarily a Python-based project so if we can hide the Ruby it may help avoid confusion...

Shouldn't be a problem.

gkellogg:
> Agreed re soft failures (running locally too).

I left errors in when running locally; you can always run `LINT_SOFT=true rspec spec/examples_spec.rb` to make them warnings, but the `--next-failure` feature only works on failures, not warnings. Anyway, it's the build that is important.

Travis now passes, with 216 "pending" tests, which would be failures if run locally.

Everything builds into and runs out of the scripts directory too.

I think it's ready to merge.

gkellogg:
One thing that would be nice, but I couldn't figure out, is to create a [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) so that the Ruby tests are run separately from the python tests.

Perhaps the way to do it is with environment variables, as those are included in the matrix. If an environment variable need to be set to run either python or ruby, this would cause them to show up as different builds. For example:

```
language: python
python: "2.7"
env:
  global:
    - SOFT_LINT=true
  matrix:
    - RUBY=true
    - RUBY=false
before_script:
  - SDK_URL=https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.49.zip
  - wget $SDK_URL -nv -O google_appengine.zip
  - unzip -q google_appengine.zip
  - rm google_appengine.zip
  - export PYTHONPATH=$PYTHONPATH:$TRAVIS_BUILD_DIR/google_appengine/
  - rvm install 2.4.0
  - bundle install --gemfile=scripts/Gemfile
script:
  - python scripts/run_tests.py
  - (cd scripts; bundle exec rake)
sudo: false
```

Python scripts could then abort if `RUBY=true` and Ruby scripts could abort if `RUBY=false`, perhaps directly in the `script` lines. Something for someone to play with.

gkellogg:
@danbri @RichardWallis Is there anything else you need addressed before you can pull this?

danbri:
Feeling reckless, let's see what happens. Merging...

danbri:
See https://travis-ci.org/profile/schemaorg and https://travis-ci.org/schemaorg/schemaorg

I see errors, "The command "eval git fetch origin +refs/pull/1502/merge:" failed. Retrying, 2 of 3." ...


Running locally the usual way i.e. ./scripts/run_tests.py  I get a pretty normal

> Ran 85 tests in 76.296s
>
> OK (expected failures=3, unexpected successes=2)

(... we expect the unexpected)

danbri:
The error might be because I merged a recent PR? https://github.com/schemaorg/schemaorg/pull/1502

gkellogg:
Looks to me like everything's running properly on Travis. It's reporting pending failures, which it's intended to. I don't see the merge failure. The history of builds for https://travis-ci.org/schemaorg/schemaorg are all clean. Point me to a specific error you're seeing, and I'll see if I can figure out anything.

danbri:
It seems to have settled down now, and looks good.

Here's a snapshot of a local run too, for archival of the errors: https://gist.github.com/danbri/7c14c96c9c11bc8431c6220ce6d60550

gkellogg:
BTW, there is a class of errors like `ERROR <http://example.org/>: htmlParseEntityRef: no name`, which are fairly opaque (thanks Nokogiri/LibXML!). This is complaining about an unescaped entity such as `Beachwalk Beachwear & Giftware`, which should be `Beachwalk Beachwear &amp; Giftware`.

danbri:
Thanks, I guess re &amp; ... those kinds of errors are an old friend :)

That said, I have been thinking about turning these examples (once fixed!) + more cornercases into something that could be available for parser suite writers to test against. One cornercase that troubles me is that with JSON-LD inside HTML (assuming even HTML-5.x rather than the XHTML family or polyglot) there are *two* layers of special character - and unicode? - escaping to worry about; HTML's and JSON(-LD)'s. Do you have any tests that cover those kinds of issue?

gkellogg:
No, I don't, and the JSON-LD test suite does not test embedded HTML. It's not even clear what the rules are.

I would think that JSON-LD that contained </script> might need to be contained within XML comments, or even CDATA (is that even valid anymore?). Still leaves the potential for the JSON-LD to contain a closing escape/comment, so you're in trouble either way.

While exploring these boundaries is useful, there's also the general advice for "it hurts when I do this" ... don't do that :)

 The HTML folks probably have such tests for other script embedding, which could be adapted.

FWIW, my RDFa parser, which is used for the tests, removes embedded CDATA and XML comments surrounding script contents before passing on to the embedded parser (JSON-LD in this case).

PR #1503, opened 2017-02-01
Linter example fixes

gkellogg:
Numerous fixes to syntax problems in examples. Created some JSON-LD versions of examples for Course examples.


gkellogg:
@danbri @RichardWallis These would seem like uncontroversial changes, as they simply fix obvious syntax issues. The means that remaining example issues are semantic.

danbri:
Fantastic, thanks. Merged! 

Did you notice any systematic syntactic errors that may have propagated widely around the Web? Microdata in particular?

gkellogg:
Most semantic errors are due to the use of obsolete properties, or combinations that are not "legal" given the mixture of classes and properties. The Travis results should be useful (if a bit verbose). For example:

* property  schema:interactionCount
   Term is superseded by schema:interactionStatistic
* property  schema:interactionStatistic
   Subject _:g53241080(schema:MusicGroup) not compatible with domainIncludes (schema:CreativeWork)
 * property  schema:video
    Subject _:g53241080(schema:MusicGroup) not compatible with domainIncludes (schema:CreativeWork)
* property  schema:url
   Object "foo-fighters-rope.html" not compatible with rangeIncludes (schema:URL)
   Object "foo-fighters-everlong.html" not compatible with rangeIncludes (schema:URL)
* property  schema:interactionStatistic
   Subject _:g65449720(schema:MusicGroup) not compatible with domainIncludes (schema:CreativeWork)
* property  schema:video
   Subject _:g65449720(schema:MusicGroup) not compatible with domainIncludes (schema:CreativeWork)

These mostly involve some value judgements about what properties and/or values best show the intent of the example without being inconsistent with the class/property definitions.

Of course, it may be that the reasoner is overly sensitive. For example, it expects that a schema:URL will be an IRI, not a literal. using Postal's law, it's reasonable to get such a value, but perhaps not to use as a canonical example. We may want a way to signal when such use is intentional.

Ideally, we'll get down to a point of zero warnings, so these can be turned back into errors against which incoming PRs are judged so that bad examples don't slip in.

PR #1517, opened 2017-02-09
Add InstructionSet

robdodson:
This is a continuation of the work started in https://github.com/schemaorg/schemaorg/pull/1087

- Recipe now inherits from InstructionSet
- Did my best to inherit subproperties but would love additional input
- Continued to use `yield` instead of `outcome` because `outcome` already [expects a MedicalEntity](http://health-lifesci.schema.org/outcome) and I wasn't sure if we'd want that in a Recipe?

If possible, I'd love for folks to review what's here so far and once we've settled on names and structure I can add in more examples.

@danbri PTAL

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -10275,7 +10275,7 @@ Standards bodies should promote a standard prefix for the identifiers of propert
 <div typeof="rdfs:Class" resource="http://schema.org/Recipe">
   <span class="h" property="rdfs:label">Recipe</span>
   <span property="rdfs:comment">A recipe. For dietary restrictions covered by the recipe, a few common restrictions are enumerated via [[suitableForDiet]]. The [[keywords]] property can also be used to add more detail.</span>
-   <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/CreativeWork">CreativeWork</a></span>
+   <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/InstructionSet">InstructionSet</a></span>
 </div>
 
 <div typeof="rdf:Property" resource="http://schema.org/totalTime">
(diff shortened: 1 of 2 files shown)


danbri:
@vholland any comments?

Main concern for me is that InstructionSet sounds like computer hardware chips. We can always tweak "outcome" if needed, but the definition will be vague.

vholland:
I have similar issues with InstructionSet, but don't have good ideas for an alternative. HowToInstructions seems odd.

I'm a little confused by ```supplies``` and ```instructionSupplies```.

- Do we need both? We should be able to infer whether the supplies apply to the instructions as a whole or an individual step based on the property subject.
- Do those properties need to accept an ItemList? If the supplies themselves don't need to be ordered, authors should be able to repeat the property or use an array as appropriate.

robdodson:
Some other options:

- `HowToGuide`
- `SetOfInstructions`
- `Guide`
- `Tasks`

> I'm a little confused by supplies and instructionSupplies.

I can use `supplies` throughout. I should also update the PR to use `estimatedCost` throughout instead of `estimatedSupplyCost`.

> Do those properties need to accept an ItemList? If the supplies themselves don't need to be ordered, authors should be able to repeat the property or use an array as appropriate.

Being a bit of a schema.org newb I don't know the right answer to this :D happy to defer to whatever you guys think is most appropriate.

dbiollo:
Sorry to jump in on the conversation uninvited, but I've been looking on schema.org for something for how-to pages and this InstructionSet looks like it would fit the bill. Is this pull request still going ahead since it's been in the works for a while? 

I'd toss in "InstructionManual" and second "HowToGuide" as suggestions for the type name.

robdodson:
I haven't had time to work on it. @dbiollo if you're interested in picking it up that would be awesome.

danbri:
I believe this was handled by http://schema.org/HowTo http://schema.org/docs/releases.html#g647 #647 #1616

Issue #1601, opened 2017-04-24
/publishingPrinciples describes organizational principles, should be applicable to Org


danbri:
http://schema.org/publishingPrinciples currently only applies to a CreativeWork e.g. NewsArticle.

Current definition is: "Link to page describing the editorial principles of the organization primarily responsible for the creation of the CreativeWork."

This originated with rNews, see http://dev.iptc.org/rNews-1-The-NewsItem-Class who phrased it as "A link to a natural-language statement of the principles adhered to by the source of the NewsItem.".

Per #1525 I've been looking into capturing more nuance around News, including different kinds of publishing principles, e.g. ethics policies vs diversity policies. It looks straightforward to capture such things via sub-properties of the publishingPrinciples property. However since the emphasis is on organizational practices, it would sometimes be more straightforward if the property could apply directly to an Organization. 

Suggested revised definition:

> "The publishingPrinciples property indicates (typically via URL) a document describing the editorial principles of an organization that relate to its activities as a publisher, e.g. ethics or diversity policies. When applied to a CreativeWork (e.g. NewsArticle) the principles are those of the organization primarily responsible for the creation of the CreativeWork."

This definition allows the property to be applied directly to an Organization. We should also add this in the schemas, as well as (an oversight...) CreativeWork to the expected values of the property.

danbri:
Ping @agebhard @kansandhaus

An open question: what might we say about content created and published by individuals, e.g. bloggers who have ethics/transparency or other statements. Should we add Person also?

agebhard:
@danbri I'll discuss with the others and get back to you here. (Personally, I think this all sounds very reasonable, including the second suggestion to include Person.)

darobin:
FWIW we use it to point to a tree of `Action`s that fully describe the peer-review process that a given `Article` has to go through. So you can see that an article has to be submitted, that there will be reviewers, that the editor will make an assessment, etc. with specific questions to the answered and all that. This may be too scholarly for #1525, just thought I'd share in case you reckon there's something you can reuse.

danbri:
@darobin so if you have an article pointing to a publishingPractice URL, where do you attach the Action? on the article, on the URL, or does publishingPractice point to an Action? i.e. is http://schema.org/potentialAction used?

kansandhaus:
I personally share Andreas' position as well. Extending to Person and Organization makes perfect sense to me.

agebhard:
[...]

Issue #1589, opened 2017-04-13
Add a new type CriticReview as a more specific type of Review


gmackenz:
Many websites (e.g. EazyDiner, Dine Out, Ou Bruncher, etc.) provide markups for both critic reviews (reviews written by professional critics) and user reviews using schema.org/Review. As schema.org/Review does not provide a clear distinction between critic reviews and user reviews at this time. It would be very useful to be able to differentiate which reviews are by critics.

I propose the adding  CriticReview as a more specific of Review.

Example markup:
```json-ld
{  
    "@context":"http://schema.org/",
    "@type":"CriticReview",
    "itemReviewed":{  
        "@type":"Restaurant",
        "image":"http://www.example.com/seafood-restaurant.jpg",
        "name":"Legal Seafood"
    },
    "reviewRating":{  
        "@type":"Rating",
        "ratingValue":"4"
    },
    "name":"A good seafood place.",
     "author":{  
          "@type":"Person",
          "name":"Bob Smith"
     },
    "reviewBody":"The seafood is great.",
    "publisher":{  
        "@type":"Organization",
        "name":"Washington Times"
     }
}
```

danbri:
This seems a worthwhile distinction to articulate, but would you care to also propose a definition for "CriticReview" too? Without a definition I fear we'll just see everyone start using CriticReview because it sounds like "an important review".

The definition ought to make fairly clear who counts as a "critic". You mention "professional", which is terminology we avoided for ClaimReview since you don't need to be paid to be a fact checker to check facts. Similarly, you might be a well regarded film, TV or food critic on the path to professional recognition e.g. via an independent popular blog.  We can say "professional" but it should be softened to allow bloggers etc and not imply that a certain form of employment or business model is necessarily required.

Perhaps something like: "A 'CriticReview' is a review written or published by a source that is specifically known for its reviewing activities. These can include newspaper columns, travel and food guides, TV and radio shows, blogs and other independent Web sites. CriticReviews are typically more in-depth and professionally written. For simpler user/visitor/viewer/customer reviews, it is more appropriate to use the Review supertype.

[...]

Issue #1525, opened 2017-02-16
Support finer-grained distinctions around NewsArticle [and news content]


danbri:
There is interest (e.g. from [The Trust Project](http://thetrustproject.org/), and conversations around fact checking and our ClaimReview markup #1061) in using schema.org to describe more detailed kinds of NewsArticle, such as allowing sites to express that an article is an opinion piece, a backgrounder document, analysis, visualization, or that it is sponsored content.  We have also touched on these themes here recently when discussing the idea of a 'Satirical News Article' subtype, see #1437.

See also https://www.theatlantic.com/technology/archive/2017/05/what-people-really-want-from-news-organizations/526902/

In the simplest case we could overload the "genre" property here, but I'd suggest (repeating myself from #1437) that a bit more structure will be more rewarding in terms of markup usability / quality. We could use a simple Enumeration list but I lean towards types since they can be instantiated (to carry properties) and even subtyped further. More compellingly, in the (increasingly popular) JSON-LD format, markup for types is simple, and even multiple types also a simple array of short tokens, whereas in general our markup for enumerated property values involves full URLs and is more error prone.

Update (May 10th) from @danbri:

Several results of these investigations:

 * Proposed new definition of NewsArticle to replace "A news article".
 * Added several (pending) subproperties of http://schema.org/publishingPrinciples
 * Tweak to publishingPrinciples - see #1601 - to allow organizations/persons directly.
 * New subtypes for NewsArticle, again pending: AnalysisNewsArticle, BackgroundNewsArticle, ReportageNewsArticle, OpinionNewsArticle, ReviewNewsArticle.
 * New Article types alongside NewsArticle: SponsoredContentArticle, SatiricalArticle.
 * A "News overview" document describing these, http://webschemas.org/docs/news.html

thadguidry:
So we covered a bit of this before with rNews and how they basically use Genre property for this currently.
But then I think again about some benefits/cons you mention and then think about @rvguha proposal of Compositional Terms / Types #1493 and feel much better about expanding types when/where there would be wide agreement.  I think in the news domain, this would get wide agreement.  But in the end....its still just taking Article and slapping a genre on its back.

So what we are REALLY trying to do here, is basically making it easier for any heavy markup publishers... like news organizations.  And saying that "its OK and beneficial to break norms and add some useful types for this heavy markup domain instead of properties"

So... +1  for expanding typing for News domain.

danbri:
[...]

Issue #1437, opened 2016-11-18
Add markup for Satirical Articles (or as a subtype of NewsArticle, SatiricalNewsArticle)


danbri:
Back in 2012, I proposed a "[vocab idea: SatiricalArticle](https://lists.w3.org/Archives/Public/public-vocabs/2012Apr/0076.html)" for schema.org.

As Jason Douglas put it at the time,

> so essentially you're looking for the schema.org equivalent to  **;-)** ?

At the time of the original 2012 proposal we had fairly detailed discussions but eventually did not progress the idea, in large part because it was unclear what incentive satirical publishers might have for including the markup. 

In 2016 the environment is somewhat different, and satirical news publishers may have some incentive to distinguish themselves from sites that falsely claim their innacurate stories to be true.

There were some useful suggestions that included official codes from IPTC (for news) and from EBU for TV/media, based on the idea that we can model these as a particular well-known [genre](http://schema.org/genre).  While expressing a mapping to those codes would be useful I suggest we optimize for publisher simplicity here, and address the basic case of satirical news articles through a subtype of NewsArticle called SatiricalNewsArticle.

This would complement the recent success of schema.org's fact checking markup, which adds a subtype of Review called [ClaimReview](http://pending.webschemas.org/ClaimReview) (and lists the specific claimReviewed), and would be highly topical given recent widespread attention to the problem of fake news. /cc #1061

rvguha:
This is a great idea. Lets do it!

danbri:
Aye aye cap'n

thadguidry:
@danbri NO NO NO !  Lets use the defacto global standard of Star Wars !  (not Star Trek)

danbri:
We're more in the [Pirates of the Caribbean](http://www.imdb.com/title/tt0325980/quotes) tradition around here

> And thirdly, the Code is more what you'd call “guidelines” than actual rules.

bquinn:
> If you were waiting for the opportune moment, that was it!

rvguha:
I would prefer Star Trek.

https://www.quora.com/Which-is-better-and-why-Star-Wars-or-Star-Trek

mediaprophet:
Kinda enjoying Potter references atm...

What about fiction, nonfiction + subtype: genre + Style?

I was thinking about genre: news, style: Article being a content genre /
style?

Others could be Style: report Style: liveCoverage

Etc.

mediaprophet:
[...]

PR #1604, opened 2017-04-26
Merge NewsArticle, CriticReview, satirical etc proposals into webschemas for review

danbri:
Addresses #1525 + #1589 + #1601 + #1437


danbri:
pushing to webschemas.org. Note: still need to update docs/releases.html with an overview of the proposed changes.

PR #1606, opened 2017-04-27
Makes Painting, Photograph and Sculpture subclasses of VisualArtwork

### LINKED ISSUE #1599: Painting, Photograph and Sculpture subclasses of VisualArtwork
Painting, Photograph and Sculpture should maybe be subclasses of VisualArtwork

Tpt:
Closes #1599

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -1499,7 +1499,7 @@ A more detailed overview of [schema.org News markup](/docs/news.html) is also av
     <div typeof="rdfs:Class" resource="http://schema.org/Photograph">
       <span class="h" property="rdfs:label">Photograph</span>
       <span property="rdfs:comment">A photograph.</span>
-       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/CreativeWork">CreativeWork</a></span>
+       <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/VisualArtwork">VisualArtwork</a></span>
     </div>
     <div typeof="rdfs:Class" resource="http://schema.org/Notary">
       <span class="h" property="rdfs:label">Notary</span>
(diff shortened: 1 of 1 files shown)


pcoloma:
I have no problem describing visual work with schema, it gets complicated for pages like this that lists for example different works and techniques :
- photographs,
- paintings,
- prints,
- drawings,
- and ceramic work
http://www.de-santa-coloma.com/gracias-a-la-vida


The other thing I need guidelines is with this :
As an artist I take part to art exhibitions that can be either SOLO or GROUPED events,
In grouped events I need to be able to list the other artists and techniques  (possibly their fields of work (author, painter).
There are no tickets for sale (as the events field "offers" implies), and most of the time the prices are not given on the web.
The other point is that I'm not hosting the shows. Places can be art galleries, or "ateliers", or art fairs.
Please advice thanks
Patrice

RichardWallis:
Potentially two new types ([Drawing](https://schema.org/Drawing) & [Poster](https://schema.org/Poster)) that were added to pending in v3.5, could also benefit from becoming subclasses of VisualArtwork.

I suggest we review adding them, or not, to this proposal before merging.

ostephens:
I guess the question is whether any of these things can be other than VisualArtwork - which probably boils down to the question - are they always, by definition, a "work of art".

To be honest, given that assigning a more specific type brings no extra qualities, I feel that making any of these subclasses is the wrong thing to do. I'd be more inclined to specify the form through VisualArtwork->artform using the scd url for the relevant form.

Bringing more modelling into this area without a review of how we describe artworks overall feels like the wrong thing to do

RichardWallis:
I have sympathy with @ostephens's thoughts on this.

The question should be, is a (Painting or Photograph or Sculpture or Drawing or Poster) the [significant] majority of the time also a [visual] work of art?

My personal answer being - _I'm not sure_.

On an individual item by item basis, the answer is in the eye and or mind of the beholder. In practice the beholder being the person creating, or creating the algorithms that produce, the metadata  describing the item.

By accepting this proposal  we would be implying that items of these types should [always] be considered as artwork.

Fortunately with Schema.org and its [MTE capabilities](https://github.com/schemaorg/schemaorg/wiki/How-to-use-Multi-Typed-Entities-or-MTEs) we have the option to decide at the time of metadata creation and/or transcription.  

Definition in the form ```"@type": ["Photograph", "VisualArtWork"],``` is both acceptable and in some ways slightly more informative.

Just my 2¢

Tpt:
That's a very good point indeed. I am going to close this PR and plan to make an other one that adds an example of Painting + VisualArtwork to give an example of MTE usage for this kind of entities.

danbri:
I agree with the conclusion here. 

Also at least sculpture is not intrinsically visual, e.g. "sculpture for the blind" 

http://www.lenkaclayton.com/sculpture-for-the-blind-by-the-blind etc.

PR #1628, opened 2017-05-19
Exhibition type for art events

pcoloma:
As an artist I take part to art exhibitions that can be either SOLO or GROUPED events,
In grouped events I need to be able to list the other artists and techniques  (possibly their fields of work (author, painter).
There are no tickets for sale (as the events field "offers" implies), and most of the time the prices are not given on the web.
The other point is that I'm not hosting the shows. Places can be art galleries, or "ateliers", or art fairs.
Thx for advice


danbri:
@pcoloma I do not think we have any more advice for you at this time

Also something odd is happening in Github with other unrelated commits always showing up associated with this thread. I will close thw Issue. We can still talk here about Exhibition markup if needed.

PR #1635, opened 2017-05-23
alias value to @value

darobin:
This sort of goes with #1634 but should probably be considered independently (hence the separate PRs).

The JSON-LD context aliases `id` and `type` to their @ variants, but not `value` to `@value`. If one adopts these aliases (which are nice) this leads to strange cases for parsed literals in which you get `{ "id": "rdf:HTML", "@value": "<p>…</p>" }`.

Optics, ergonomics, etc. matter, so this little bit would surely help clean that up and make it look neat to newcomers.

### DIFF
--- a/api.py
+++ b/api.py
@@ -999,6 +999,7 @@ def GetJsonLdContext(layers='core'):
         jsonldcontext = "{\n  \"@context\": {\n"
         jsonldcontext += "        \"type\": \"@type\",\n"
         jsonldcontext += "        \"id\": \"@id\",\n"
+        jsonldcontext += "        \"value\": \"@value\",\n"
         jsonldcontext += "        \"@vocab\": \"http://schema.org/\",\n"
         jsonldcontext += namespaces
 


danbri:
Unlike 'id' and 'type', we have a property called 'value': http://schema.org/value

Perhaps it can be generalized?

darobin:
Duh, I thought of that and then somehow forgot. ☕️

That said, I wonder if the two could usefully overlap without destroying the underlying RDF model. Maybe @msporny would have an idea?

gkellogg:
@darobin In JSON-LD 1.0, there is no way for a term to be interpreted in two different ways: it either expands to `http://schema.org/value` or `@value`, not both or either.

Note that the use of `@value` in schema.org markup is likely not that wide-spread, as the schema.org context does not use `@type: @id` for many terms, so values will generally be interpreted as literals anyway. The issue may come up when an author wants to associate a datatype with the value other than the default, but this is advanced markup, so the use of `@value` probably isn't a real issue, and it can be avoided by using a local context.

The work in JSON-LD 1.1 _does_ allow terms to take on different values; for example, for the property `http://schema.org/name`, a terms-specific context could be defined that expands `value` to `@value`, but I wouldn't expect the schema.org to make general use of this ability, presuming that the 1.1 version were ever adopted.

I'd suggest just forgetting this issue and not trying to create a more "friendly" term for `@value`.

msporny:
> ```{ "id": "rdf:HTML", "@value": "<p>…</p>" }```

Yes, that's horrible and most Web devs, including yours truly, will run away screaming upon the sight of such a monstrosity. If you can't alias @value, which we can't w/o backwards-compat issues in schema.org, the next best thing would be something like this:

You could use JSON-LD's type coercion mechanism, but in a way that is hidden for developers. So, schema.org could define a new term for HTML content items:

```javascript
"contentHtml": {"@id": "schema:contentHtml": "rdf:HTML", "@type": "rdf:HTML"}
```

and that would give you the clean expression that most web devs would like to see:

```javascript
  "contentHtml": "<p>Hello&nbsp;world!</p>"
```

danbri:
Ok, where are we with this one?

@gkellogg wrote
> I'd suggest just forgetting this issue and not trying to create a more "friendly" term for @value.

What would the friendly term be? Can we have an example of using it?

danbri:
I'm going to close this. We don't have an associated Issue explaining what problem is solved here. It seems to touch on several, especially usability and unambiguity of content that contains markup. But a pull request isn't the right place to hash out the problem statement.

PR #1646, opened 2017-06-05
HTTPS for links to Schema.org

johnnypeck:
HTTPS for links to Schema.org, small grammar fixes and markup for links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,29 +2,28 @@ Welcome to Schema.org
 =====================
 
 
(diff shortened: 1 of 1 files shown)


Download:
Seems like a useful change!  👍

danbri:
We haven't fully flipped to https (there are some subtleties since 'http' appears in a *lot* of data), but I think it's time to speed up that migration, will merge this, thanks...

Issue #1657, opened 2017-06-14
schema:Person colleagues can be anything?


teledyn:
I'm proposing removing line 4200 in the current schema file: as it sits, while other people associated with schema:Person (children, parents, knows) have an rdfs:range of schema:Person, only the colleague is allowed to be either schema:Person or any URI. That seems odd.  I mean, hey, I love my phone, but I wouldn't (yet) consider Google Assistant to be a colleague ;)

Was there some reason why colleague was defined differently?

nicolastorzec:
Agree. Changes committed. See PR 1665.

PR #1665, opened 2017-06-15
Changed range of colleague from "Person or URL" to "Person".

nicolastorzec:
See [issue 1657](https://github.com/schemaorg/schemaorg/issues/1657).

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -4197,7 +4197,6 @@ The place is __open__ if the [[opens]] property is specified, and __closed__ oth
       <span property="rdfs:comment">A colleague of the person.</span>
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Person">Person</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/Person">Person</a></span>
-      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/colleagues">
       <span class="h" property="rdfs:label">colleagues</span>


danbri:
Actually this is a little more complicated than it looks - we should talk through the consequences. 

I totally agree that the inconsistency with other similar properties of Person is not good.

There are two levels of complication. The first is that the way we've used "expected type: URL" in schema.org is slightly magical, in that it doesn't refer to a kind of thing that can be a "colleague" but a manner of describing colleagues that uses hypertext links instead of inline descriptions.

The second is that JSON-LD required us to go through all our properties and indicate which ones default to URL values, and which to textual values. This was difficult since in many cases we allowed both. The idiom used for generating our JSON-LD context file was based on whether a property "expects" URL values. This is because the biggest consequence of the JSON-LD distinction is on the handling of values. For example, this controls whether "colleague: "../people/person_123" is parsed into a simple string or into an absolute URI.

We may want to model this separately from "expected type: URL", e.g. by adding new schema annotating constructs into meta.schema.org. Regardless, URL will still be somehow special (especially as we blur URLs for things vs pages that have those things as main entities).

I'd prefer not to merge this change for v3.3 as it will change the behaviour of all JSON-LD parsers that read our context file, but I would very much like to make progress on clarifying these issues and on making these Person-properties more consistent.

nicolastorzec:
Got it. I actually suspected something like this. We have similar design conflicts in the ontology that backs the Yahoo Knowledge Graph.

danbri:
@nicolastorzec - it's a variation of the tensions that always existed between XML schemas vs RDF schemas. One talks about data structures, the other about the things those structures describe. RDF has always been strong on the latter, weak on the former. It may now be interesting to try applying https://www.w3.org/TR/shacl/ to schema.org, although I think it still takes a per-document view and doesn't directly address the issue of descriptions spread across multiple linked documents.

PR #1667, opened 2017-06-16
Fix typo - acccommodation spelling

RehanSaeed:


RichardWallis:
Thanks for the patch.  Unfortunately it will not work as these csv files are created from the main definition files at release time so would be overwritten.

The source needs fixing for http://schema.org/numberOfRooms.

@danbri I believe you are in master right now - do you want to fix this simple typo for numberOfRooms in schema.rdfa for the next release.

RehanSaeed:
@RichardWallis Where is the source definition file?

RichardWallis:
@RhanSaeed It's in data/schema.rdfa 

~Richard

danbri:
Fixed in schema.rdfa, we'll regenerate the .csv at some point and it will be fixed automatically so I shall close this Pull Request. Thanks @RehanSaeed :)

Issue #1640, opened 2017-05-23
Broaden the domain of schema:issn


tfrancart:
Currently schema:issn is declared with a schema:domainIncludes on type Periodical only. While this is not strictly wrong according to the semantic of "domainIncludes", this is restrictive and might be misleading according to the possible kind of things on which an ISSN can be assigned; in particular **an ISSN can be assigned to any kind of series, but also to datasets or websites**. This is documented in the official ISSN documentation, the ISSN Manual, section 0.2 and 0.3 (see http://www.issn.org/understanding-the-issn/assignment-rules/issn-manual/#0-2-scope-of-issn and the PDF version at http://www.issn.org/wp-content/uploads/2013/09/ISSNManual_ENG2015_23-01-2015.pdf)

We thus suggest to broaden the domain of schema:issn to the following types :

- CreativeWorkSeries (which includes Periodicals)
- Blog
- Dataset
- Website

This suggestion has been discussed and approved by the ISSN International Center (http://www.issn.org/).

nicolastorzec:
+1

@danbri @RichardWallis @vholland @chaals ?

RichardWallis:
Makes sense. +1

dbs:
Generally +1

I'm not keen about the revised sentence in the description "You can repeat this property to identify its different medium versions." though; the direct antecedent of "its" is "this property" instead of the intended "serial publication" from the previous sentence, which will be confusing, and "different medium versions", while used in the ISSN documentation, will take more effort for the target audience to scan and understand than the original "different formats".

Perhaps "You can repeat this property to identify different formats of this serial publication." to maintain the structure of the original and repeating "serial publication" to tie the second sentence of the description back to the first?

nicolastorzec:
Thanks for the feedback Dan.

I don't have a strong opinion either way. I will change later this week to provide more time for feedback.

tfrancart:
FYI, the ISSN numbers for the different formats of a continuing resource (e.g. printed vs. electronic) are aggregated into a single ISSN-L ("Linking ISSN") : http://www.issn.org/understanding-the-issn/assignment-rules/issn-manual/#3-1-designation-of-issn-l. The ISSN-L is a unique identifier for the continuing resource, independant from the medium/format.

RichardWallis:
I support @dbs's suggested tweaking of the description.

dbs:
[...]

PR #1669, opened 2017-06-16
Broadened the domain of issn and updated its definition.

nicolastorzec:
Proposed fix for #1640

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -8720,8 +8720,11 @@ postponing for 1.6.
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/issn">
       <span class="h" property="rdfs:label">issn</span>
-      <span property="rdfs:comment">The International Standard Serial Number (ISSN) that identifies this periodical. You can repeat this property to (for example) identify different formats of this periodical.</span>
-      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Periodical">Periodical</a></span>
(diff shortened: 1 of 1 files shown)


nicolastorzec:
Updated description of _issn_ per #1640

danbri:
LGTM. Do you know if these are actually used much with datasets in practice?

danbri:
Minor bug: we added domainIncludes of "Website" instead of "WebSite".

Odd, the runs-in-cloud test integration should have caught this. Maybe I missed it? 

I caught it via commandline test runner. Fixing.

nicolastorzec:
Re: ISSN and datasets, dataset publishers don't use ISSN in my experience. However, research datasets often come with a DOI, e.g. via DataCite.

(Re: domain validation, I didn't see anything suspicious at the time either)

ostephens:
It's worth noting a difference between the use case for DOIs and ISSNs - the former is meant to permanently identify a static resource, while the latter is specifically for a resource where updates or additions are issued periodically. An ISSN is also specific to the publication in/on a specific medium.

My experience reflects what @nicolastorzec says - ISSNs not usually used for datasets but it may depend on your definition of 'dataset'! ISSNs *are* used for databases (see http://www.issn.org/understanding-the-issn/what-is-an-issn/) with updating content - for example the the AgroForsetTree database (http://www.worldagroforestry.org/output/agroforestree-database) has an ISSN - but only for it's CD-ROM incarnation (2079-1399)

PR #1676, opened 2017-06-26
Price - currency issues

CEWebDesign:
Price/currency issues

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -696,7 +696,7 @@ MICRODATA:
   </div>
 
   <div itemprop="offers" itemscope itemtype="http://schema.org/AggregateOffer">
-    Priced from: <span itemprop="lowPrice">$35</span>
+    Priced from: <span itemprop="priceCurrency" content="USD">$</span><span itemprop="lowPrice">35</span>
     <span itemprop="offerCount">1938</span> tickets left
   </div>
 </div>
@@ -724,7 +724,7 @@ RDFA:
   </div>
 
   <div property="offers" typeof="AggregateOffer">
-    Priced from: <span property="lowPrice">$35</span>
+    Priced from: <span itemprop="priceCurrency" content="USD">$</span><span itemprop="lowPrice">35</span>
     <span property="offerCount">1938</span> tickets left
   </div>
 </div>
@@ -747,7 +747,8 @@ JSON:
   },
   "offers": {
     "@type": "AggregateOffer",
-    "lowPrice": "$35",
+    "lowPrice": "35",
+    "priceCurrency": "USD",
     "offerCount": "1938"
   },
   "startDate": "2016-04-21T20:00",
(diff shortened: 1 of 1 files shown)


danbri:
Thanks for the suggestion. It seems that most (all?) of these are about bringing our syntax into an XML-friendly flavour of HTML. Much as I am fond of XML it seems the world has moved the other way and people aren't self-closing their tags in the HTML5 world. I don't think we are going to re-ignite those debates by trying to enforce XHTML compatibility here (see also "itemscope" in Microdata, e.g. latest WD for W3C's version, https://w3c.github.io/microdata/ ). Are there cases here that go beyond a preference for XHTML-style tag structures? /cc @unor

CEWebDesign:
Thanks for looking at this. Just tried to edit inconsistencies stated as invalid in #1678 by @unor . So nothing beyond that as far as I can remember.

unor:
The slash in the start tag of void elements (`img`, `meta` etc.) is

* optional in the HTML syntax ([where it has no effect](https://www.w3.org/TR/2017/REC-html52-20171214/syntax.html#start-tags)), and
* required in the XHTML syntax.

So it could be argued that having the slash is preferable, because it’s valid in both syntaxes, and the only downside for the HTML syntax is having one/two more characters per void element.

I think both ways are fine. Nothing wrong with showing only slashless HTML-syntax examples, nothing wrong with showing examples that work in both syntaxes.

----

@CEWebDesign I didn’t intend to imply in #1678 that `meta` elements *must* have a slash. I just stated that the two example `meta` elements are valid like that :)

CEWebDesign:
Thank you, so nothing wrong with showing both slash and slash-less examples either?
Maybe a hint will be helpful, because of the importance of void elements for schema.org? 
If a hint is given, just slash-less examples are fine, because it is not schema.org to support XHTML. But with slash to be valid in both syntaxes or having both for better diversity are good arguments too. I would prefer having the slash because it is valid for most users when they copy and paste examples.

Hint suggestions:
a) "Examples are both in HTML and XHTML syntax. End slashes in the start tag of void elements (img, meta, link etc.) are optional in HTML syntax and required in XHTML syntax."

b) "Examples are with end slashes in the start tag of void elements (img, meta etc.) to validate in HTML and XHTML syntax. End slashes in the start tag of void elements (img, meta, link etc.) are optional in HTML syntax and required in XHTML syntax."

c) "Examples are without end slashes in the start tag of void elements (img, meta etc.). End slashes in the start tag of void elements (img, meta, link etc.) are optional in HTML syntax and required in XHTML syntax."

Or just 
d) "End slashes in the start tag of void elements (img, meta, link etc.) are optional in HTML syntax and required in XHTML syntax."

@unor Yes, there are other fixes. Thank you for reminder. Maybe there should be an extra commit for these fixes?

PR #1722, opened 2017-08-21
CI: check the validity of the RDFa file

dunglas:
To prevent future errors those fixed in #1721.


dunglas:
(The build will be green when #1721 will be merged.)

dunglas:
@danbri, do you think you can merge #1721 and this one?

danbri:
What is vnu, what does "java -jar dist/vnu.jar data/schema.rdfa" do, and what license is the code?

danbri:
Ok, I see it's the world famous MIT-licensed HTML validator. Sure, though you might want to be consuming one of the distributed formats from http://schema.org/docs/developers.html instead of the HTML/RDFa/RDFS thing we use within the project.

dunglas:
Thanks for the merge @danbri!

PR #1729, opened 2017-08-29
Adds CreativeWork as superclass of HowToDirection and HowToTip.

earljwagner:
Adds ListItem as superclass to HowToStep 
Renames property to step, allowing multiple steps to be specified.

Also updates example:
Sets names for example HowToItems, removes incorrect ids.
Removes extraneous text for HowToItems.


earljwagner:
Replaces #1707

danbri:
I'll amend this slightly after merging to put in a 'supersededBy' from the earlier spelling, "steps".

danbri:
A test failed, "INFO:test_graphs:Term 'http://schema.org/recipeInstructions' has nonexistent super-property: 'http://schema.org/steps'", hmm we should've seen that here. Fixing.

PR #1744, opened 2017-09-20
PR for Issue 1743

tfrancart:
See #1743 
  - deleted inverse properties
  - Changed definition of "Legislation" to make it clear this is a document
  - Fixed Range on legislationType (CategoryCode)
  - Added legislationJurisdiction as a subproperty of spatialCoverage
  - Fixed Definitive into DefinitiveLegalValue to align with other values
  - expanded the definition of legislationIdentifier to clarify that it can apply to string or URIs
  - fixed and adjusted the examples accordingly (But I still don't know how to express inverse properties in microdata)


danbri:
Could you file a separate bug to get some decent inverse property examples in Microdata? /cc @chaals

PR #1761, opened 2017-10-02
Changed mis-abbreviation of Friday as "Fri" to "Fr"

ruthtillman:
Per the guidelines for openingHours, Friday should be abbreviated as
"Fr." In one set of examples, it was abbreviated as "Fri." This error
was corrected in each instance of the example.

Additionally, in a paragraph associated with the same example, two
sentences has no space after a period. Space was added in each instance
of the example.

(My editor appears to have stripped out some additional spaces.)


danbri:
Oh dear, how did we go years without noting that? Ahem. Thank you very much!

PR #1764, opened 2017-10-06
DC Terms for spatialCoverage using spatial not coverage?

thadguidry:
Maybe it was a previous typo by someone but I think http://dublincore.org/2012/06/14/dcterms#coverage has the intent we were after ?
The main difference being between rdfs:range dcterms:LocationPeriodOrJurisdiction and just rdfs:range dcterms:Location


RichardWallis:
Certainly looks like it was a typo.

A disadvantage of not making such encoded relationships such as owl:equivalentProperty visible in the UI

danbri:
Mapping from schema:contentLocation to dct:coverage with owl:equivalentProperty seems too strong.

Am I misunderstanding? http://schema.org/contentLocation is entirely about place/location/spatial, I don't see a temporal aspect.

thadguidry:
@danbri Your correct, it is not equivalentProperty at all to `coverage`...  dc: "The spatial or temporal topic of the resource" does indeed state `or` and we currently applied the correct mapping against only the `spatial` relationship (that without a `temporal` quality).

Thus, I withdraw my PR, as I didn't read closely enough at the time. Current mapping against `spatial` is correct.

Issue #1447, opened 2016-11-29
Expand domain of video to Thing?


RichardWallis:
Following on from an issue (#1446) regarding some errors in an example, it seems the intention of the creator of the example was to attach a [video](http://schema.org/video) property to a MusicGroup type, to link to a video of the group in question.

As the domainIncludes of _video_ is only _CreativeWork_ this is not possible. 

We could propose extending the domain to include _MusicGroup_ but that would only address a simple symptom of a potentially broader problem.

In this age of Snapchat, YouTube, etc., etc., there is a blurring of boundaries between what is an image or a video (moving) image.  How do we classify an IPhone Live Photo for example?

In the current Schema.org structure any _Thing_ can have an _image_.  Is it time to say that say any _Thing_ can also have a video image?

This could be addressed either by changing the domain of _video_ from _CreativeWork_ to _Thing_, or adding _VideoObject_ to the range of _image_.

Discuss...

jvandriel:
This overlaps with issue #292 (something I'd really like to see resolved). The original request/discussion was about expanding the domain of ```video``` to ```schema.org/Thing```.

danbri:
On the blurred lines front, we could also decide 'video' is a special case (subproperty) of 'image', i.e. image(s) that move.

jvandriel:
Sorry to be nitpicky but I don't see a video as images that move. At best it's an collection of images that should be watched in a specific order (at a certain speed).

Now I understand the reluctance to keep adding new properties to ```schemna.org/Thing``` but given the fact video plays such an important role on the web nowadays and that you can find videos about any type of thing, wouldn't it just be easiest to add ```video``` to ```schema.org/Thing```.

After all, it would actually resolve real world issues webmasters have.

RichardWallis:
I'm just revisiting this (in respect of Tourism) and believe @jvandriel's suggestion of adding video to Thing makes sense

thadguidry:
@RichardWallis just one bit of a worry... it might be used as a mentioned....rather than a sameAs...or both.  We need to be clear about what we want in the description.

vholland:
+1 to adding to Thing.

felipesanti:
Hi, from a Tourism industry perspective I support the idea of adding video content to anything that can be of interest to potential tourists.

[...]

Issue #1670, opened 2017-06-19
Add "subjectOf" as a reverse for "about"


vholland:
tl;dr: I propose adding http://schema.org/subjectOf as a named reverse property of http://schema.org/about. The domain would be http://schema.org/Thing and the range would be http://schema.org/CreativeWork and http://schema.org/Event.

There are lots of cases where the main subject of given page is a Person, Organization, etc and authors want to mention there is a work about that entity. It would be nice to have a "subjectOf" property to link to the work. (In JSON-LD, you can use @reverse, but there is no reverse mechanism in microdata yet.)

jvandriel:
Well, there is a proposal for adding @itemprop-reverse to microdata. If that'd exist would you still want to add ```sujectOf```?

Maybe @danbri or @chaals can give us an update about the @itemprop-reverse proposal?

vholland:
Honestly, yes. Most authors I know get confused using @reverse in JSON-LD and resort to naming the reverse in the @context, so why make them go through the extra work.

jvandriel:
> _"I know get confused using @reverse in JSON-LD"_

I hear ya (really) and 'm not saying we shouldn't introduce _subjectOf_ but the difficulty I have with the above statement is that IMHO it's an excuse many use for not having to make the effort of explaining how @reverse works.

Most I introduce it to are confused at first due to being completely unaware of the existence of a reverse mechanism, though I haven't encountered anybody who wasn't able to grasp it - after some explanation. Maybe a bit more guidance and education could do the trick already (through some schema.org notes, or even by having the sponsors state something about it in their webmaster guidelines).

danbri:
we should do both, imho.

akuckartz:
If `subjectOf` is added then please include a comment to that and `about` in the reverse mechanism.

vholland:
Added to pending.

HughP:
Am I correct to notice that there subjectOf is a superset of http://schema.org/WriteAction? I am just noticing that there appears to be nohttp://schema.org/SpeakAction or http://schema.org/RecordAction for the audio modality corollary to the written modality described in http://schema.org/WriteAction. I'm not sure where to make that remark.

RubyMoonProductionCo:
I personally think the domain should be Http:/19/GIG-thing.org

RubyMoonProductionCo:
please and thank you  :)  if possible or at least let me contribute since I have very personal intimate connection to the Iot device.

[...]

Issue #292, opened 2015-01-27
Change domain of associatedMedia to Thing


jvandriel:
Proposal can be found at: https://www.w3.org/wiki/index.php?title=WebSchemas/SchemaDotOrgProposalsArchive#2011-2014_Proposals_for_Schema.org

Discussion at:
http://lists.w3.org/Archives/Public/public-vocabs/2014May/0224.html

danbri:
From the proposal, https://www.w3.org/wiki/WebSchemas/associatedMediaToThing

    Currently, http://schema.org/associatedMedia, which has a range of http://schema.org/MediaObject and which would otherwise seem suitable for this purpose, has a domain of CreativeWork. 
    This means that http://schema.org/MedicalProcedure and other children of Thing 
    that do not descend from CreativeWork have no straightforward means of 
    including media objects that complement their content.

    See the discussion at http://lists.w3.org/Archives/Public/public-vocabs/2014May/0204.html and http://lists.w3.org/Archives/Public/public-vocabs/2014May/0224.html

    Proposed solution
    Change the domain of associatedMedia to Thing. The benefit of this proposal is that we would 
    cover AudioObject, DataDownload, ImageObject, MusicVideoObject, VideoObject, and 
    any other MediaObject subclass that comes into being.

All of the examples listed in the proposed solution,

* http://schema.org/AudioObject
* http://schema.org/DataDownload
* http://schema.org/ImageObject
* http://schema.org/MusicVideoObject
* http://schema.org/VideoObject

... are already subtypes of CreativeWork they can already be used with http://schema.org/associatedMedia without problems.

The only other type offered as motivation for this was MedicalProcedure.

We would prefer to add specific additional domains to a property rather than just open it up for random use in any situation. Schema.org does not have so many types directly under Thing. So let's take this case-by-case.

Let's consider one: BroadcastService. Currently http://schema.org/BroadcastService is a direct subtype of Thing (and this should probably be reconsidered). An example of this would be the UK's broadcasting service BBC1. According to https://en.wikipedia.org/wiki/BBC#Finances "The BBC has the second largest budget of any UK-based broadcaster with an operating expenditure of £4.896 billion in 2012/13". 
"In December 2012, the BBC completed a digitisation exercise, scanning the listings of all BBC programmes ... They identified around five million programmes, involving 8.5 million actors, presenters, 
[...]

PR #1774, opened 2017-10-21
Expand domain of video & audio to Thing

RichardWallis:
As per Issue #1447  
Expand domain of _video_ property to _Thing_ plus update description to clarify use and preserve its continued suitability for current usage on _CreativeWork_.

> A video of, or video component of, the item. This can be a URL or a fully described VideoObject.

Only expanding the domain of _video_ creates an anomaly where any _Thing_ can have an _image_ or a _video_, but only CreativeWorks can also have an _audio_.  To remove the anomaly the domain and description of _audio_ is also expanded.

> An audio recording of, or audio component of, the item. This can be a URL or a fully described AudioObject.

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -3791,7 +3791,7 @@ The place is __open__ if the [[opens]] property is specified, and __closed__ oth
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/audio">
       <span class="h" property="rdfs:label">audio</span>
-      <span property="rdfs:comment">An audio recording of, or audio component of, the item. This can be a URL or a fully described [[AudioObject]]</span>
+      <span property="rdfs:comment">An audio recording of, or audio component of, the item. This can be a URL or a fully described [[AudioObject]].</span>
       <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Thing">Thing</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/AudioObject">AudioObject</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -3791,9 +3791,10 @@ The place is __open__ if the [[opens]] property is specified, and __closed__ oth
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/audio">
       <span class="h" property="rdfs:label">audio</span>
-      <span property="rdfs:comment">An embedded audio object.</span>
-      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/CreativeWork">CreativeWork</a></span>
+      <span property="rdfs:comment">An audio recording of, or audio component of, the item. This can be a URL or a fully described [[AudioObject]]</span>
+      <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/Thing">Thing</a></span>
       <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/AudioObject">AudioObject</a></span>
+      <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
     </div>
     <div typeof="rdf:Property" resource="http://schema.org/author">
       <span class="h" property="rdfs:label">author</span>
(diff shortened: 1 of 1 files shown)


danbri:
Can we just attach them to CreativeWork, plus any other cases that emerge?

Cluttering Thing has a big effect across the site... /cc @rvguha

RichardWallis:
As neither of the two cases referenced in the original issue (#1447) were CreativeWorks (a Tourist attraction and a MusicGroup) and the related issue (#292) references thing such as Events and Products; I think the 'adding them to other cases as they arise' approach will not be sufficient - in the current environment folks want to have videos of everything in the same way we get images of most any type of thing.

I understand the cluttering effect of adding things to Thing, but I think this is one of those cases.

nickevansuk:
+1 here, `video` is native in HTML5, it should be native on `Thing` here too

philbarker:
This seems to change the semantics of the property video from "here is a video which is embedded in this CreativeWork" to "here is a video of this Thing". 

I can see the arguments for having the latter, but it seems like a new property. Rather than clutter up Thing and all sub classes, how about a new property of perhaps of CreativeWork, `depicts` expected range: Thing,  definition: the thing shown or portrayed.

nickevansuk:
Noting that `image` is currently used as "here is an image which is embedded in this CreativeWork", as well as "here is an image of this Thing".

I imagine the same debate happened in HTML5 with `<video>` vs `<embed>` (why define `video` as a first class citizen when you can just use `embed` to do it more generally), and I guess the logic goes that given that the web platform is moving towards native video in general, schema.org should follow suit.

vholland:
+1 to @philbarker's point about not changing the definition of http://schema.org/video. We have http://schema.org/subjectOf in pending.

nickevansuk:
I can see why http://schema.org/subjectOf is a great solution where the main subject of a given page is a Person, Organization, etc and authors want to mention there is a creative work about that entity (https://github.com/schemaorg/schemaorg/issues/1670), however this seems to be saying "here is an array of works related to this thing" which I would suggest is a disjoint use case compared with "here is a video of this Thing".

I suppose it comes down to the [perceived ubiquity of video](https://www.wordstream.com/blog/ws/2017/03/08/video-marketing-statistics)?

Asking the question in reverse, when `subjectOf` comes out of pending, would we advocate deprecating `image` in favour of using `subjectOf` to reference images relating to Things? If not, why not?

--

@vholland noting that you'd previously suggested `embeddedVideo` to resolve issues with changing the definition (https://github.com/schemaorg/schemaorg/issues/1447#issuecomment-329811219)?

As an aside: the JSON-LD example on [`video`](https://schema.org/video) is no longer valid as `MusicGroup` is not a `CreativeWork`.

thadguidry:
@nickevansuk Your use case of "here is a video of this Thing" is handled already.  :)  Many folks just use VideoObject or MediaObject and then to describe the content you can fill in the
"caption", "about", "contentLocation", "genre", "keywords", "mainEntity", "mentions", "spatialCoverage", "temporalCoverage".

nickevansuk:
Sure @thadguidry - we put together a [video on openactive.io](https://www.youtube.com/watch?v=a2FdmspmCNk) too to help explain it, feel like this might be the right PR to plug that ;)

In terms of fitness activities (physical activity opportunities across a range of sports, both physical and virtual): `Brand`, `Event`, and `Organization` all have videos about them. There's evidence that helping a participant understand what they can expect from an event increases their likelihood of attending, and video is a key part of that.

The Brand "Back to Netball" has a video: https://www.youtube.com/watch?v=N268gBOvnzo
The Organization "GoodGym" has a video here: https://www.youtube.com/watch?v=BPR0vhHgeWA
The Event "London Marathon" has a video here: https://www.youtube.com/watch?v=2KOLxQEfmCI

I understand that we could use subjectOf for this, however was hoping (as above) that given the amount of discussion and the existence of this pull request, that there'd be an appetite to bring `video` in line with `image`.

For example:

```
{
  "type": "Brand",
  "name": "Back to Netball",
  "description": "Running across England since 2010, over 60,000 women have taken part in Back to Netball and realised the benefits of getting involved.  From losing weight, developing the body’s cardiovascular system and improving muscle tone, taking part in Back to Netball is great for you in so many ways.",
  "logo": "http://hertsnetball.co.uk/js/plugins/imagemanager/files/B2N_logo.jpg", 
  "url": "https://www.englandnetball.co.uk/backtonetball/", 
  "video": [
    {
      "type": "VideoObject",
      "description": "Video explanation of the programme",
      "url": "https://www.youtube.com/watch?v=N268gBOvnzo",
      "duration": "T1M33S",
      "name": "What is Back to Netball?",
      "thumbnail": "http://example.com/js/plugins/imagemanager/files/video.png"
    }
  ]
}
```

thadguidry:
@nickevansuk Sounds like you just need some support help with Schema.org first and foremost.  So...please use our mailing list as it is a much better place to provide you support...and if we hit roadblocks in describing your structured data needs, THEN we can come back into various Github issues and make a case for a new type or property or clarification.  Thanks!

Example for our mailing list:  Hi ! I work for OpenActive.io , where we do/help with X ...so... How would I go about telling the world that a video contains "blah", or is about "blah", or has some X relationship to X Thing ? Here are my entities and here's the relationships that I am trying to describe about them...

Issue #1775, opened 2017-10-22
Change CategoryCodes to TermDefinitions


RichardWallis:
Following discussion in issue #1711 about the use of CategoryCodes (introduced into pending in #1255 from issue #894) there is a consensus that the terms should be renamed as follows:

- CategoryCode->TermDefinition
- CategoryCodeSet->TermDefinitionSet
- inCodeSet->inDefinedTermSet
- codeValue->termCode
- hasCategoryCode->hasTermDefinition

This issue created to carry out that work

l00mi:
@RichardWallis Thank you for pushing this!

philbarker:
@RichardWallis Thanks for the name change.

What about the changes to the definitions of DefinedTerm, DefinedTermSet and termCode?

DefinedTerm is currently 
>  a defined term

Suggest a more helpful definition of
> a word, name, acronym, phrase, etc. with a formal definition. Often used in the context of category or subject classification, glossaries or dictionaries, product or creative work types, etc. Use the name property for the term being defined, use termCode if the term has an alpha-numeric code allocated, use description to provide the definition of the term.

termCode is currently
> A short textual code that uniquely identifies the term.

Suggest that codes are only locally unique (and does textual include numbers?)
> A code that identifies this DefinedTerm within a DefinedTermSet

DefinedTermSet is currently
>A set of Defined Terms.

Suggest to help clarify what scope is
> A set of defined terms for example a set of categories or a classification scheme, a glossary, dictionary or enumeration.

RichardWallis:
Thanks @philbarker for pointing out that I forgot to change the descriptions - I will get on to it asap.

philbarker:
@RichardWallis I have made changes locally, would it help if I sent a pull request?

RichardWallis:
It would make sense for me to update the current PR (#1776), instead of creating yet another one, so I'll do that

RichardWallis:
Updated descriptions now part of PR (#1776)

RichardWallis:
Implemented in PR #1776

jaygray0919:
@RichardWallis i'm late to the party here, but would just like to confirm these is active:
```
https://schema.org/DefinedTerm
https://schema.org/CategoryCode
```
and that all of the above discussion is related to the current specification for `@DefinedTerm` and `@CategoryCode`.
We use those two types and their properties extensively; was worried that a change might be coming.

RichardWallis:
[...]

Issue #251, opened 2015-01-21
Schema.org should define how to use SKOS-like vocabularies


danbri:
Migrating in from https://www.w3.org/2011/webschema/track/issues/30

```
    Stephane Corlosquet, Jean Delahousse and others have explored proposals
    for representing SKOS in Schema.org. After much discussion on 
    public-vocabs@w3.org and amongst schema.org partners, we didn't 
    achieve consensus on doing this. A followup proposal "MiniSKOS" 
    has received broadly positive feedback. This issue tracks the 
    MiniSKOS proposal.

    See http://www.w3.org/wiki/WebSchemas/SKOS for earlier discussion, 
    http://lists.w3.org/Archives/Public/public-vocabs/2013Nov/0121.html for miniskos proposal.
```

This also relates to the issues we sometimes discuss as 'external enumerations'.

danbri:
see also https://www.w3.org/2011/webschema/track/issues/31  Schema.org miniSKOS proposal needs careful design for LRMI targetUrl property

        "From Phil Barker, http://lists.w3.org/Archives/Public/public-vocabs/2013Nov/0128.html 
        http://lists.w3.org/Archives/Public/public-vocabs/2013Nov/att-0128/00-part
        also on LRMI list, see https://groups.google.com/forum/#!topic/lrmi/jwgo02DpQzY

        Phil notes that the original miniSKOS proposal declares targetUrl as having a 
        possible expected type of 'Topic'. The properties of that Topic (described 
        locally or remotely) are similar to those of the 
        http://schema.org/AlignmentObject which has the http://schema.org/targetUrl property."

mfhepp:
See also the recent discussion re isSimilarTo, https://github.com/schemaorg/schemaorg/issues/582

PR #1776, opened 2017-10-22
Introduced TermDefinition as a super type of CategoryCode

RichardWallis:
Changed CategoryCodes to TermDefinitions Issue #1775

Updated pending definitions and associated examples.

Plus dependant terms in other areas (_MedicalCode_ & _legislationType_)

### DIFF
--- a/data/ext/health-lifesci/med-health-core.rdfa
+++ b/data/ext/health-lifesci/med-health-core.rdfa
@@ -1039,7 +1039,7 @@
   <span class="h" property="rdfs:label">MedicalCode</span>
   <span property="rdfs:comment">A code for a medical entity.</span>
   <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/MedicalIntangible">MedicalIntangible</a></span>
-  <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/DefinedTerm">DefinedTerm</a></span>
+  <span>Subclass of: <a property="rdfs:subClassOf" href="http://schema.org/CategoryCode">CategoryCode</a></span>
   <link property="http://schema.org/isPartOf" href="http://health-lifesci.schema.org" />
 </div>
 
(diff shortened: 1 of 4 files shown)


l00mi:
I am not well educated about the schema.org process. What is necessary that this PR gets accepted?

RichardWallis:
Hi @l00mi,
Process is that between releases support, or otherwise, for the proposed change is voiced in the comments stream.   Ahead of the next release (3.4 is next - date yet to be fixed) PRs are added to the list of potential additions - an aggregating issue is usually created to handle this by @danbri.

Then subject to any further comments, the list of changes for the release is commented on and agreed by the members of the steering committee.  See also http://schema.org/docs/howwework.html

~Richard.

l00mi:
Splendid, thank you for the thorough explanation @RichardWallis!

danbri:
I believe this also attempts to address https://github.com/schemaorg/schemaorg/issues/251

I suggest exit criteria for moving it out of the pending area include 1.) examination of how / whether it applies to all existing schema.org properties that kinda-sorta take controlled code lists, but which currently say something like "expects URL or Text" 2.) some thought and explanation for how this relates to the (pretty widely adopted in its area) W3C SKOS work, particularly from the perspective of parties publishing controlled code lists, taxonomies and and other such enumerations.

l00mi:
@RichardWallis if you need help with one of these tasks please ping me.

Personally I see that SKOS will be the full fledged version of a TermDefinition. 

Short possible propositions regarding SKOS:
1. Where a TermDefinition will be valid, a SKOS Concept is also valid.
2. There is a Property on the TermDefinition which is pointing to the according SKOS Concept.
3. We define that a TermDefinition Instance shall have a SKOS Concept class.

RichardWallis:
If/when this proposal moves out of pending, as @danbri commented, we need to document how TermDefinition etc. relate to other areas (especially SKOS).

Such documentation should be targeted at a few groups:

- Those that have SKOS defined data and want to add Schema to it
- Those that want to build/markup new control code lists and understand, or have never heard of, SKOS

As @danbri also comments we need to review default, and or specific, use of TermDefinition as an expected type for schema properties.

Let's see how things map out initially in pending first.

PR #1799, opened 2017-11-20
Change redundant 'text' entry to 'name' in ld+json example for Question

fisxoj:
Hello!  I did a quick look around for contribution guidelines, and hopefully this is simple enough not to be too much trouble, regardless.

The example for `Question` had the field `text` twice.  I believe the first entry is supposed to be `name`, based on the other examples for the same item.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -9540,7 +9540,7 @@ JSON:
 {
     "@context": "http://schema.org",
     "@type": "Question",
-    "text": "What is attr_accessor in Ruby?",
+    "name": "What is attr_accessor in Ruby?",
     "upvoteCount": "196",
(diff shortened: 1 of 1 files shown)


ian-axelrod:
This is a landmark PR. Thank you for all your hard work!!

gkellogg:
Note that errors like this are actually detected in Travis-CI: https://travis-ci.org/schemaorg/schemaorg, although these have only been in as unexamined warnings for some time, due to the volume of examples having semantic or syntactic problems (most of the syntactic problems have been found, at least). The idea would be to eventually make errors fail CI, which would prevent a PR from being merged until corrected, and thus not introduce new example errors.

Issue #652, opened 2015-07-21
Listing products and tools using schema.org


danbri:
Historically we have been perhaps over-cautious about linking from schema.org to products that use schema.org. However  even for the management of the vocabulary it would be good to have good examples and URLs of products that use schema.org. Whether these end up in the main site or not is something for the steering group to figure out, but I'm going to collect a few links here anyway.
## Yandex

Screenshots and docs
- organizations: https://yastatic.net/doccenter/images/help-ru/webmaster/freeze/rLWVT6HqTrWQ269UW_mVyKoLa90.png
- yandex.auto: http://yastatic.net/doccenter/images/help-ru/webmaster/freeze/Iv9k44HZTyXSVP0RHktM8UrAFUE.png
- reviews (e.g. cars): http://yastatic.net/doccenter/images/help-ru/webmaster/freeze/kvJWr-8O1SkJFjz6i4Z7uamvL2U.png
- recipes: http://yastatic.net/doccenter/images/help-ru/webmaster/freeze/QVEcFCacPwM3BEG_bn3zq5HtJzE.png
- products: http://yastatic.net/doccenter/images/help-ru/webmaster/freeze/5Zvu2GgLxbpOWfIEypOWbBh61bE.png
- directories/services - ya.auto and ya.directory
- http://yastatic.net/doccenter/images/help-ru/webmaster/freeze/Foc0WDvW8tgmhMRHQ6qz90WmVQ0.png
- English docs http://help.yandex.com/webmaster/schema-org/what-is-schema-org.xml
- Russian docs: http://help.yandex.ru/webmaster/schema-org/what-is-schema-org.xml
## Bing
- [Bing webmaster docs](http://www.bing.com/webmaster/help/marking-up-your-site-with-structured-data-3a93e731)
- [Bing markup validator](http://www.bing.com/toolbox/markup-validator)
## Yahoo
- [Glimmer, from Yahoo Research](https://lists.w3.org/Archives/Public/public-lod/2013Jun/0381.html)
## Google
- Almost everything is in http://developers.google.com/structured-data/
- See also Gmail https://developers.google.com/gmail/markup/
- Custom Search, https://support.google.com/customsearch/answer/4544182?hl=en

Aaranged:
Big +1 to building and maintaining a list of schema.org data consumers.  This will be very useful for webmasters seeking practical examples, and for marketers looking for use cases.

Some more to add:

## Pinterest

[...]

PR #1845, opened 2018-02-08
Add link to Google Rich Snippets Testing Tool

tcope25:


danbri:
Thanks for the suggestion. We may add more links to industry applications / tools at some point, but for now we are not. See also #652

PR #1856, opened 2018-03-09
Changes to support V2 of Google CSE Search API

RichardWallis:


AymenLoukil:
Hello @RichardWallis 
Why not put CSE code into external JS ?

RichardWallis:
@danbri I'm a little confused at the Travis build failure - I'm going to close this PR and create another one in a short while on a branch created directly from Master.

@AymenLoukil - it sort of is  (a JS hosted at Google).
I could rationalise the parameterised calling of that into a single local JS, which could be included in all the headers.  I probably will do at a later date, but the objective here was to make as few changes as possible as we want to apply [patch] it to the currently running version.

This doesn't negate the need to insert the search-box & results placeholders into the html on all the docs pages.  Fortunately this is templated on the dynamically created pages.

PR #1857, opened 2018-03-09
Changes to master to support V2 of Google CSE Search API

RichardWallis:
Replaces closed PR(#1856)


RichardWallis:
Yes - it applies the changes, we patched into live, into master.

PR #1859, opened 2018-03-13
changing duplicate example id

### LINKED ISSUE #1858: Privider examples are the same
I am looking at: https://schema.org/provider

I notice that there are two examples. These seem to be exactly the same. Is this a data error of some kind? Why would there be two examples instead of one? 

It would be helpful if there were one example for "person" and one example for "organization".

philbarker:
fixes #1858


danbri:
Thanks @philbarker !

Issue #1585, opened 2017-04-05
Edit vocab of RentalCarReservation to be vehicle-neutral


cougrimes:
I'm working with a major transportation and logistics company to implement schema markup across their reservation. Because they have items that are consumer-facing as well, they'd like to use RentalCarReservation to show this data to their customers. 

However, they do not rent cars. They rent trucks.

This card could also be used for other vehicle registrations, such as boats or motorcycles--could the label be revised to be RentalVehicleReservation (or even RentalReservation)?

vickitardif:
I guess it depends on what is meant by they rent trucks. For example, renting a passenger pick up truck would probably fall under RentalCarReservation, but I think we would need something different for renting moving trucks or commercial vehicles.

hey-casey:
In this exact case, this is renting moving trucks, which makes the language awkward.

hey-casey:
Just as an update/proof point: the auto extension is using RentalVehicleUsage, which would make RentalVehicleReservation a more natural alignment.

danbri:
Could we make a first pass at this through changing the textual definitions, rather than the type name? The words we choose for types are rarely perfect matches for the underlying reality. Changing type names affects all publishers and consumers, so we tend to be relatively conservative about that; whereas wording fixes are relatively easy to make...

hey-casey:
Hi @danbri , 

The core concern is around publishers (specifically in assistive technology) using rental car-specific language when displaying to end users. I'm within a few weeks of having an international rental vehicle company having schema data throughout their web and email properties to take advantage of things like Google Now, Proactive Assistant, etc. 

While all of RentalCarReservation's properties apply to vehicle reservation, it's jarring to see "Rental Car" for a 26" box truck. If there is a way to change textual definitions (and ideally alias RentalCarReservation to RentalVehicleReservation for minimal disruption) that'd work.

PR #1877, opened 2018-03-30
Rental Car-> Rental Vehicle (#1585)

cougrimes:
Updated all usage of "rental car" and "RentalCarReservation" to neutral terms "rental vehicle" and "RentalVehicleReservation"


danbri:
I don't think we can take this PR as-is, but thank you for also noting the general issue in https://github.com/schemaorg/schemaorg/pull/1877

I will close the PR here, as there are some specific problems: it edits some files like our stats that it shouldn't touch, and it makes a very substantial change (renaming terms) alongside some easier-to-make changes (wording). Can we keep talking in #1585 for now?

Issue #490, opened 2015-05-13
Site HTML should be responsive to people using tiny pocket computers


danbri:
CSS needs some attention. There are still 2 copies of the same thing (#460) in the repo, and it makes no attempt to adapt to mobile-friendly layouts.

thgh:
Here is a proposal for how schema.org could be responsive: https://schemaorg-redesign.now.sh

dbs:
It's a good start--offline support via service workers is nice--but there a few current drawbacks:

- enumeration values are missing (e.g. from https://schemaorg-redesign.now.sh/BookFormatType)
- samples are missing
- all of the RDFa has been stripped out of the pages; compare http://linter.structured-data.org/?url=http:%2F%2Fschema.org%2FProduct vs. https://schemaorg-redesign.now.sh/Product
- at least one of the design changes affects accessibility; for example, the `.strike` class (for superseded attributes like "reviews") receives an `opacity: 0.2` attribute, which makes the text a very hard to read light-grey on white. The opacity attribute appears to not be taken into account by automated accessibility audits like Lighthouse and wave.webaim.org.
- internationalization may be made more difficult by CSS `::before` being used to prepend text ("expects") in a sentence style in responsive mode

While I support the aims of the proposal, instead of a wholesale replacement I would suggest iterating towards responsiveness (e.g. start with a `<meta viewport>`, remove the `font-size: 75%` on the body element, drop the fixed-width CSS assumptions, and move from the current table layouts to something more flexible (a liberal use of `display:block` for table elements in responsive mode works, so why not)

RichardWallis:
Good suggestions.

There is some work in progress to address some infrastructure issues/potential enhancements with the application that supports the Schema.org site over the next few months.  Included in that will be enhancements to the UI.

~Richard

thgh:
Thanks for checking it out! Totally agree on iterating towards responsiveness. This was just an experiment to get to know Sapper/Svelte.

My hope is that schema.org becomes responsive soon and those were my 2 cents 😃

danbri:
Thanks again to @dbs for his changes. While there's always more we can do, I'm going to close out this issue.

BTW the site should now be a lot faster too, which ought to help for the mobile experience.

PR #1878, opened 2018-04-02
Responsive design commit series

dbs:
This branch includes a series of commits meant to improve the responsive design of schema.org for device display, and is meant to be the first steps towards addressing #490. More can certainly be done, but this is much more usable than the current experience, and makes very little impact on existing large-screen displays.

### DIFF
--- a/docs/schemaorg.css
+++ b/docs/schemaorg.css
@@ -44,7 +44,7 @@ h1
 font: bold 24px Helvetica, Arial, sans-serif;
 color: #990000;
 letter-spacing: -1px;
-margin: 2em 0 0 0;
+margin: 1em 0 0 0;
 }
 
 h2 {
--- a/docs/schemaorg.css
+++ b/docs/schemaorg.css
@@ -528,6 +528,14 @@ a.ext.ext-attic:hover {
 .atn    { color: #000; }    /* href, datetime,  */
 .custom { color: #660003; } /* itemscope, itemtype, etc,. */
 
+@media (max-width: 640px) {
+  table.definition-table th, table.definition-table td {
+    display: inline-block;
+  }
+  table.definition-table br {
+    display: none;
+  }
+}
 
 @media (min-width: 960px) {
   #mainContent, #footer, .wrapper {
--- a/docs/schemaorg.css
+++ b/docs/schemaorg.css
@@ -12,13 +12,8 @@ position: relative;
 }
 #mainContent
 {
-/* margin: 0 50px 15px 50px; */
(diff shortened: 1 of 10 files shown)


rvguha:
Before much more effort is expended in this direction, lets take a step
back.

I am really not convinced at all that the Schema.org site needs to be
redone with with 'responsive design'.

guha

dbs:
Google's "Mobile-Friendly Test" at https://search.google.com/test/mobile-friendly?id=TacwddyZ3zuudEcsBMGrIQ says "Page is not mobile friendly: This page can be difficult to use on a mobile device" and lists the following major issues:

* Text too small to read
* Viewport not set
* Clickable elements too close together

The branch I offered tackles the worst of these issues via minimal changes to the current schema.org CSS and code.

It's 2018. Having a website that requires zooming & side-to-side scrolling on phones, or that requires manually changing font size on hi-res laptop and desktop screens simply to be legible, stopped being acceptable years ago. And yes, many of us do consult schema.org on phones and hi-res laptop and desktop screens.

rvguha:
This looks fine to me.

guha

danbri:
I'm merging this in, but we have a minor conflict with the CSE fixes I've also just merged into master: https://github.com/schemaorg/schemaorg/pull/1857

The CSE fix puts generic header tags into a dedicated sub-template file which gets included during template processing: 

> ```<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />```

@dbs 's fix had a charset declaration in each file, using a different notation:

> ```<meta charset="utf-8">```

I'm resolving the conflict in favour of the abovementioned CSE-fix from  #1857 for now. Do we have any evidence that the http-equiv notation is any less mobile-friendly? It would be trivial to also add the charset attribute too, but let's not do it without reason.

dbs:
@danbri `<meta charset="utf-8">` is HTML5 vs. the older XHTML syntax. Per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta:

> The <meta> element with a charset attribute is a synonym for the pre-HTML5 `<meta http-equiv="Content-Type" content="text/html; charset=IANAcharset">`, where _IANAcharset_ contains the value of the equivalent charset attribute. This syntax is still allowed, although no longer recommended.

danbri:
@dbs thanks, how '90s of us! I've merged this with the CSE fixes and updated to HTML5. Previously we had most of the site in Polyglot semi-XMLish, but I think that ship has sailed, so I've left this as 

```<meta charset="utf-8">```

rather than

```<meta charset="utf-8" />```

Issue #1866, opened 2018-03-22
Add cryptocurrenciesAccepted property to LocalBusiness


ayumihamsaki:
Many businesses now accept cryptocurrencies as real payment exchanges. I am proposing to add the property: "`cryptocurrenciesAccepted`" to the "`LocalBusiness`".

You can use the same  ISO 4217 currency format as for `currenciesAccepted`. For example bitcoin would be "BTC". Because all cryptocurrencies use a Trading Ticker exactly the same as currencies.

I also propose updating "`paymentAccepted`" and adding "`Cryptocurrency`" to the valid list.

An example code is below:

```
<script type="application/ld+json">
{
        "@context": "http://schema.org",
        "@type": ["TouristAttraction", "AmusementPark"],
        "name": "Disneyland Paris",
        "description": "It's an amusement park in Marne-la-Vallée, near Paris, in France and is the most visited theme park in all of France and Europe.",
        "openingHours":["Mo-Fri 10:00-19:00", "Sa 10:00-22:00", "Su 10:00-21:00"],
        "isAccessibleForFree": false,
        "currenciesAccepted": "EUR",
        "cryptocurrenciesAccepted": [ "BTC", "ETH" ],
        "paymentAccepted":"Cash, Credit Card, Cryptocurrency",
        "url":"http://www.disneylandparis.it/"
}
</script>
```

I look forward to the schema team looking into this proposal in more detail.

Many Thanks.

magico:
Is there a reason why `BTC` and `ETH` can not come up under  `currenciesAccepted` ? Technically they are currencies and I don't see a benefit from separating them.

ayumihamsaki:
Central Banks and the World bank won't accept cryptocurrencies as normal currencies. So cryptocurrencies are classed as a different asset for legal reasons. That's why I suggested a new property called: "cryptocurrenciesAccepted"

ayumihamsaki:
Central Banks and the World Bank won't accept cryptocurrencies as normal currencies. So cryptocurrencies are classed as a different asset for legal reasons. That's why I suggested a new property called: "cryptocurrenciesAccepted"

See here for more details about that according to the G20: https://www.bloomberg.com/news/articles/2018-03-20/when-is-a-bitcoin-not-a-bitcoin-when-it-s-an-asset-says-g-20

"Countries increasingly see cryptos as an asset, not a currency"

RichardWallis:
[...]

PR #1894, opened 2018-04-24
Added reference to cryptocurrencies in relevant property descriptions

RichardWallis:
Issue (#1866) 
Added reference to cryptocurrencies in relevant property descriptions:

- currenciesAccepted
- discountCurrency
- price
- priceCurrency
- currency
- paymentAccepted


RichardWallis:
Updated the description additions to reference ISO 4217 formats, cryptocurrency Ticker symbols, LETS and other currency types including single inline examples for each.

Issue #1677, opened 2017-06-27
Add supporting document for autos


danbri:
Mirek & co have drafted something, staged at http://sdo-auto-fix.appspot.com/docs/automotive.html

I'd like to add this under /docs/ alongside our other supporting documentation.

Ping @vholland @scor @nicolastorzec @rvguha @tmarshbing @tilid - can you take a look?

tmarshbing:
This is very comprehensive. I like it! A few suggestions:
1. Add a link to the examples at the top. I imagine most readers will be most interested in the examples, so let's get them there quickly.
2. In the hybrid car example, shouldn't we use value instead of maxValue for the weightTotal and wheelbase attributes? It also might be nice to show the kWh for the electric engine. Finally, the pre-markup example is missing the offer that is in the markup.

sopekmir:
Thanks! We will consider adding a line with in-page links to all the sections (we have 7 of them). 
Your remarks about hybrid car example will also be taken care of!

trypuz:
@tmarshbing, thank you for your suggestions. Almost all of them have been applied (see http://sdo-auto-fix.appspot.com/docs/automotive.html), i.e.:
- we added a table of contents to the document (as @sopekmir suggested)
- we used value instead of maxValue for the wheelbase property
- we added description "Battery Energy 1.56 kWh" to the example (please note that in our example we used Ah to describe the battery capacity - see fuelCapacity property)
- we added the offer info to the pre-markup example

I think it still makes sense to use maxValue for the weightTotal property. 

@tmarshbing, would you say that the standard for measuring battery capacity is Watt Hours or Kilo-Watt Hours not of Amp Hours?

tmarshbing:
Thanks for the changes. A few things:

- I don't see the change for wheelbase. I still see maxValue.
- For weightTotal, I double checked its definition, and it looks like it can be used either to describe the max weight supported or to describe the empty weight. Apparently, you're supposed to use the name attribute on the quantitative value to differentiate. Maybe we could add name: "Permitted total weight"? I think that minValue/maxValue are only supposed to be used if expressing a range. Although we have an implied range here [empty weight, max permitted weight], I think it's still more helpful to add the name + use value.
[...]

PR #1900, opened 2018-04-27
Doc for auto.schema.org

trypuz:
Implementation of community suggestions from #1677


RichardWallis:
In the checking of this, alongside other pages, for the correctness of their HTML output (issue #1901) various errors were identified with the html.

The [W3C Html Validator ](https://validator.w3.org/ )was used to check the page.

@trypuz can you check and fix please.

Note the error for a bad element name of 'gcse:searchbox-only' can be ignored, it is part of a wider issue being fixed elsewhere.

~Richard.

Issue #1899, opened 2018-04-27
Review of updated HowTo examples


RichardWallis:
Following commit of PR #1895, the updated examples need reviewing.
- In [comment](https://github.com/schemaorg/schemaorg/pull/1895#issuecomment-384793938) @danbri raised a concern that the steps section is broken. 
- @dbiollo identified that the name of the HowToTool will not be set with this syntax.

RichardWallis:
There are two approaches to getting the name of the HowToTool, and HowToSuppy, to be picked out.
```
<div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">Spare tire</div>
```
or
```
<div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">
     <span itemprop="name">Spare tire</span>
</div>
```
My preference is for the second of these as it could be expanded thus:
```
<div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">
     <span itemprop="name">Spare tire</span>
     <img itemprop="image" src="spare-tire.jpg" />
</div>
```

dbiollo:
Maybe makes sense to show both ways in the sample? From a unscientific survey of "howtos" on various web sites, it looks like plain text tools is more common, but there are definitely a few that also show images.

If I had the job of marking up pages, I'd tend to prefer the simplest possible markup :)

RichardWallis:
Released in V3.4

PR #1902, opened 2018-04-27
Update sdo-howto-examples.txt

RichardWallis:
Fixed issues on the new HowTo examples.   Issue (#1899)

@dbiollo thanks for your input on this - I identified a mismatch between the term definitions I was working from and those already committed to webschemas.org.  It has resulted in much simpler nesting of types!

### DIFF
--- a/data/sdo-howto-examples.txt
+++ b/data/sdo-howto-examples.txt
@@ -94,94 +94,95 @@ MICRODATA:
 <div itemscope itemtype="http://schema.org/HowTo">
     <strong><span itemprop="name">Change a Flat Tire</span></strong>
     <div>About <span itemprop="estimatedCost" itemscope itemtype="http://schema.org/MonetaryAmount">
-        <meta itemprop="currency" content="USD"></meta>
-        <meta itemprop="value" content="20"></meta>
+        <meta itemprop="currency" content="USD"/>
+        <meta itemprop="value" content="20"/>
     </span>$20</div>
     <div>About <span itemprop="totalTime" content="PT30M">30 minutes</span></div>
     <div>Necessary Items:</div>
-    <div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">Spare tire</div>
-    <div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">Lug wrench</div>
-    <div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">Jack</div>
-    <div itemprop="tool" itemscope itemtype="http://schema.org/HowToTool">Wheel wedges</div>
-    <div itemprop="supply" itemscope itemtype="http://schema.org/HowToSupply">Flares</div>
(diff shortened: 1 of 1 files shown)


data/sdo-howto-examples.txt
```diff
@@ -94,94 +94,95 @@ MICRODATA:
...
+        <img itemprop="image" src="wheel-wedges.jpg"/>
+    </div>
+    <div itemprop="supply" itemscope itemtype="http://schema.org/HowToSupply">
+        <span itemprop="name">Flares</span>
+        <img itemprop="image" src="flares.jpg"/>
+    </div>
+    <div itemprop="step" itemscope itemtype="http://schema.org/HowToSection">
+        <div itemprop="name">Preparation</div>
+        <meta itemprop="position" content="1"/>
+        <div itemprop="itemListElement" itemscope itemtype="http://schema.org/HowToDirection">
```
dbiollo: I believe there should be a HowToStep wrapping this HowToDirection + HowToTip.
So the structure would be:
HowTo
  - HowToSection
     - HowToStep (aka "step 1")
        - HowToDirection
        - HowToTip
     - HowToStep (aka "step 2")
        - HowToDirection
etc

Maybe a bit verbose, but the Direction + Tip I believe were intended to be parts of an individual Step.

data/sdo-howto-examples.txt
```diff
@@ -303,131 +305,120 @@ JSON:
...
-            "name": "Wheel wedges"
+            "name": "Wheel wedges",
+            "image": "wheel-wedges.jpg"
         }
     ],
     "supply": {
         "@type": "HowToSupply",
-        "name": "Wheel wedges"
+        "name": "Flares",
+        "image": "falares.jpg"
```
dbiollo: Minor issue: typo - flares.jpg

data/sdo-howto-examples.txt
```diff
@@ -303,131 +305,120 @@ JSON:
...
-                            "text": "Don't use the wrench just yet."
-                        }
-                    ]
+    "step": [
+        {
+            "@type": "HowToSection",
+            "name": "Preparation",
+            "position": "1",
+            "itemListElement": [
+                {
```
dbiollo: Similar to comment above, this would be a list of 2 HowToStep's. 
HowToDirection & HowToTip would be itemListElement nested in the HowToStep

RichardWallis:
@dbiollo I missed the significance of the HowToStep type in the updated version for nesting pairs of HowToDirection & HowToTip.

Also fixed typo.

Thanks again for your input on this - I think we are getting there!

danbri:
Ok it looks like @RichardWallis and @dbiollo have worked this through, I will merge and publish to the staging site but do please take another sanity-check look once I've done that, this markup is really easy to get wrong.

danbri:
PTAL, http://webschemas.org/HowToStep

PR #1911, opened 2018-05-03
Fix physician subclass

Amathon:
`Physician` is an official subclass of `MedicalOrganization` (see http://schema.org/Physician) and should be reflected as such in the rdfa.


RichardWallis:
The declaration (that builds the pages on Schema.org) in the href is correct, but as you have identified the label on that declaration is incorrect to include the 'health:' prefix.

Well spotted thanks.

PR #1919, opened 2018-05-14
Fixes for the HowTo sample

dbiollo:
- This makes all children (itemListElement field) of HowToSection be of type HowToStep. Effectively wraps any HowToDirection that was previously a child of HowToSection. My understanding of the design is HowToDirection & HowToTip should only be found inside a HowToStep.
- Added "position" attribute to all instances of itemListElement
- Fixed the "...Media" properties of HowToDirection in Microdata example

This seems to fix issues found in Google Structured Data Testing Tool, though not 100% certain about the validity of the before/during/after-Media attributes.

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -11191,6 +11191,7 @@ Typical unit code(s): C62 for person</span>
     <span property="rdfs:comment">A media object representing the circumstances after performing this direction.</span>
     <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/HowToDirection">HowToDirection</a></span>
     <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/MediaObject">MediaObject</a></span>
+    <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
 </div>
 
 <div typeof="rdf:Property" resource="http://schema.org/beforeMedia">
@@ -11198,6 +11199,7 @@ Typical unit code(s): C62 for person</span>
     <span property="rdfs:comment">A media object representing the circumstances before performing this direction.</span>
     <span>Domain: <a property="http://schema.org/domainIncludes" href="http://schema.org/HowToDirection">HowToDirection</a></span>
     <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/MediaObject">MediaObject</a></span>
+    <span>Range: <a property="http://schema.org/rangeIncludes" href="http://schema.org/URL">URL</a></span>
 </div>
 
 <div typeof="rdf:Property" resource="http://schema.org/duringMedia">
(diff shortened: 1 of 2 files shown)


dbiollo:
@danbri @RichardWallis - some more fixes for the HowTo samples. Mind taking a look?
Note that I'm not totally sure if the media attributes are totally correct, but now they are consistently specified between JSON-LD, RDFa, Microdata.

RichardWallis:
@danbri  The structure looks fine.

I share your questioning thoughts about the media attributes.  They are [defined](http://webschemas.org/beforeMedia) as expecting [MediaObject](http://webschemas.org/MediaObject).  What is shown in the example are defaulting to URL, which is OK, but not very representative of what might be expected.   There again, MediaObject is probably a bit heavyweight in many real world situations, where for example a link to a YouTube video could be expected.

Perhaps adding URL to the expected type for these properties may clarify things for simplistic situations whilst not loosing the potential descriptive power of being able to use MediaObject.

dbiollo:
Here's an update with using MediaObject - *url* - property (previously it was using the *name*, at least in Google SDTT). Note that for Microdata, I didn't have to create a wrapper div, though I did for RDFa.
For MediaObject, though, it seems that *contentUrl* would be a more accurate property?

I'm gonna follow up with a change to contentUrl in this pull request.

I do like the idea of supporting URL for the media properties. Do we want to do that in this release? If so, perhaps a followup pull request after this one?

One thing comes to mind is that with MediaObject, providers could use the subtypes ImageObject or VideoObject which give semantics about what kind of media it is.

RichardWallis:
@dbiollo Thanks for your efforts here.

By default URL is supported as a value for these properties (in fact the value for [any property can be a Text, URL, or Role](http://schema.org/docs/datamodel.html#conformance)).   I would be in favour of explicitly stating this to help understanding in this area.

As it is simple to do, I suggest we could do it in the upcoming release with this PR.

@danbri what are your thoughts?

danbri:
SGTM, go for it!

dbiollo:
I've added the Docs for URL, and changed the "duringMedia" example to be just a URL. Please have a look and make sure this is what you meant.

I tested this in Google Structured Data TT and it doesn't seem to recognize the media attributes as URL. Maybe this is b/c the Range change hasn't been published? Not sure if there will be issues or if this will be OK once the schema gets released?

RichardWallis:
From looking at the sources this looks OK.  Probably needs a final look over once it is merged and put up on Webschemas.org.

Yes, the lack of recognition of URL as an expected type in the SDTT, and other testing tools, will be because it does not get updated until after a new release.

danbri:
Merged, I'll get it onto webschemas.org for final tire-kicking tomorrow... Thanks!

dbiollo:
Thanks Dan! Looking forward to trying out the updated webschemas.org!

danbri:
done - http://webschemas.org/HowTo and nearby ought to be same as github master branch now

dbiollo:
Uh-oh, the attributes table is missing from http://webschemas.org/HowTo ...

RichardWallis:
I’ll take a look.

RichardWallis:
Fixed now

dbiollo:
Thanks, looks good now!

PR #1921, opened 2018-05-17
Update releases.html

hartca:
date correction


danbri:
Thank you - fixed this independently when someone else also reported it!

Issue #1925, opened 2018-05-24
Flip the switch to use https


danbri:
@RichardWallis - can you look into this? From https://medium.com/google-cloud/app-engine-transition-your-website-to-https-b280f0bf5147 and similar it seems a change we make in the .yaml files. We need to make sure the 'http' form of the site continues to work (by redirecting).

RichardWallis:
As discussed previously (#716) this, as a simple request is comparatively easy to implement as we already handle various redirects in the user interface, and/or we could have a blanket 301 redirect to https implemented in .yaml files.  However it has some ramifications which need thinking through.

**Current state:**
- The underlying appengine instances for both webschemas and schemaorg operate in both http & https.   
- Directed from the actual domains - schema.org operates in both http & https, webschemas.org only operates in http mode.
  - To fix this, the owner of the webschems.org domain (with access to DNS configuration) needs to follow the instructions that flow from this command ```gcloud domains verify webschemas.org``` See Issue (#1739)
- There are no redirects from http to https on either configuration.
- The canonical URLs for terms shown on the term pages, regardless of method of access are http based.
- The JSON-LD Context file, retrieved by accessing either http://schema.org or https://schem.org requesting content of type ```application/ld+json```, only references  http://schema.org in its "@vocab" and "schema" entries.

**Options to consider:**
- redirect all html [only] traffic to https
  - This would result in the odd situation of a https request resulting in a page which declares a canonical URL in http form.  This is the current state for users typing https into their browsers.
  - Machine-readable versions (xml, turtle, json-ld, csv, etc. plus json-ld context) would still operate from the canonical http URL as well as https.
- Redirect all traffic to https
  - This would result in the odd situation of a https request resulting in a page which declares a canonical URL in http form.
  - Machine-readable calls will all be redirected to https, yet the URLs in the returned data definitions will be http based.
- move to https for all interactions and for term and vocabulary canonical URLs
  - This could create compatibility problems for sites that use the [canonical] schema.org URIs/URLs in the processing of consumed data or in providing supporting tools.
  - This could be partially ameliorated by including a ```sameAs``` triple, referencing the [old] http version, into to [new] https term definitions.

danbri:
Thanks for the detailed notes.  I think we're heading towards option 2.5 with 3.0 later down the line...

RichardWallis:
If we go for full [redirected] https operation, we need to fix the webschemas.org https settings.  @danbri can you organise that anyway.

[...]

Issue #1739, opened 2017-09-16
enable https on webschemas.org


danbri:
https://cloudplatform.googleblog.com/2017/09/introducing-managed-SSL-for-Google-App-Engine.html

danbri:
> Verification steps completed!
> The domain webschemas.org has been successfully verified.

millette:
All that is left is a redirect from http to https :-)

danbri:
There is some fiddlyness in the corresponding appengine cloud panels, looks easier to do for exactly "webschemas.org" than "*.webschemas.org". I am seeing "Managed certificate for webschemas.org activated." and HTTPS works for me. 

This btw also seems to mean that CORS headers are served for the webschemas copy of the site.

I have been trying to adapt our old D3 visualization to use webschemas instead of the official site, but for some reason https://webschemas.org/docs/tree.jsonld and http: variant are both 500 errors. But that's another problem!  http://bl.ocks.org/danbri/1c121ea8bd2189cf411c

danbri:
We did this

PR #1932, opened 2018-06-03
Changes to move to https only (for UI, not for vocab URIs)

RichardWallis:
Implements issue (#1925)
- .yaml changes for forced 301 redirect to https - for schema.org but not webschemas.org
  -  Webschema.org can be done retrospectively when Certificates have been updated. See Issue (#1739)
- In m/c readable outputs aded sameAs statements from http to https URIs
- Added help link from canonical URL on term pages to faq about http/https
- Added rel=canonical link to http://schema.org/{term} in term page header.


danbri:
Ok, gonna try this...

PR #1937, opened 2018-06-07
Add a direct link to Developers page from Schemas page

thadguidry:
Quite often folks ask on the mailing list where to get machine readable files, after seeing the HTML pages.  This simple change should help provide a bit more clarity of what is available right on the Schemas page.


RichardWallis:
Agree with the proposed addition of references to downloads on the schemas page.
 
However, it should be more appropriately located after the the html based options.  All users are interested in the html views of the vocabulary, whereas only a [developer] subset are interested in machine readable versions.

PR #1953, opened 2018-06-15
Fix missing / slash at beginning for a few new Types and Properties

thadguidry:
...otherwise, the links do not work.


thadguidry:
@danbri Is this fixed alternatively somewhere with deployment and Python-fu? or can this PR be merged as useful?

PR #2003, opened 2018-07-09
Improvements to descriptions of CampGround and CampingPitch 

RichardWallis:
to emphasise the differences between British & American English usage of the term 'campsite'.


RichardWallis:
Fixed typo.

I believe that the text sourced from Wikipedia sufficiently clarifies the difference between campsite (British English) and campsite (American English).  Especially when campsite is not a Schema.org type, and each respective type description starts with the clear statements:

**CampGround:**
A camping site, campsite, or _Campground_ is a place used for overnight stay in the outdoors, typically containing individual _CampingPitch_ locations. 

**CampingPitch:**
A _CampingPitch_ is an individual place for overnight stay in the outdoors, typically being part of a larger camping site, or _Campground_.

Issue #373, opened 2015-03-06
Proposal: Podcast and PodcastEpisode


Racum:
## Introduction

New schema types: **Podcast** and **PodcastEpisode**.

Podcasts are part of the common Internet slang for more than 10 years, they are everywhere, they are slowly becoming very popular, but they are not quite well represented by the Semantic Web.

So far, if anyone wanted to classify a podcast page, it suppose to use a generic “CreativeWork” type for the list of content and an “Episode” type for each instance. Expanding CreativeWork into “Podcast” type gives it the proper meaning, plus a few useful properties, and, specifying Episode into “PodcastEpisode” makes clear the difference between it and other episodes (RadioEpisode/TVEpisode).
### Definition:

A podcast, referring to a main theme containing a series of related sequential episodes.
### Inheritance

`Thing > CreativeWork > Podcast`
### Properties

| Property | Expected Type | Description |
| --- | --- | --- |
| `network` | Organization | An Organization specialized in provide or host a group of Podcasts, not only this one. “Directories” (like iTunes) does not qualify. |
| `category` | Text | The category of the Podcast. It is a good practice to follow the [taxonomy from iTunes](https://www.apple.com/itunes/podcasts/specs.html#category). |
| `rssFeed` | URL | RSS Feed URL. |
### Mapping from RSS "channel" sub-tags:

Just for reference:

| RSS Tag | Property | from |
| --- | --- | --- |
| `<title>` | name | Thing |
| `<description>` | description | Thing |
| `<link>` | url | Thing |
| `<language>` | inLanguage | CreativeWork |
| `<copyright>` | copyrightHolder | CreativeWork |
| `<managingEditor>` | editor | CreativeWork |
| `<lastBuildDate>` | dateModified | CreativeWork |
| `<image>` | image | Thing |
| `<category>` | category | Podcast |
| `<itunes:author>` | author | CreativeWork |
| `<itunes:category>` | category | Podcast |
| `<itunes:image>` | image | Thing |
| `<itunes:explicit>` | [not] isFamilyFriendly | CreativeWork |
| `<itunes:owner>` | accountablePerson | CreativeWork |
| `<itunes:subtitle>` | alternativeHeadline | CreativeWork |
| `<itunes:summary>` | text | CreativeWork |
| N/A | rssFeed | Podcast |
| N/A | network | Podcast |
### Notes
- The following `<channel>` sub-tags aren't mapped because they are implementation details:
  - `<webMaster>`
  - `<generator>`
[...]

PR #2007, opened 2018-07-13
Issue #373: Created PodcastSeries and PodcastEpisode

vholland:


unor:
### About `rssFeed`

I still think that we should use `webFeed` instead of `rssFeed`. It’s what Wikipedia calls it (article: [Web feed](https://en.wikipedia.org/wiki/Web_feed)), and the name would include formats that are not called RSS, i.e., Atom:

> The two main web feed formats are RSS and Atom. 

While "RSS feed" is [sometimes used as synonym](https://en.wikipedia.org/wiki/Web_feed#Confusion_between_web_feed_and_RSS "Confusion between web feed and RSS") for "Web feed", I don’t think that we should follow this convention for naming the property (authors searching for "RSS" will easily find the property thanks to its description). 

I would suggest to mention the two main formats in the description, e.g., something like:

> The URL for the feed (e.g., Atom or RSS) associated with the podcast series or episode.

And I wonder: get feeds really associated with a podcast *episode*? How’s that used? For jumping to specific parts in the episode?

I think this property should also be added to `WebSite`/`Blog`, but this could happen later, I guess.

vholland:
@unor I changed rssFeed to webFeed. I will remove that property from PodcastEpisode.

Racum:
Hello @vholland, can you re-issue the CI manually? it couldn't download a dependency because GitHub was unstable exactly during that time.

vholland:
@Racum Done. Yeah, I submitted during their outage. It should be fixed now.

danbri:
Shall we do this? having an approach to podcasts makes sense, but maybe a general interaction with rss/atom feeds needs discussion beyond the audio aspect...

/cc @rvguha

danbri:
We talked about this at Google and don't have a big use for this vocabulary currently, but we can park the design in case other usecases and applications arise. It was only going to be in Pending anyway, so we could go ahead and do that. Is it better to park something usably in Pending.schema.org rather than as a mere github issue? Probably...

davidbarratt:
> Shall we do this? having an approach to podcasts makes sense, but maybe a general interaction with rss/atom feeds needs discussion beyond the audio aspect...

Yeah I was kind of suprirsed that RSS feeds aren't represneted, but you can kind of do it with:
DataFeed > encoding > contentUrl
You could argue that a Podcast is just a type of datafeed... but on the flip side, it's also kind of like a TVSeries and TVEpisode (just for audio instead of video).

philliard3:
I think it's worth taking a look at http://schema.org/OnDemandEvent and https://schema.org/EventSeries to address similarities and distinctions. OnDemandEvent specifically refers to representing podcasts, but doesn't provide support for continued works that operate somewhere between an event, a series, and a creative work. There's also EventSeries to address that particular issue, but there is a disconnect in these concepts that I think the PodcastEpisode and PodcastSeries concepts introduced here highlight. There's not really a good way right now in Schema.org to represent continued content in general. Things like podcasts, web series, web comics, web novels, etc. require elements of CreativeWorkSeries and EventSeries.

Could we generalize this to work for all continued content, not just podcasts?

Doing so may help address inconsistencies between what is picked up by search engines and stored in their knowledge graphs. As an example, among two popular podcasts I listen to, one shows up in Google's results as type Podcast Feed, whereas the other has no type, only its Wikipedia summary.

Abir-H:
It would be great to get this merged

PR #2028, opened 2018-08-03
Issue1993: isBasedOn definition and examples

### LINKED ISSUE #1993: isBasedOn definition
Current definition for [`isBasedOn`](https://schema.org/isBasedOn):

> A resource that was used in the creation of this resource. This term can be repeated for multiple sources. For example, http:\//example.com/great-multiplication-intro.html.

I don’t understand what the last sentence (with the example URL) is supposed to mean (just to show how a URL value is supposed to look like?). And the second sentence doesn’t seem to be necessary, as this is the case for every property.

So I think the definition could just be:

> A resource that was used in the creation of this resource.

\+ examples that show the case of having one resource and the case of having multiple resources.

Opinions?

philbarker:
Redefines IsBasedOn, provides examples, closes #1993


data/schema.rdfa
```diff
@@ -5376,7 +5376,7 @@ Unregistered or niche encoding and file formats can be indicated instead via the
 
     <div typeof="rdf:Property" resource="http://schema.org/isBasedOn">
       <span class="h" property="rdfs:label">isBasedOn</span>
-      <span property="rdfs:comment">A resource that was used in the creation of this resource. This term can be repeated for multiple sources. For example, http://example.com/great-multiplication-intro.html.</span>
+      <span property="rdfs:comment">A resource from which this work is derived or of which it is a modification or adaption.</span>
```
thadguidry: "from which"..."from which" would be preferred, instead of "from which"..."of which"

PR #2036, opened 2018-08-10
Fix broken `url` term in `ItemList` example

BigBlueHat:
Cheers!
🎩


gkellogg:
Now I won't be able to point to this as a reason for checking keyword-like terms in JSON-LD 😄 .

BigBlueHat:
We can break something else. 😉

danbri:
Thanks!

danbri:
Thanks!

Issue #1591, opened 2017-04-17
Feature: Add property for remote with datatype boolean in JobPosting


sstankov:
Hi guys,

with more and more jobs are open to be worked from remote location, is good to add property that mark if this is possibility or not.

Keep in mind that sometimes remote is open for a country or region or other subset.

Cheers,
Slav

vholland:
Rather than a boolean a new subtype like RemoteJobPosting may be better. That would allow us to add properties to specify things like the applicable country or region.

EricAxel:
My preference is for property. There are jobs that are presented as in-office OR remote.

Additionally, could jobLocation not be used to describe a region that names the scope of "remote?"

Cheers,
--Eric

Eric Axel Franzon

vholland:
A boolean would not let an author express in-office or remote.

Overloading jobLocation means you can't differentiate between "remote within the USA" and remote or USA.

sstankov:
current there is:

`jobLocation:Place`

if we add new one:

`jobLocationRemote:Place`

if we have both, job will be allowed for on site and remote. If we have just one, we know what is the case.

I'm not sure if Place or PostalAddress is better. Pros for the first time is that we follow the same pattern, but PostalAddress is more accurate for remote usage.

danbri:
Can anyone find examples of sites where the job's remoteness appears to be explicitly represented in their infrastructure?

sstankov:
Hmm, not sure, but you can see sites like:

* https://remoteok.io or https://jobspresso.co/ or http://stackoverflow.com/jobs

But cases include:

No remote in a certain area:

<img width="1249" alt="2017-04-19_20-52-41" src="https://cloud.githubusercontent.com/assets/3425712/25196891/543565de-2542-11e7-8869-78faf28f9216.png">

Remote, without limiting the area:

<img width="795" alt="2017-04-19_20-55-05" src="https://cloud.githubusercontent.com/assets/3425712/25196982/9f5fac04-2542-11e7-819f-7f109c83e3c1.png">

RichardWallis:
> if we add new one:
> 
> jobLocationRemote:Place

Remote could mean "anywhere on the planet" - how would that work as a _Place_ ?

jvandriel:
> _"Remote could mean "anywhere on the planet" - how would that work as a Place ?"_

Wouldn't this work for that?
```
"jobLocationRemote":
{
  "@type":"Place",
  "name":"anywhere on the planet",
[...]

PR #2039, opened 2018-08-14
Issue #1591: Added jobLocationType to pending.

vholland:


danbri:
@vholland this seems good except for the little point we discussed elsewhere - that properties with "type" in their name tend to expect a bit more structure than just Text.

Here's a quick list of precedents,

```<http://schema.org/eligibleCustomerType>
<http://schema.org/publicationType>
<http://schema.org/playerType>
<http://schema.org/learningResourceType>
<http://schema.org/legislationType>
<http://schema.org/seatingType>
<http://schema.org/albumReleaseType>
<http://schema.org/priceType>
<http://schema.org/typeOfGood>
<http://schema.org/engineType>
<http://schema.org/additionalType>
<http://schema.org/alignmentType>
<http://schema.org/subtype>
<http://schema.org/fuelType>
<http://schema.org/interactionType>
<http://schema.org/additionalType>
<http://schema.org/vehicleInteriorType>
<http://schema.org/touristType>
<http://schema.org/sampleType>
<http://schema.org/albumProductionType>
<http://schema.org/exerciseType>
<http://schema.org/legislationType>
<http://schema.org/serviceType>
<http://schema.org/encodingType>
<http://schema.org/permissionType>
<http://schema.org/procedureType>
<http://schema.org/typeOfBed>
<http://schema.org/audienceType>
<http://schema.org/interactivityType>
<http://schema.org/contactType>
<http://schema.org/employmentType>
<http://schema.org/bankAccountType>
<http://schema.org/legislationType>
<http://schema.org/mapType>
<http://schema.org/lodgingUnitType>
<http://schema.org/bodyType>
<http://schema.org/loanType>
<http://schema.org/contentType>
<http://schema.org/codeSampleType>
```

via 

> grep Property all-layers.nt | awk '{print $1}' | grep -i type

danbri:
Had a quick chat with @vholland and we agreed an approach: since the trajectory here is towards eventually adding more structure, it still makes sense to use "type" in the property name, even if the initial values are just textual.

Pinging @rvguha for sanity check.

PR #2065, opened 2018-09-22
missing 12th column on line 2 for "about"

thadguidry:


RichardWallis:
@thadguidry  This fix will not roll out into a release as these files are generated by scripts at release time.
I raised an issue (#2065 2066) to identify why this has occurred.

~Richard.

PR #2082, opened 2018-10-15
Fix RDFa example for Website

lukasvarkalis:
Change meta tag attribute 'href' to 'content'


unor:
It must be a `link` instead of a `meta` element (required by the HTML spec):

```html
<link property="url" href="http://www.example.com/"/>
```

lukasvarkalis:
@unor I are right. 👍  I updated it. :)

danbri:
todo: releasenotes

PR #2100, opened 2018-11-15
Fix continuity of URL in AggregateRating example 4

jnpkr:
The `url` property in example 4 of AggregateRating doesn't reflect the original "Without Markup" information


danbri:
Thanks!

PR #2128, opened 2019-01-19
Update Gemfile

gkellogg:
to silence warnings and fix a bug detecting superseded terms.


danbri:
Thanks for the update, @gkellogg

PR #2137, opened 2019-02-05
Fix Tourist typo

atimmer:
In TouristAttraction and TouristDestination.

I left `releases/3.4` alone because I assumed that the releases should be unchanging.


RichardWallis:
@atimmer Thanks for this.  

Whilst reviewing it I spotted another typo - more technical than spelling.  As you have the branch backing this PR it might be easier for you to correct and apply to keep things simple.

In the ```rdfs:comment``` there is currently this:
``` one or more [[TouristAttractions]],```

The [[{term}]] wrapping of a term name  gets translated to ```https://schema.org/{term}``` from the current state this translates to ```https://schema.org/TouristAttractions``` which is incorrect.  To fix this it needs to read:
``` one or more [[TouristAttraction]]s,```

It would be great if you could do this.

BTW.  There is no need to apply fixes to docs/schemaorg.owl.  This file is generated from the vocabulary sources as part of the release process, so the changes you have made to that file will be overwritten at that time.

atimmer:
@RichardWallis I've applied these changes.

RichardWallis:
Thanks @atimmer

nicolastorzec:
Fixed the conflict.

PR #2146, opened 2019-02-18
Introduce Feb19 software tweaks

RichardWallis:
Remove subdomain navigation of extensions
Plus caching fixes/tweaks


danbri:
(as discussed)

Thanks, Richard. I'll test it on webschemas.org first.

danbri:
(this is live on webschemas.org now)

PR #2154, opened 2019-02-22
Display of Equivalent Property/Class links

RichardWallis:
Cleaned up and enabled code for the display of Equivalent Property/Class links. Displayed references are clickable taking user to the URI of the equivalent Property or Class.

**Example display formats:**
Equivalent Class: [dctype:Event](http://purl.org/dc/dcmitype/Event)
Equivalent Property: [dcterms:identifier](http://purl.org/dc/terms/identifier)

For terms without equvalent links nothing is displayed 

Most of code was already in place - we already output ```owl:equivalentClass``` & ```owl:equivalentProperty``` references in the on page rdfa, and in other dump formats.

Most terms only have only one equivalent - (Dataset has three)
#### identifier:
image

#### Class
image

#### legislationType
image

#### Dataset
image

#### Event
image

#### issn
image

### DIFF
--- a/apirdflib.py
+++ b/apirdflib.py
@@ -31,6 +31,22 @@ VOCABLEN = 0
 ALTVOCAB = "https://schema.org"
 STORE = rdflib.Dataset()
 #Namespace mapping#############
+EXTERNALNAMESPACES = {
+            'xsd':      'hhttp://www.w3.org/2001/XMLSchema#',
+            'skos':     'http://www.w3.org/2004/02/skos/core#',
+            'owl':      'http://www.w3.org/2002/07/owl#',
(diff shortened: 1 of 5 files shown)


danbri:
It is good to display this information but it is is for the 1% of the 1% who care about this. I wouldn't use up the first screenful of information on displaying it.

For that matter, we no longer really need to say "canonical URL", that was a workaround to stop people writing http://bib.schema.org/Person etc...

How about putting it all in the "more..." area?

RichardWallis:
Moved Canonical URL and Equivalents display into the more... area, where the 1% of 1% would hopefully find it.

**Display with more... closed:**
image
**Display with more... open:**
image

danbri:
LGTM

Issue #2194, opened 2019-04-03
DataType seems to point to itself as a parent?


mitar:
Since April 1 release, when you look at https://schema.org/DataType, it shows at the top in breadcrump:

```
DataType > DataType
```

So `DataType` is child of `DataType`? Is this on purpose? This makes it a circular definition which our parser does not consume gladly.

rok:
Is `DataType` a `DataType`?

RichardWallis:
Fixed in 3.5 post release patch fixes #2196

PR #2196, opened 2019-04-05
3.5 post release patch fixes

RichardWallis:
Small number of fixes applied after 3.5 release and applied to live site:

- Fixed rendering of docs/tree.jsonld
- Removed duplicate DataType in breadcrumbs on DataType page (issue #2194)
- From analysing site logs applied fixes:
  - Tightened up checking of invalid node names
  - Add exception catching around HeaderStore put() - contention reported under heavy loads.
  - Added a thread lock around sorting of VTerm lists - list changed during sort error reported.
- Reduced HeaderStore (mencache) write activity - only store if new header different to stored one.
- Moved call to add default http headers (including access-control-allow-origin: * ) to all generated pages. (was missing from JSONLD context file)
- Removed redundant rdfa files (containing properties with no domain or range) - they confuse some build scripts.

Tagged as "v3.5-stable"


RichardWallis:
State at time of merge, tagged as "v3.5-stable"

PR #2199, opened 2019-04-07
Grammar - remove extra yet

aretecode:


RichardWallis:
Reasonable change.

However it has been applied to a file that is automatically produced, hence will be overwritten, at version release time.

The source for this is, in RDFa form, in file ```data/ext/pending/issue-1688.rdfa```

aretecode:
@RichardWallis Thanks. I've updated the change in that file.

Issue #2173, opened 2019-03-12
Add a property on CreativeWork to specify access rights


anarchivist:
It would be useful to specify access rights for CreativeWorks. This was originally surfaced as part of the proposal made for Schema Architypes, an extension for archives and their collections, but in discussion, it became clear that this applies to lots of other domains, including other materials held by cultural heritage institutions, datasets created and used in research, and so forth. 

I believe access rights should be understood as distinct from `license` and `hasDigitalDocumentPermission`.

Examples:

An archival collection:

```jsonld
{
  "@context": "http://schema.org/",
  "@type": "Collection",
  "url": "https://www.lib.ncsu.edu/findingaids/mc00351",
  "name": "Animal Rights Network records",
  "accessRights": "This collection is open for research; access requires at least 48 hours advance notice."
}
```

A digitized letter from an archival collection:

```jsonld
{
  "@context": "http://schema.org/",
  "@type": "CreativeWork",
  "url": "https://purl.stanford.edu/bp158wj7223",
  "name": "Margaret Mitchell letter to Herschel Bricknell, 1936 August 22",
  "accessRights": "Online content is available for access via the Special Collections Reading Room."
}
```

A dataset:

```jsonld
{
  "@context": "http://schema.org/",
  "@type": "Dataset",
  "@id": "https://doi.org/10.7910/DVN/0OLJLM",
  "name": "Restricted Access WFHS Leef Manager Linking Data, Version 1.0",
  "accessRights": "Access to this dataset can be obtained by visiting the Work, Family & Health Network web page and following the instructions: http://projects.iq.harvard.edu/wfhn/restricted-access-data."
}
```

Additional context and examples:

* For archives, see standards like [ISAD(G) §3.4.1](https://www.ica.org/sites/default/files/CBPS_2000_Guidelines_ISAD%28G%29_Second-edition_EN.pdf), [DACS §4.2](https://www2.archivists.org/standards/DACS/part_I/chapter_4/1_conditions_governing_access), and [MARC21's 506 field](https://www.loc.gov/marc/bibliographic/bd506.html), and [EAD's `<accessrestrict/>` tag](https://www.loc.gov/ead/tglib/elements/accessrestrict.html). Detailed examples can be found at [#1784 (comment)](https://github.com/schemaorg/schemaorg/pull/1784#issuecomment-471126664).
* Related properties: `dct:accessRights` ([proposed in DCAT revision](https://w3c.github.io/dxwg/dcat/#license-rights))
[...]

PR #2205, opened 2019-04-10
Add new property 'accessRights' with domainIncludes 'CreativeWork'

RichardWallis:
Implements issue: #2173


RichardWallis:
Perhaps ```accessConditions``` may be more appropriate name, reflecting the advisory nature of many of the expected uses of this property.

Issue #2192, opened 2019-04-01
ISCO08 classification for occupationalCategory instead of BLS O*NET-SOC


developerbuzz:
> @developerbuzz,
> 
> It seems to me we should follow the example set in https://schema.org/Organization, where we have a field for isicV4 and naics industry codes.
> 
> When I write a script for a client, I provide both. I don't want my US client to be excluded from a Euro-centric search and vice-versa.

Whilst isicV4 would go some way and I do agree an industry classificationis would be beneficial, it doesn't tackle the classification of the occupation itself., only the industry in which the occupation resides.  ISCO08 would be a far more suitable classification for the occupation and is an international standard not just targeted to Europe or North America

_Originally posted by @developerbuzz in https://github.com/schemaorg/schemaorg/issues/1698#issuecomment-478712804_

developerbuzz:
opened in error

WeaverStever:
I disagree with the one system approach, I'd prefer to see separate properties, one for the North American SOC and another for the International ISOC(08).  While the systems are similar, they are not identical and the previous version of the ISOC was in 1988. Thirty years between updates for occupations is far too long.  The SOC (NAICS) has major updates every decade, and minor updates during the decade.  

Using https://schema.org/Organization as an example, the schema provides a field for the North American (NAICS) industry system and the ISIC international industry system.  It makes sense to do the same for occupations and let the consumer decide which system, or both to provide or query for.

Additionally, it appears the schema is missing a way to provide Classification of Instructional Programs (CIP) codes.  (College majors and minors)
https://nces.ed.gov/ipeds/cipcode/Default.aspx?y=55

I don't know if there is an international equivalent to the NCES-IPEDS CIP system.

rlzijdeman:
hi peeps, I (working with occupational classification systems as daily job) agree with @WeaverStever. Even if you were to find agreement amongst yourselves to find BLS or ISCO-08 to be 'the one classification', please know that these classifications are updated every two decades or so. (e.g. for ISCO, there's also 1968 and 1988). Thus minimally, you would need to provide the context with which you can adopt occupational classification systems (for this also check XKOS ( http://htmlpreview.github.io/?https://github.com/linked-statistics/xkos/blob/master/xkos.html )

developerbuzz:
[...]

Issue #1698, opened 2017-07-13
Proposal for new Occupation type


vholland:
A few times we have discussed adding an Occupation or Profession type (most recently issue #1410), but always from the perspective of the Person with the Occupation. For example, Jane is a Software Engineer or Joe is a Dentist.

For my work, I am more interested in the occupations themselves. What qualifications do you need? What is the estimated salary? Do these things vary by jurisdiction? (For example licensure requirements can differ from country to country.)

After reading through the past discussions, I have put together the following proposal, which I hope addresses the issues raised without having to create a lot of complexity.

- New type Occupation: A profession, may involve prolonged training and/or a formal qualification.
    - Many of the relevant properties already exist for JobPosting. The domains would be extended to include Occupation for:
        - [educationRequirements](http://schema.org/educationRequirements)
        - [experienceRequirements](http://schema.org/experienceRequirements)
        - [occupationalCategory](http://schema.org/occupationalCategory) 
        - [qualifications](http://schema.org/qualifications)
        - [responsibilities](http://schema.org/responsibilities)
        - [skills](http://schema.org/skills)
   - New properties would be:
        - **estimatedSalary**: The estimated salary for this occupation. The salary would be expressed as a MonetaryAmountDistribution. (See below.)
        - **occupationLocation**: The region/country for which this occupational description is appropriate. Note that educational requirements and qualifications can vary between jurisdictions.

- New type QuantitativeValueDistribution: A subtype of QuantitativeValue for describing a statistical distribution of values.  Properties include:
    - **duration**: The time period the distribution is measured over.
    - **percentile10**: The 10th percentile value
    - **percentile25**: The 25th percentile value
    - **median**: The median value
    - **percentile75**: The 75th percentile value
    - **percentile90**: The 90th percentile value

- New type MonetaryAmountDistribution: A subtype of QuantitativeValueDistribution for describing a statistical distribution of monetary amounts.  Properties include:
    - **currency**: The currency for the monetary values.

- New property on JobPosting:
    - **relevantOccupation**: The Occupation for the JobPosting.

- New property on Person:
    - **hasOccupation**: The Person's occupation. For past professions, use Role for expressing dates.

thadguidry:
+1 generally Vicki.  You did a great job ! covering the major potential pieces.  Let's ensure we are not missing anything else.  To that point, perhaps browsing actual datasets available can be helpful https://www.bls.gov/data/

[...]

PR #2207, opened 2019-04-10
Expanded rangeIncludes of occupationalCategory &  jobTitle 

RichardWallis:
Added CategoryCode to range of occupationalCategory and DefinedTerm to rangeIncludes of jobTitle.
Created examples for both.
Addresses issue #2192


danbri:
Thanks for the PR. I'd like @philbarker and @vholland to take a look at this (and #2192) first, as it touches on a bunch of related work.

philbarker:
I'm a little confused. occupationalCategory and relevantOccupation are both properties of JobPosting. 

-  Why? What's the difference between them? 
-  The example provided puts the occupational category code into the Occupation object of the relevantOccupation, why not into the occupationalCategory of the JobPosting direct?

Additionally, can we remove or at least loosen the "use O*NET" part of the occupationalCategory definition?

I would like to discuss this in the TalentSignal W3C group, to check with a wider set of people.

RichardWallis:
@philbarker 
> I'm a little confused. occupationalCategory and relevantOccupation are both properties of JobPosting

> Additionally, can we remove or at least loosen the "use O*NET" part of the occupationalCategory definition?

Perfectly valid questions, but I think they would be better discussed in a separate [new] issue.

This PR specifically only addresses the rangeIncludes of the two properties (occupationalCategory & jobTitle) to allow the use of predefined references.

philbarker:
@RichardWallis OK I'll raise an issue or 2, but I think the issue about the definition change from "use O*NET" is directly relevant to a PR that is only useful if other taxonomies are used.

Also, the PR includes an example which is in question.

RichardWallis:
@philbarker  Good point on the definition for occupationalCategory.

How about: 
_Category or categories describing the job. Preferably use a taxonomy such as [BLS O*NET-SOC](http://www.onetcenter.org/taxonomy.html). Repeat property for multiple categories._

Do you have suggestions with regard to the proposed examples?

philbarker:
@RichardWallis This is the definition I suggested to to TalentSignal:

> Category or categories describing the job. Use a taxonomy such as BLS O*NET-SOC http://www.onetcenter.org/taxonomy.html , ISCO-08 https://www.ilo.org/public/english/bureau/stat/isco/isco08/ or similar. Ideally the taxonomy identifier, category textual label and formal code should be provided, with the property repeated for each applicable value.
> 
> Note: for historical reasons any textual label and formal code provided as a literal may be assumed to be from O*NET-SOC

Granted it's a bit wordy, but already people are pointing out that "to correctly specify O'NET, you'd need to also specify the year of the O'NET revision you're talking about." so it may actually not do far enough. I don't know if the note is needed but I am guessing about how current tools dealing with the existing definition will continue to handle strings.

RichardWallis:
With the hrefs as links:

> Category or categories describing the job. Use a taxonomy such as [BLS O*NET-SOC](http://www.onetcenter.org/taxonomy.html), [ISCO-08](https://www.ilo.org/public/english/bureau/stat/isco/isco08/) or similar. Ideally the taxonomy identifier, category textual label and formal code should be provided, with the property repeated for each applicable value.
> 
> Note: for historical reasons any textual label and formal code provided as a literal may be assumed to be from O*NET-SOC

It's not 'that' wordy.

philbarker:
@RichardWallis after some discussion in the TalentSignal group, the consensus seems to be for:

A slight change to the definition: 

> A category describing the job, preferably using a term from a taxonomy such as [BLS O*NET-SOC](http://www.onetcenter.org/taxonomy.html), [ISCO-08](https://www.ilo.org/public/english/bureau/stat/isco/isco08/) or similar, with the property repeated for each applicable value. Ideally the taxonomy should be identified, and both the textual label and formal code for the category should be provided.
> Note: for historical reasons, any textual label and formal code provided as a literal may be assumed to be from O*NET-SOC.

An example of how to provide information for which year's edition of O*Net is being used (there's a new edition just coming out, so this will be important):
```
<script type="application/ld+json">
{
  "@context": "http://schema.org/",
  "@type": "JobPosting",
  "name": "Systems Research Engineer",
  "hiringOrganization": {
    "@type": "Organization",
    "name": "ACME Software"
  },
  "occupationalCategory": {
     "@type": "CategoryCode",
     "inCodeSet": {
          "@type": "CategoryCodeSet",
          "name": "O*Net-SOC",
          "dateUpdated": "2019",
          "url": "https://www.onetonline.org/"
       },
     "codeValue": "15-1211.00",
     "name": "Computer Systems Analysts",
     "url": "https://www.onetonline.org/link/summary/15-1121.00"
 } 
}
</script>
```
Still undecided on on using JobPosting property relevantOccupation Vs occupationalCategory. I guess options are to give examples of both (add example above to your existing example) or to provide examples only as instances of  `@type Occupation` (i.e. have no JobPosting top level object) until that question is resolved.

RichardWallis:
@philbarker thanks for this feedback.

For [occupationalCategory](https://schema.org/occupationalCategory) we have three challenges here:
1. Expanding its range to include [CategoryCode](CategoryCode).  I believe we are in agreement on that.

2. The anomaly that its domain includes both [JobPosting](https://schema.org/JobPosting) and [Occupation](https://schema.org/Occupation) .  
The result of this situation, which arose via issue #1698 (implemented in V3.5), is that when marking up a [JobPosting](https://schema.org/JobPosting), you can either use the _occupationalCategory_ property, or reference an _Occupation_, via the _relevantOccupation_ property, and then use the _occupationalCategory_ for that _Occupation_, or both.
Do we want to maintain this dual approach?  
It is understandable that you would want to define the _occupationalCategory_ for an _Occupation_ that is not associated with a JobPosting.  Equally, you would want to define the _occupationalCategory_ for a _JobPosting_ without having to create a full _Occupation_ description of the _relevantOccupation_ property.

3. If we want to keep both options, how do we construct a succinct set of examples for _JobPosting_ that do not confuse the heck out of those that view them?

philbarker:
1.  yes, we are in full agreement
2.  I don't know (sorry). Being able to link a JobPosting to an Occupation is very useful, necessary for some use cases surrounding a person who practises some occupation looking for a job. I think we want to keep this, in some ways it is a better solution that just having a category code. Specifying occupationalCategory via a blank Occupation object isn't too bad, no worse that specifying price via an Offer, but I suspect there is a lot of historical code generating job adverts that uses occupationalCategory as a direct property of JobPosting.
3.  Yep, that's the problem.

But this problem is not introduced by this pull request. I suggest trimming the examples so that they don't make point 3 worse, and attempt to address this via a separate issue. After that we can put back the more extensive examples.

Here's a trimmed version of the O*Net 2019 example (which also fixes error of dateUpdated for dateModified):
'''<script type="application/ld+json">
{
  "@context": "http://schema.org/",
  "@type": "Occupation",
  "occupationalCategory": {
     "@type": "CategoryCode",
     "inCodeSet": {
          "@type": "CategoryCodeSet",
          "name": "O*Net-SOC",
          "dateModified": "2019",
          "url": "https://www.onetonline.org/"
       },
     "codeValue": "15-1211.00",
     "name": "Computer Systems Analysts",
     "url": "https://www.onetonline.org/link/summary/15-1121.00"
 } 
}
</script>'''

vholland:
Perhaps we should expand the example to use some other properties of Occupation like qualifications or estimatedSalary.

I am not opposed to allowing DefinedTerm for jobTitle, but I rarely see controlled vocabularies here. The titles tend to be more like "Coding Ninja" than "Software Engineer".

RichardWallis:
Applied discussed change to description of _occupationalCategory_
Example replaced with one provided by @philbarker 
Updated one previous example of using _occupationalCategory_ on _JobPosting_ so that there are now two variations available.

/cc @danbri  This PR should be OK to merge now.
Area might benefit from a future review of examples, and/or an overview document for guidance around Occupations, JobPostings etc.

vholland:
Why did we remove occupationalCategory from JobPosting? This breaks a lot of existing markup on the web.

philbarker:
AFAIK, we didn't remove occupationalCategory from JobPosting, we just didn't use it in the examples.

vholland:
It looks like it moved in schema.rdfa? I understand the change, but I miss webschemas.org as an easier way to review changes.

PR #2261, opened 2019-06-04
Fixed formatting error on line 4–11

damenleeturks:
No need to have both `<br/>` and actual hard returns on something that should've been all on one line.


RichardWallis:
@damenleeturks unfortunately files such as ```all-layers-properties.csv``` are programmatically produced at release time for a new version.  So your update if merged would be overwritten.

That is not to say that there is not an issue, it however is either with the source data for [accelerationTime](https://schema.org/accelerationTime) or potentially the process that produces the csv files.  

I will close this pull request, but raise an issue to look into it.

PR #2263, opened 2019-06-04
Made all rows only one line

damenleeturks:
- add missing (if empty) column
- all rows to only one line


damenleeturks:
Closing this PR because this is a generated file, per https://github.com/schemaorg/schemaorg/pull/2261#issuecomment-498741947

PR #2285, opened 2019-06-18
Further corrections to examples identified as errors by Travis CI scripts

RichardWallis:
Results of a concentrated effort to reduce the number of errors reported by Travis CI tests when checking examples files.

Following previous efforts in this area (#2189) files from all sections are now checked.

Many individual changes/fixes to example syntax, both HTML & Schema.org, reduced number of errors reported down from 120 to a total of 45.

More work to do!


gkellogg:
Looking good, once you get them down to 0, we can change the Ruby tests to fail on error, making the whole PR system work much better at keeping errors from being introduced.

PR #2308, opened 2019-07-16
add Place type to domainIncludes of latitude and longitude properties

tjann:


tjann:
Also waiting for the CI error to be fixed

Issue #2139, opened 2019-02-07
Create a new MediaGallery type


vholland:
Schema.org has http://schema.org/ImageGallery, but increasingly pages are collections of mixed-media. It would be nice to have http://schema.org/MediaGallery for this use case. ImageGallery could then be made a subtype of MediaGallery.

danbri:
We should do this! 

Can we say anything about how MediaGallery and MediaObject relate? A mixed-media e.g. JSON-LD example would be great.

danbri:
Would ImageGallery be subtype of MediaGallery?

gmackenz:
Yes, a subtype (see [https://github.com/schemaorg/schemaorg/issues/2139#issue-407793956](vholland) above)

danbri:
Ah I missed that sentence, thanks @gmackenz!

gmackenz:
I'm almost certain I did this wrong Dan, where should have the commits been be set to push to? Certainly not schemaorg:master... Been too long since I've edited in Schema.Org (or used GitHub).

vholland:
Included in schema.org 6.0

PR #2371, opened 2019-10-21
MediaGallery

gmackenz:
#2139 

Creation of new MediaGallery type and making it the supertype for ImageGallery and VideoGallery

### DIFF
--- a/data/schema.rdfa
+++ b/data/schema.rdfa
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+/med<!DOCTYPE html>
 <html lang="en">
   <head>
     <title>Schema.org master file: RDFS in RDFa</title>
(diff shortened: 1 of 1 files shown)


danbri:
Thanks for this, @gmackenz 

Our convention recently has been for new terms to go into "Pending", which means (a) the filepath is under data/ext/pending/issue-*.rdfa and (b) there is an explicit annotation on the term to assert that it is in the pending section of schema.org. The pending files often also tie new terms into schema.org core by annotating core properties with additional domainIncludes or rangeIncludes types. In these cases the tentative pending triples doing this also live in data/ext/pending/ but should avoid making additional duplicate assertions about the labels, categories and other properties of the core terms since we already have those in schema.rdfa.

danbri:
Actually having written all that, I'm less sure it is 100% relevant in this case since we are introducing a term into the middle of the hierarchy. So there is a case to make that we just propose the new term for core without it hanging around in Pending. The UI gets a bit odd when the Pending terms are terms and are not leaf nodes in the type hierarchy, i.e. have non-pending terms beneath them. 

For WebContent type we did *not* do this, because it is less of a "no brainer" addition. I can't see anyone finding much to discuss with MediaGallery, so let's stick it straight in.

Issue #1457, opened 2016-12-14
Adding Schedules


ldodds:
A quick proposal based on [discussion](https://github.com/schemaorg/schemaorg/issues/240#issuecomment-262511081) in #240.

Schema.org doesn't currently have a way to express recurring events.

While it will always be useful to share data on individual instances, e.g. the event happening this Tuesday at 6pm. There are also scenarios where it is useful to be able to share a general rule, e.g. this event happens at 6pm every Tuesday.

Use cases include sharing data about physical activities, e.g. regular gym classes or similar events.

iCalendar already refines a model and terms for expressing recurring events. And this 
is independently supported in libraries like [rrecurjs](https://github.com/coolaj86/rrecurjs). We could draw on this to define a similar model for schema.org

The proposal is to define a new `schedule` property, a `EventSchedule` type and some additional properties for describing the schedule.

The `schedule` property will be associated with `Event`s. It will allow the description of a schedule, expressed as a recurrence rule for describing the frequency of upcoming events. 

The `EventSchedule` type will have properties based on those in iCalendar. But in the below, the naming has been aligned with existing schema.org properties.

* `startTime` - the existing property, date/time at which the schedule starts
* `endTime` - existing property, date/time at which the schedule ends
* `frequency` - "daily", "monthly", "yearly"
* `day` - DayOfWeek
* `month` - month of the year
* `hour` - hour
* `minute` - minute
* `count` - optionally specfiy the number of recurring events
* `exceptionRule` - repeating property, specifies date-times when the schedule doesn't apply, equivalent to `exrule` in ical

I think this covers the core aspects of a schedule. I may have missed something essential (am not an iCal expert) but I think this type of rule will cover many use cases so worth considering.

This structure can be easily expressed in, e.g. JSON-LD, but may be trickier in RDFa. I've not tried that yet, but wanted to float this for discussion.

magicomartinez:
Hi ldodds,

Thanks for putting this together. I had some comments on your proposal:
- `startTime`: if you’re referring to the date when the schedule starts then `startDate` may be more appropiate
- `endTime`: as above. May be worth mentioning that if no `endDate` is specified then it’s assumed the event has been scheduled indefinitely
- `frequency`: what’s the expected type here? An easy way may be to specify a number, so it knows how many to skip
- `day`: a distinction between `DayOfWeek` (e.g. to create a schedule on every Monday) and `DayOfMonth` (a new type) may be useful.
[...]

PR #2388, opened 2019-11-06
Fixing domain and range of eventSchedule property

thill-odi:
Related to issue #2277 - though there's a discrepancy between the error reported from Travis (type expected is `EventSchedule`) and the problem identified (domain is given as `Schedule` and range as `Duration`)

PR #2389, opened 2019-11-06
Issue 1457 extending temporal props [for Schedule type]

thill-odi:
Introduces the following changes to the pending Schedule type:

* adds the 'duration' property
* extends the range of 'byDay' to allow for iCal-style rrule syntax
* adds the 'scheduleTimezone' property

These are discussed at #1457 and at https://github.com/openactive/modelling-opportunity-data/issues/197.


ldodds:
@danbri are you happy with these changes? There's applying things that we've found useful in OpenActive and there's now deployed data using this and other additions.

ldodds:
@danbri It would be nice to get this added to Version 6.0 if possible along with the bug in #2388

Issue #2382, opened 2019-10-29
Allow BroadcastService to specify the language(s) they broadcast in


vholland:
It is often useful to know the language(s) a BroadcastService primarily broadcasts in. I propose expanding the domain for https://schema.org/inLanguage to include https://schema.org/BroadcastService.

vholland:
Implemented in PR #2399

vholland:
Included in schema.org 6.0

PR #2399, opened 2019-11-13
Issue #2382: Added BroadcastService to the domain for inLanguage

vholland:


danbri:
abandoned?

vholland:
Oops. No. I got a bit carried away with cleaning up old branches.

danbri:
Merged. I think these are all in the core already, so I'd have done them in data/schema.rdfa, but we can fix that later.

Issue #2421, opened 2019-12-09
Extend the domain of interactionStatistic to include Person and Organization


vholland:
Many social media sites track interaction statistics like the number of "likes", followers, etc at the person or organization level rather than per post, article, video, etc.

If we extend the domain of ```interactionStatistic``` to include ```Person``` and ```Organization```, authors could mark up these statistics.

vholland:
Implemented as PR #2423

vholland:
Included in schema.org 6.0

PR #2423, opened 2019-12-10
Issue #2421: Added Person and Organization to the domain for interactionStatistic

vholland:


danbri:
I'm not sure why this is in ext/pending/ directory since it doesn't seem to define a new pending-status term. We can move it into data/schema.rdfa with the next batch that migrate though, so leaving as-is for now.

Issue #2405, opened 2019-11-19
A proposal for two new types: Guide & Recommendation


gmackenz:
**Guide**, a new subtype of **CreativeWork**
_Thing > CreativeWork > Guide_

This is allow us to markup pages that are **Guide** pages in content, that recommend specific products or services, or aspects of a thing for a user to consider.

A **Guide** may represent a Buying Guide ([example](https://www.rei.com/learn/expert-advice/hiking-boots.html)) and detail aspects of products or services for a user to consider.

A **Guide** may represent a Product Guide ([example](https://thewirecutter.com/reviews/best-tv)) and recommend specific products or services.

A **Guide** may represent a  Ranked List ([example](https://cars.usnews.com/cars-trucks/best-midsize-cars-for-the-money)) and recommend specific products or services with ranks.

----

Relevant existing properties most likely to be used with **Guide**:

**about**
What this guide is about across all of its reviews/recommendations - hiking boots, TVs, etc.
**hasPart**
**hasPart** would contain multiple **Review**s or **Recommendation**s that are part of this guide.
**reviewAspect**
(Add this property to **Guide** and propose expanding the expected types of **reviewAspect** to include **PropertyValue** as well as the existing **Text**)
A common aspect shared across all sub-**Review**s and **Recommendation**s. For example: all recommendations (“Budget Pick”, “Top Pick”, “Best Value” ), all look at “price” and “quality” as aspects of the review.

=====================================================================

**Recommendation**, a subtype of **Review**
_Thing > CreativeWork > Review > Recommendation_

**Recommendations** are **Reviews** that suggest or propose something as the best option or best course of action.

**Recommendations** may be for products or services, or other concrete things, as in the case of a ranked list or product guide. A **Guide** may list multiple recommendations for different categories. For example, in a **Guide** about which TVs to buy, the author may have several **Recommendation**s. 

----------------------------

Relevant existing properties:

**category**
Each **Recommendation** may set the **category** property to be the type of Recommendation—for example ““Best Value”, “Budget Pick”, “Best for Families”, “Best for 2019”.

gmackenz:
Examples for above:

BuyingGuide - How to Shop for Hiking Boots
This is an example of a buying guide based on the site
https://www.rei.com/learn/expert-advice/hiking-boots.html
It tells users about the different aspects of hiking boots to consider when determining which one to buy.
Version 1, very simple:

```
{
  "@context": "schema.org",
  "@type": "Guide",
[...]

PR #2426, opened 2019-12-11
Add Guide and Recommended as new types

gmackenz:
As per issue #2405 

Add a new subtype of CreativeWork called Guide and a new subtype of Review called Recommendation (for more information see #2405).

### DIFF
--- a/data/ext/pending/issue-2405-examples.txt
+++ b/data/ext/pending/issue-2405-examples.txt
@@ -1,0 +1,440 @@
+TYPES: Guide
+
+PRE-MARKUP:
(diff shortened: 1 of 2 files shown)


RichardWallis:
Not so sure that the generically useful term **Guide** should be narrowly used for purchase/buying guides only.

What about travel guides (a set of travel instructions), tour guides (person that guides you round a tourist destination), mountain guides (person that directs you up a mountain), restaurant guides (recommended restaurants),  usage guides (the best way to use or operate a tool or product), etc.?

PR #2443, opened 2020-01-21
Revert PR2250 - Enum rationalise

RichardWallis:
Revert config to preserve previous definition files so we can revisit later.


Eyas:
:+1: I was looking at the current 6.0 schema on webschemas.org, and some of the rationalization was interesting:

- `DrugClass` has a type `MedicalEnumeration` (which has typically meant a value is an enum _value_) but appears in `domainIncludes` of the `drugClass` category as a type in and of itself.
  - If `DrugClass` is referenced as a type, then it should have the `rdf:type` of `rdfs:Class`

danbri:
Ok, we are reverted

PR #2448, opened 2020-01-26
Remove overly vague medical health properties (issue 2429)

tjann:


danbri:
Thanks @tjann! I've merged this for our next release.

I'll also add /subtype, and find some way of recording the names of deleted terms in the RDFS representation in case tools want to use that information to help people know when they're using outdated terms.

smrgeoinfo:
Isn't there some kind of status property on terms in schema.org to indicate proposed/adopted/supercededby/deprecated

Issue #2475, opened 2020-02-26
Issue with nokogiri (< 1.10.8) gem causing security alert


RichardWallis:
[Alert](https://github.com/schemaorg/schemaorg/network/alert/scripts/Gemfile.lock/nokogiri/open) flagging security problem with nokogiri gem < 1.10.8

cc. @gkellogg can you handle this for us.

gkellogg:
Will do.

RichardWallis:
Fix in PR #2476

PR #2476, opened 2020-02-26
Update Gemfile.lock, fixes nokogiri warning.

gkellogg:
At some point, the python setup might be simplified, as it repeats tests from the previous stage. All that is really necessary is example files and nquads.nt.


gkellogg:
Fixes #2475.

Issue #1842, opened 2018-02-06
Online-only events


Aaranged:
schema.org currently (as far as I know) has no way of indicating that an [Event](http://schema.org/Event) is online-only.  The ability to provide this information could be beneficial to data consumers, who would then be able to distinguish such virtual events from brick-and-mortar events and, if they wish, represent them differently in output products.

For example, Google expects a value for [location](http://schema.org/location).  While this is a Google-specific requirement, in then nonetheless highlights the fact that there's no method by which could inform Google that an Event _lacks_ a physical location.

Online events are ubiquitous and fall under many categories.  Some examples:

- [A Facebook Live event](https://live.fb.com/)
- [A Reddit AMA (Ask Me Anything)](https://www.reddit.com/r/AMA/)
- [A Twitch livestream of video game play](https://help.twitch.tv/customer/en/portal/articles/2777118-how-to-use-events)
- [A web conference (i.e. a webinar or remote meeting)](https://en.wikipedia.org/wiki/Web_conferencing)

**Possible approaches**

- Add a boolean property like `isOnline`
- Add a suitably modeled type that could be used as a value for [location](http://schema.org/location)
- Add a Event subtype like `OnlineEvent` (with possibly subtypes under that) - although an online event might _also_ be a instance of a more specific type already, like [BusinessEvent](http://schema.org/BusinessEvent)

To this last point see also #298, "applicability of Event to online events e.g. consider Webinar subtype", where @danbri notes that the "Event type was never meant to be so restricted" with regard to the "things with a physical location that you might attend in person".

**Issues and considerations**

- An Event often takes a place at a physical location _and_ is live-streamed.  Should we take account for that?
- An online event may have a URL that describes the event (and/or allows users to register for the event), and a different URL where the event is streamed.  Does this need to be accounted for, or is [url](http://schema.org/url) somehow sufficient here (though conceivably a registration URL could be declared using [Offer](http://schema.org/Offer))?
- Further to the above, an online Event is often archived after the conclusion of an event at yet a different URL.

philbarker:
This would be relevant to online courses (or rather [CourseInstance](http://schema.org/CourseInstance)s. We currently have a courseMode property which can have the value 'online'. I wouldn't mind having an OnlineEvent subtype, and using multiple types in markup for online course instances.

Putting the registration URL in the Offer would be compatible with how we dealt with Courses (though we didn't deal with registration)

[...]

PR #2488, opened 2020-03-13
Issue #1842: Added VirtualLocation

vholland:


danbri:
There is a bit of the /location URL missing in the definitions, I will fix post-merge as I have a couple more additions to make anyway.

Issue #2493, opened 2020-03-19
Add startTime/endTime to Schedule


RichardWallis:
Following issue #2486, [Schedule](https://schema.org/Schedule)  now has a start/end date for the schedule.

Unfortunately startTime and endTime are still missing.  

The Schedule description and all _Schedule_ examples include references to startTime and endTime.

PR #2494, opened 2020-03-19
Added startTime / endTime to Schedule

RichardWallis:
Implements issue #2493

### DIFF
--- a/data/ext/pending/issue-2493.rdfa
+++ b/data/ext/pending/issue-2493.rdfa
@@ -1,0 +1,14 @@
+<div>
+<!-- startDate/endDate added to Schedule, issue 2493 -->
+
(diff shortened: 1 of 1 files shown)


RichardWallis:
@ldodds @vholland 

Is this change OK with you?

vholland:
We just added startDate/endDate. Do we still need this?

RichardWallis:
Yes we do still need this.

startDate/endDate is for the start/end of the schedule
startTime/endTime is the start/end of the event on the scheduled day.

See [examples](https://schema.org/Schedule#Event-gen-241) already published

vholland:
That makes sense. Thanks.

danbri:
@tmarshbing pointed out that we really need better general documentation of modeling/patterns here.  I think the discussion above proves that point. /cc @ldodds @lucy-kind

PR #2495, opened 2020-03-19
Added example for the geographic region of a SpecialAnnouncement

vholland:


tmarshbing:
Should those be @type instead of type?

Issue #2514, opened 2020-03-30
Differentiating geographic region vs specific locations in SpecialAnnouncement


vholland:
Some of the early markup we are seeing for SpecialAnnouncement has different needs related to locations:

1. The geographic region the announcement relates to. For example, a shelter-in-place order for a city or state.

2. The specific location like a business or testing facility. For example, special opening hours for a supermarket.

3. A geographic region and specific locations. For example, locations where children in a particular city can get free meals.

I propose the following:

- The existing ```spatialCoverage``` property for the applicable region if needed.
- Creating a new property ```announcementLocation``` for specific sites related to the announcement. The expected type for ```anouncementLocation``` is ```LocalBusiness```.

As an example:

```
{
  "@context": "http://schema.org/",
  "@type": "SpecialAnnouncement",
  "category": "https://www.wikidata.org/wiki/Q81068910",
  "name": "Breakfast and lunch for school children",
  "text": "The city of Boston will provide free breakfast and lunch meals will be provided for school age children attending Boston schools.",
  "datePosted": "2020-03-17",
  "newsUpdatesAndGuidelines": "https://www.bostonpublicschools.org/Page/8098",
  "spatialCoverage": {
    "@type": "City",
    "name": "Boston, MA"
  },
  "announcementLocation": [
    {
      "@type": "School",
      "name": "Charlestown High School",
      "address": {
        "@type": "PostalAddress",
        "streetAddress": "240 Medford Street",
        "addressLocality": "Charlestown",
        "addressRegion": "Massachusetts"
      }
    },
    {
      "@type": "School",
      "name": "East Boston High School",
      "address": {
        "@type": "PostalAddress",
        "streetAddress": "86 White Street",
        "addressLocality": "Charlestown",
        "addressRegion": "Massachusetts"
      }
    }
  ]
}
```
 
This would also help prevent people from overloading ```about``` with extraneous information.

cc: @danbri @tmarshbing

danbri:
@vholland this fits with what I've been seeing too. 

The use of "about" was in our longstanding tradition of trying to re-use properties where possible. I think in this case it wasn't a terrible starting point, but I prefer the direction you outline here as it captures the common case of there being a general spatial coverage as well as specific locations mentioned in the announcement. 

@tmarshbing how does this look to you?

vholland:
In the interest of expediency, I created PR #2515

tmarshbing:
[...]

PR #2515, opened 2020-03-30
Issue #2514: Added announcementLocation and updated examples

vholland:


danbri:
Merging - will make the changes discussed in https://github.com/schemaorg/schemaorg/issues/2514 from this as a starting point. Thanks Vicki!

PR #2524, opened 2020-04-03
Fix typo in versions.json

ThibaultGerrier:
With schema.org 7.03 a JSON syntax error was introduced in versions.json. 
Seems like this is why the "download latest" at https://schema.org/docs/developers.html isn't working.
This PR fixes that typo.


danbri:
Oops, good catch, thanks @ThibaultGerrier :) will merge and fix

PR #2596, opened 2020-06-04
schema:url equivalent to foaf:page

rawmatt:
Adds an alignment between schema:url and foaf:page.

I use DCAT, Schema.org, and FOAF to describe my dataset products. The alignment between Schema.org and FOAF currently exists in DCAT https://github.com/w3c/dxwg/blob/c58b79313c10be2ec13c32d8296098fbc050d216/dcat/rdf/dcat-schema.ttl#L223, not in Schema.org. 

The effect is that even though I may have no usage of DCAT vocabulary, whether I include the DCAT definitions or not affects my data described using Schema.org. 

I believe the alignment between FOAF and Schema.org is better served being in Schema.org, rather than being in a third vocabulary.

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -9183,6 +9183,7 @@ we define a supporting type, [[SpeakableSpecification]]  which is defined to be 
     :domainIncludes :Thing ;
     :rangeIncludes :URL ;
     rdfs:comment "URL of the item." .
+    owl:equivalentProperty <http://xmlns.com/foaf/0.1/page> .
 
 :urlTemplate a rdf:Property ;
     rdfs:label "urlTemplate" ;
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -9182,7 +9182,7 @@ we define a supporting type, [[SpeakableSpecification]]  which is defined to be 
     rdfs:label "url" ;
     :domainIncludes :Thing ;
     :rangeIncludes :URL ;
-    rdfs:comment "URL of the item." .
+    rdfs:comment "URL of the item." ;
     owl:equivalentProperty <http://xmlns.com/foaf/0.1/page> .
 
 :urlTemplate a rdf:Property ;


stain:
Using `owl:equivalentProperty` means (at least in OWL) you also get the inference that the range of <http://schema.org/url> is a [foaf:Document](http://xmlns.com/foaf/spec/#term_Document) however https://schema.org/url does not require the URL to be (pointing at) a _document_.  For instance the URL can be a `#section` within a page, a JPEG picture, a mp4 video, a mp3 podcast or a web form.

With the OWL inference you also get along the inverse [foaf:topic](http://xmlns.com/foaf/spec/#term_topic) which is a bit strong implication for what may not be a document.

Personally I would think a more natural mapping of topic/page would be:

* [foaf:page](http://xmlns.com/foaf/spec/#term_page) == <https://schema.org/subjectOf>
* [foaf:topic](http://xmlns.com/foaf/spec/#term_topic) ==  <https://schema.org/about> (both inverses of above)

Then we also have the more specific subproperties on both sides:

* [foaf:primaryTopic](http://xmlns.com/foaf/spec/#term_primaryTopic) == <https://schema.org/mainEntity>
* [foaf:isPrimaryTopicOf)](http://xmlns.com/foaf/spec/#term_isPrimaryTopicOf) == <https://schema.org/mainEntityOfPage>


To be honest I am not sure what is the originally purpose of <http://schema.org/url> although its ambiguousness have always came in handy when `@id` is too strong and `subjectOf` too documenty - it's like a generic `<a href="">` without a `rel` and without having to say anything more about the URL.

I guess @danbri as original FOAF author might have views!

alex-jansen:
No activity in the last 4 years so closing as obsolete. Please feel free to add a new issue if this is still relevant.

rawmatt:
The underlying issue is DCAT makes 5 statements about the alignment between FOAF and SDO: https://github.com/w3c/dxwg/blob/c58b79313c10be2ec13c32d8296098fbc050d216/dcat/rdf/dcat-schema.ttl.

It seems to me that DCAT should not be making alignment statements about DCAT, not about FOAF and SDO. It feels rude for DCAT to add FOAF and SDO definitions. There is no way to close a vocabulary to prevent other vocabularies from changing the definitions.

I'll add an issue to DCAT.

Issue #2599, opened 2020-06-09
Add byMonthWeek as a property of schema.org/Schedule


vholland:
As we work with authors on using schema.org/Schedule, we realize there is no way to specify a schedule for particular weeks of the month. For example, a meeting that is the first and third week of each month.

Using a name similar to the existing schema.org/byMonthDay, I propose adding schema.org/byMonthWeek for this use case.

danbri:
@ldodds fyi

danbri:
@vholland do your weeks start on Sunday? Monday? Can we define a function that takes a date and tells us which week of a specified month we're in?

Glancing around,

 * Wikipedia [says](https://en.wikipedia.org/wiki/Sunday#:~:text=According%20to%20the%20Hebrew%20calendar,seventh%20day%20of%20the%20week.) "According to the Hebrew calendar and traditional Christian calendars, Sunday is the first day of the week. But according to the International Organization for Standardization ISO 8601, Sunday is the seventh day of the week."
  * timeanddate.com [says](https://www.timeanddate.com/calendar/days/monday.html), "Monday is the first day of the week according to the international standard ISO 8601, but in the US, Canada, and Japan, it's counted as the second day of the week."

vholland:
I don't think we will be able to enforce something. I added something to the description stating using schema.org/byDay is useful for clarifying for things like the 1st Monday.

Also, I converted to .ttl.

danbri:
Talked to Vicki, and realised "For example, a meeting that is the first and third week of each month." meant "For example, a meeting that is *the Tuesday* of the first and third week of each month."

Which is a lot less ambiguous; I think we should go ahead and merge this.

philbarker:
Say June begins on Saturday (and weeks begin on Sunday) is the first Saturday of June in the first week of June or the last week of May? Is the day before in first week of June? Is the first Sunday in June in the second week? I can't answer these questions in a consistent way that doesn't seem counterintuitive.

Why not have the property byMonthWeekDay and define it as the ordinal numbers of that week days in a month, e.g. first tuesday.

danbri:
Examples (from a colleague),

- An event happening every 1st and 3rd week of the month, we would then populate byMonthWeek=[1,3]
- A festival happening first friday of each month 5-7pm,  we would then populate byMonthWeek=1, byDay=Friday

PR #2600, opened 2020-06-09
Issue #2599: Added byMonthWeek to Schedule

vholland:


vholland:
I will recreate as turtle

Issue #2520, opened 2020-04-02
Allow more Expected Types for ApplicationDeadline


reginafcompton:
# Description
The WorkBasedProgram and EducationBasedProgram have an `ApplicationDeadline` field with `Date` as the expected type. 

I am currently working with Goodwill International to store their program data using Schema.org definitions (ultimately, for ingestion by Google pathways). Goodwill programs have a flexible understanding of "application deadlines," in that the deadline can be "ongoing" or "rolling." 

# Requested Change
Can the Program schema broaden how it stores information about application deadlines? E.g., allow for "Text" or "DefinedTerm" as an Expected Type, so that "rolling" or "ongoing" can be acceptable values for this field.

reginafcompton:
@vholland – I am tagging you per a suggestion from @philbarker.

Do you have any thoughts on this? I could open a PR, if you're not opposed to it. Thanks in advance!

vholland:
Supporting schema.org/Text makes sense to me.

With that said, taking off my schema.org hat and putting on my Google hat, if authors are interested in providing markup for Google features, they should follow the recommendations in the Google developer docs.

reginafcompton:
Thank you @vholland ! 

I opened a PR to address this issue: https://github.com/schemaorg/schemaorg/pull/2607 I welcome your review on it, Vicki.

(And for your "Google hat" – I agree. Goodwill plans to follow the Pathways documentation for the time being, at least until/if Pathways adjusts the ApplicationDeadline field given the (hopefully) new Schema.org definition.)

RichardWallis:
Closed in favour of that PR #2607

PR #2607, opened 2020-06-22
Issue #2520: Add Text to Range of ApplicationDeadline

reginafcompton:


RichardWallis:
@reginafcompton Unfortunately your first PR coincided with our move from rdfa to ttl.

A issue-2520.ttl version will need creating in replacement for the issue-2520.rdfa  file.

To help, here is what would be the contents:
```@prefix : <http://schema.org/> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .

:applicationDeadline a rdf:Property ;
    rdfs:label "applicationDeadline" ;
    :category "issue-2419" ;
    :domainIncludes :EducationalOccupationalProgram ;
    :rangeIncludes :Date,
        :Text ;
    :source <https://github.com/schemaorg/schemaorg/issues/2419> ;
    rdfs:comment "The date at which the program stops collecting applications for the next enrolment cycle. Flexible application deadlines (for example, a program with rolling admissions) can be described in a textual string, rather than as a DateTime. " .
```

RichardWallis:
@reginafcompton I have just checked back into this PR - I see you have been having a few issues.

- The Ruby linter issue was nothing to do with your PR - the problem has been fixed elsewhere and the checks now pass on your PR.
- The problem with your comment amendment appears to be that it resulted in a multi-line string.  In Turtle, double quotes can only encompass a single line of text.  Checking out the [Turtle Standard regarding literals](https://www.w3.org/TR/turtle/#literals) you will find that if either you should remove the line breaks from within your comment or enclosed it with triple quotes (''').

Hope that helps.

reginafcompton:
@RichardWallis - thank you for your notes! I fixed my PR (in part by directly editing an existing tty file within the `pending` dir, which I previously overlooked).

Please pass along additional feedback, and let me know any next steps I can take.

alex-jansen (approved):

PR #2620, opened 2020-06-30
Various examples added

Gnomus042:
Some JSON-LD examples added for these schemas: Airport, Apartment, BookSeries, BusStop, DefinedRegion, Dentist, OfferShippingDetails, PoliceStation, SingleFamilyResidence, TrainStation.

### DIFF
--- a/data/sdo-airport-examples.txt
+++ b/data/sdo-airport-examples.txt
@@ -1,0 +1,38 @@
+TYPES: Airport
+
+PRE-MARKUP:
+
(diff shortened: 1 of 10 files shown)


RichardWallis:
An excellent contribution @Gnomus042 !

One minor point - the value for _cutoffTime_ in the _OfferShippingDetails_ example should be a time not a date.

Gnomus042:
@RichardWallis, thanks for the comment, fixed :)

Issue #2602, opened 2020-06-11
Docker support seems broken


spacebel:
Hello,

When mounting the latest version on master to the mentioned docker image (not updated since 5 years), the image fails.

The following command fails (because site/handler.yaml does not exist) 
`docker run --rm -it --name appengine --expose 8080 -p 8080:8080 -v `pwd`:/workspace mgood/appengine-python dev_appserver.py --use_mtime_file_watcher --host=0.0.0.0 /workspace`

Moreover, if I try from the container to install the software using ./runpythonapp.sh , it states first that git is not installed, and after including git that the current directory is not schemaorg/

Is this Docker image still a possible mean for installing / running schema.org.

Thanks.

RichardWallis:
Hi @spacebel 

Although there was a proposed integration (#1007) of docker support into the Schema.org application codebase back in 2016, it was never proceeded with due to perceived complexity.

Since that time the code has moved on significantly, not least the recent move to produce a statically deployable, within Google Appengine, version.

As such the code base has never supported a docker environment.

To run a copy of the development environment you need a (linux / macos) environment with the Google Appengine SDK installed. Plus a clone of the scheaorg repo - see https://github.com/schemaorg/schemaorg/blob/master/SOFTWARE_README.md

spacebel:
Thank you. I think it would really make sense to edit https://github.com/schemaorg/schemaorg/wiki/AppEngine with this statement. The page is confusing regarding that aspect.

Regards

RichardWallis:
Wow that is way out of date!

I'll take a look at editing it.

~Richard

Issue #1903, opened 2018-04-28
How to run without Google App Engine


NeoCN:
How to run this site without Google App Engine ？ Just want to mirror schema.org because it can not be visited in China~

RichardWallis:
The data behind the site is available in various forms (JSON-LD, RDF/XML, Turtle, Triple, CSV) to download from the [Developers](http://schema.org/docs/developers.html) page.  The 'all-layers' version will contain all you need.

The html web pages are created dynamically in the App Engine so, currently, they are not available for download/mirroring or running in an another architecture.   Also the generated html includes calls to script code, for on page features such as site search and formatting etc., hosted by Google.

Making it easier for access from restricted geographies such as China is something I am looking at, alongside other architectural considerations.

I would be interested to hear how you would intend to implement such a mirror and keep it in sync with the main site.

~Richard

NeoCN:
@RichardWallis Google related services are banned in China, so if you want to make it easier for access from restricted geographies such as China, `Google Cloud` should not be used. 

My simple way to mirror and keep it sync with main site is just use `git pull` and run it under another domain~ Now I realized that it is not simple to run it without `Google App Engine`~ We can run it in Docker with `local development server` but feel weird~

RichardWallis:
In the short term that is probably the best way.  You will still have issues with included files (eg.jquery.js) sourced from Google, but that should be comparatively easy to handle..  Also the search function is handled by a Google Custom Search Engine.

To help me as I explore this area, can you share:  Is Github directly visible in China?  Are there other cloud services, such as Amazon Web Services that are visible from China?  Are there equivalent Cloud services available in China, with APIs available outside China.

NeoCN:
Github is directly visible in China. [`AWS China`](https://amazonaws-china.com/?nc2=h_lg) may be a choice.

PR #2621, opened 2020-06-30
WIP: Attempt to run schemaorg instance through Docker

stain:
Here is my successful attempt of running the schemaorg appengine server through Docker, as I could not find any instructions on how to get the Python dependencies of `runpythonapp.sh` installed locally (requirements.txt anyone?) but loads of Devs know how to run Docker.

I mark this as `WIP` as it for now only works with `docker run --net=host` as I could not find any instructions on modifying the scripts to do equivalent of `--host=0.0.0.0`. There are also too many hardcoded instances of `localhost`. The side-effect is that this docker method only works on Linux for now, as you need to bind `0.0.0.0` to escape the container.

It would also be nice if there was a way to disable the interactive prompting and put it straight in local mode - perhaps someone more knowledgable with the `runpythonapp.sh` can suggest better command line args that the container can start?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM google/cloud-sdk:latest
+RUN apt-get install -y rsync git
+ADD . /schemaorg
+#RUN mkdir -p /schemaorg
+WORKDIR /schemaorg
+RUN git submodule update --init --recursive
+EXPOSE 8080
+CMD ./runpythonapp.sh
+
--- a/SOFTWARE_README.md
+++ b/SOFTWARE_README.md
@@ -48,6 +48,20 @@ Note: If you are informed of an update to the sdopythonapp submodule, use the co
 Note: To run subdomain areas of the application on a local development system eg.  `http://bib.localhost:8080/` the subdomains will need to be added to the local system's host file.  eg. `127.0.0.1 localhost bib.localhost pending.localhost` etc.
 
 
+Docker
+------
+
+If you are using [Docker](https://www.docker.com/) you can alternatively run:
+
+    docker build -t schemaorg .
+    docker run -v `pwd`:/schemaorg --net=host -it schemaorg
+
+then access <http://localhost:8080/>
+
+Note that because the sdopythonapp currently binds to `localhost` inside the
+container, the above needs `--net=host` and will probably only work on Linux.
+
+
 Internals
 =========
 


stain:
Link to issue #2602 and #1903

RichardWallis:
I could make some comment on this if I were viewing it in context.

Have you got a step-by-step set of instructions that can take me from an empty Linux environment to this point, via getting a container with the right Python (2.7), schemaorg source ?

stain:
This Dockerfile will do exactly that, but based on [google/cloud-sdk](https://hub.docker.com/r/google/cloud-sdk/) base image so that the tricky bit of installing the cloud-sdk and Python 2.x is already there.

Then the host machine can run whatever Python it wants (or none), as it just needs [Docker](https://www.docker.com/) or [Singularity](https://singularity.lbl.gov/) to run this container.

We could then look at publishing this on https://hub.docker.com/ then the users would not even need to do the "docker build" step.

However where I need help is to fix so that runpythonapp.sh that is launched by RUN ideally:

* Binds to 0.0.0.0 so it's accessible outside the container (particularlty important on Windows/Mac that use VMs)
* Don't stop with interactive questions

stain:
https://github.com/schemaorg/schemaorg/blob/master/.travis.yml#L13 does something similar, but is only directly runnable by Travis.

christianlupus:
@stain, may I ask you what your goal is with the docker image? Building the site page locally, viewing the locally built page in a dev server or publishing to google cloud? IIf I look at the [SOFTWARE_README](https://github.com/schemaorg/schemaorg/blob/main/SOFTWARE_README.md), I do not find anything related to a file `runpythonapp.sh`.

RichardWallis:
Ahead of the upcoming V11 release, there have been some changes in the local development environment to remove some old dependencies, move to Python 3, simplify things a little, and reflect the fact that the objective has moved over time to produce a static version of the site for deployment to cloud hosting.  These were introduced in PR #2706 

This WIP activity has yet to reflect these changes.

@stain - In this updated dev environment ```runpythonapp.sh``` has effectively been replaced by ```devserv.py``` which accepts  commandline arguments for host and port.  See the updated [SOFTWARE_README](https://github.com/schemaorg/schemaorg/blob/main/SOFTWARE_README.md) for details.

alex-jansen:
Since this seems outdated I am going to close this, feel free to reopen if you disagree.

Issue #2624, opened 2020-07-02
Change /cutoffTime from Text to Time


alex-jansen:
/cutoffTime must be a Time type (instead of the current Text) and include the timezone (important, when changing from Daylight Savings Time to Standard Time and back for example)

alex-jansen:
Fixed as part of https://github.com/schemaorg/schemaorg/pull/2630

PR #2625, opened 2020-07-02
Issue #2624 (Change type of /cutoffTime from Text to Time)

alex-jansen:


danbri:
Included elsewhere, I believe.

Issue #2605, opened 2020-06-19
Add subtype of /Collection for collections of products (/ProductCollection)


danbri:
Add a type indicating a group of products that are offered as a single collection - e.g. camera plus a lens.

jvandriel:
For this I'd normally publish something like this:
```
{
  "@type":"Product",
  "name":"Camera plus a lens",
  "offers":
  {
    "@type":"Offer",
    "price":"xxx",
    "priceCurrency":"USD",
    "itemOffered":
    [
      {
        "@type":"Product",
        "name":"product 1"
      },{
        "@type":"Product",
        "name":"product 2"
      }
    ]
  }
}
```
Would ```Product``` (as mentioned in the example) become ```ProductCollection``` with this proposal?

mfhepp:
Hi all:

The original idea for bundles is

Offer -> https://schema.org/includesObject -> https://schema.org/TypeAndQuantityNode -> typeOfGood -> Product/Service

This allows specifying the quantities of the individual parts of the bundle, as well as the business function. 

Much more flexible and precise - you can also use varying units of measurement ("1 camera + 200 g of rice") etc.

https://schema.org/TypeAndQuantityNode has

    amountOfThisGood - The quantity of the goods included in the offer.
    businessFunction - The business function (e.g. sell, lease, repair, dispose) of the offer or component of a bundle (TypeAndQuantityNode). The default is http://purl.org/goodrelations/v1#Sell.
    typeOfGood -  The product that this structured value is referring to.
    unitCode / unitText - a unit of measurement for being more precise on the amount value (e.g. not units/pieces but grams, liters, gallons)

So 

    https://schema.org/includesObject

is IMO the correct way, since 2008 (1st release of GoodRelations) and 2012 (added to schema.org).

;-)

Best wishes
Martin Hepp

-----------------------------------
martin hepp  http://www.heppnetz.de
@mfhepp

jvandriel:
Thanks for pointing that out @mfhepp (I should have known that without you having to point it out). I guess the example should have been like this instead:
```
{
  "@type":"Product",
  "name":"Camera plus a lens",
  "offers":
  {
    "@type":"Offer",
    "businessFunction":"http://purl.org/goodrelations/v1#Sell",
    "price":"xxx",
    "priceCurrency":"USD",
    "includesObject":
    [
      {
        "@type":"TypeAndQuantityNode",
        "typeOfGood":
        {
          "@type":"Product",
[...]

PR #2632, opened 2020-07-06
Add collectionItem to ProductCollection

alex-jansen:
We need to add a reference to the Products that are part of a ProductCollection


danbri:
I think we've agreed to use includesObject instead, now?

jvandriel:
We discussed the use of ```includesObject``` in #2605. I was under the impression that this is how we'd do it.

danbri:
Yup.

Issue #2597, opened 2020-06-05
It's time to update our examples to use https:// 


danbri:
We should update all our examples to use https://schema.org/

Several years ago we established this as a direction, see https://schema.org/docs/faq.html#19

> There is a general trend towards using 'https' more widely, and you can already write 'https://schema.org' in your structured data. Over time we will migrate the schema.org site itself towards using https: as the default version of the site and our preferred form in examples. However 'http://schema.org' -based URLs in structured data markup will remain widely understood for the forseeable future and there should be no urgency about migrating existing data. This is a lengthy way of saying that both 'https://schema.org' and 'http://schema.org' are fine.

@tmarshbing @tilid @nicolastorzec @scor @rvguha 

/cc @RichardWallis 

If there are no objections I think we should set this in motion asap

danbri:
This was implemented in https://github.com/schemaorg/schemaorg/pull/2669 - thanks @RichardWallis

PR #2635, opened 2020-07-07
WebAPI update based on #1423

nickevansuk:
As requested by @danbri in https://github.com/schemaorg/schemaorg/issues/1423#issuecomment-465270374, and in consultation with @MikeRalphson as per discussion in https://github.com/webapi-discovery/rfcs/issues/6, this PR implements the latest recommendations from the W3C WebAPI Discovery Community Group.

@RichardWallis would be really interested in your thoughts here. See https://webapi-discovery.github.io/rfcs/rfc0001.html for the rationale behind these proposed changes.

### DIFF
--- a/data/ext/pending/issue-1423.ttl
+++ b/data/ext/pending/issue-1423.ttl
@@ -7,7 +7,11 @@
     :category "issue-1423" ;
     :isPartOf <http://pending.schema.org> ;
     :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
-    rdfs:comment "An application programming interface accessible over Web/Internet technologies." ;
+    rdfs:comment """An application programming interface accessible over Web/Internet technologies.
+
+The [[documentation]] property gives specific details of the actual endpoint instances, while the [[conformsTo]] property is used to indicate the general standard or specification that the endpoints implement. The [[url]] property indicates a landing page for the API, which is a Web page that can be navigated to in a Web browser to gain access to the API and/or additional information.
+
+      """ ;
     rdfs:subClassOf :Service .
 
 :documentation a rdf:Property ;
@@ -18,7 +22,12 @@
     :rangeIncludes :CreativeWork,
         :URL ;
     :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
-    rdfs:comment "Further documentation describing the Web API in more detail." .
+    rdfs:comment """A resource that provides further documentation of the services available via the [[WebAPI]], including their operations, parameters etc.
+
+The [[documentation]] property gives specific details of the actual endpoint instances, while the [[conformsTo]] property is used to indicate the general standard or specification that the endpoints implement. The [[url]] property indicates a landing page for the API, which is a Web page that can be navigated to in a Web browser to gain access to the API and/or additional information.
+
+Documentation may be expressed in a machine-readable format, such as an OpenAPI (Swagger) description, an OGC GetCapabilities response WFS, ISO-19142, WMS, ISO-19128, a SPARQL Service Description, an OpenSearch or WSDL20 document, a Hydra API description, otherwise in HTML, text or some other informal mode. The [[encodingFormat]] of the [[CreativeWork]] should be used to indicate the use of such formats via their MIME type.
+      """ .
 
 :termsOfService a rdf:Property ;
     rdfs:label "termsOfService" ;
@@ -30,3 +39,46 @@
     :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
     rdfs:comment "Human-readable terms of service documentation." .
 
+:endpointUrl a rdf:Property ;
+    rdfs:label "endpointUrl" ;
+    :category "issue-1423" ;
+    :domainIncludes :WebAPI ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :EntryPoint,
+        :URL ;
+    :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
+    rdfs:comment "The root location or primary endpoint of the API." .
+
+:conformsTo a rdf:Property ;
+    rdfs:label "conformsTo" ;
+    :category "issue-1423" ;
+    :domainIncludes :WebAPI ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :URL ;
+    :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
+    rdfs:comment "The URL reference of an established standard to which the described API conforms, for example `https://jsonapi.org/format/1.0/`, `https://grpc.io/`, or `http://www.hydra-cg.com/spec/latest/core/`." .
+
+:accessService a rdf:Property ;
+    rdfs:label "accessService" ;
+    :category "issue-1423" ;
+    :domainIncludes :Dataset,
+        :DataDownload ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :WebAPI ;
+    :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
+    rdfs:comment "An API that provides access to the dataset." .
+
+:apiTransport a rdf:Property ;
+    rdfs:label "apiTransport" ;
+    :category "issue-1423" ;
+    :domainIncludes :WebAPI ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :Text ;
+    :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
+    rdfs:comment "The type of transport used for the API, such as `HTTP`, `HTTPS`, `SMTP`, `MQTT`, `WS`, `WSS`." .
+
+:version a rdf:Property ;
+    :domainIncludes :WebAPI .
+
+:license a rdf:Property ;
+    :domainIncludes :WebAPI .
--- a/data/ext/pending/issue-1423-examples.txt
+++ b/data/ext/pending/issue-1423-examples.txt
@@ -6,25 +6,114 @@ See JSON example.
 
 MICRODATA:
 
-TODO
+Example is JSON-LD only.
 
 RDFA:
 
-TODO
+Example is JSON-LD only.
 
 JSON:
 
 <script type="application/ld+json">
 {
-  "@context": "http://schema.org/",
+  "@context": "https://schema.org/",
   "@type": "WebAPI",
   "name": "Google Knowledge Graph Search API",
   "description": "The Knowledge Graph Search API lets you find entities in the Google Knowledge Graph. The API uses standard schema.org types and is compliant with the JSON-LD specification.",
-  "documentation": "https://developers.google.com/knowledge-graph/",
+  "url": "https://developers.google.com/knowledge-graph/",
+  "documentation": "https://developers.google.com/knowledge-graph/reference/rest/v1",
   "termsOfService": "https://developers.google.com/knowledge-graph/terms",
+  "endpointUrl": "https://kgsearch.googleapis.com/",
   "provider": {
     "@type": "Organization",
     "name": "Google Inc."
   }
 }
 </script>
+
+TYPES: WebAPI, documentation, termsOfService, endpointUrl, apiTransport, conformsTo
+
+PRE-MARKUP:
+
+See JSON example.
+
+MICRODATA:
+
+Example is JSON-LD only.
+
+RDFA:
+
+Example is JSON-LD only.
+
+JSON:
+
+<script type="application/ld+json">
+{
+  "@context": "https://schema.org/",
+  "@type": "WebAPI",
+  "name": "Google Knowledge Graph Search API",
+  "description": "The Knowledge Graph Search API lets you find entities in the Google Knowledge Graph. The API uses standard schema.org types and is compliant with the JSON-LD specification.",
+  "documentation": [
+    {
+      "@type": "CreativeWork",
+      "encodingFormat": "text/html",
+      "url": "https://developers.google.com/knowledge-graph/reference/rest/v1"
+    },
+    {
+      "@type": "CreativeWork",
+      "encodingFormat": "application/json",
+      "url": "https://kgsearch.googleapis.com/$discovery/rest?version=v1"
+    },
+    {
+      "@type": "CreativeWork",
+      "encodingFormat": "application/vnd.oai.openapi+json;version=2.0",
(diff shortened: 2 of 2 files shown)


westurner:
Is the context still just `http://` or does `https://` work as well?

nickevansuk:
`https://` works too (see [FAQ](https://schema.org/docs/faq.html#19)).

We went with `https` here based on the intention for schema.org to move towards it (see https://twitter.com/AlanBleiweiss/status/1221906673545043971 through to https://twitter.com/danbri/status/1222043362506747904), and that a pass would be made over the examples to align them (see same thread https://twitter.com/danbri/status/1222226001326944258). Can easily change if we want to keep everything `http` until there such a pass is made?

nickevansuk:
Looking at #2597, the https change now looks imminent - so perhaps good to stick with this in here?

data/ext/pending/issue-1423.ttl
```diff
@@ -30,3 +51,46 @@
...
+    rdfs:comment "The root location or primary endpoint of the API." .
+
+:conformsTo a rdf:Property ;
+    rdfs:label "conformsTo" ;
+    :category "issue-1423" ;
+    :domainIncludes :WebAPI ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :URL ;
+    :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
+    rdfs:comment "The URL reference of an established standard to which the described API conforms, for example `https://jsonapi.org/format/1.0/`, `https://grpc.io/`, or `http://www.hydra-cg.com/spec/latest/core/`." .
```
smrgeoinfo, line 71: https://jsonapi.org/format/1.0/ is a good example
http://www.hydra-cg.com/spec/latest/core/ is  a vocabulary for describing WebAPIs, not a specification of an API. Not a good example.
https://grpc.io/ is about an RPC framework "gRPC is based around the idea of defining a service, specifying the methods that can be called remotely with their parameters and return types" Doesn't look like an API specification, but a framework for defining interface. Not a good example.

Here's some API definitions that should be better examples
https://mmisw.org/orrdoc/api/
https://www.ncdc.noaa.gov/paleo-search/api
http://www.opengis.net/doc/IS/wfs/2.0.2
http://service.iris.edu/irisws/timeseries/1/

nickevansuk: The idea with these examples is to show the breadth of `conformsTo`: it is not limited to just API specifications, but indeed any applicable standard or specification.

westurner:
When is something appropriate for endpointDescription and/or conformsTo?

nickevansuk:
@westurner see the documentation included in this PR:

https://github.com/schemaorg/schemaorg/blob/c0dabcdf099622e4b93375e06c8900e6ca96303f/data/ext/pending/issue-1423.ttl#L41

Also see notes here for the derivation: https://webapi-discovery.github.io/rfcs/rfc0001.html#new-property-endpointdescription

This distinction is taken verbatim straight out of DCAT v2, see https://www.w3.org/TR/vocab-dcat-2/#Property:data_service_endpoint_description

westurner:
Is there already an example where both are specified?

nickevansuk:
Yep, the main comprehensive covers both, though in fairness the example data is not accurate (i.e. the Google Knowledge Graph Search API does not comply with JSON:API, or any specific wider standard/spec that I could identify?). We'd have to use a different API in the example if we wanted it to include accurate data...

https://github.com/schemaorg/schemaorg/blob/c0dabcdf099622e4b93375e06c8900e6ca96303f/data/ext/pending/issue-1423-examples.txt#L58-L69

https://github.com/schemaorg/schemaorg/blob/c0dabcdf099622e4b93375e06c8900e6ca96303f/data/ext/pending/issue-1423-examples.txt#L104-L106

westurner:
While I'm sure it's not possible to satisfy every combination, it may be most helpful to include examples of at least these relatively common and/or linked data APIs:

- [ ] REST
- [ ] SPARQL
- [ ] GraphQL
- [ ] LDP

If there aren't Enumerations of the common specs, there's not likely to be any consistency in how people cite which version of the specs the API implements; so examples to copy/paste might make  adoption easier and more likely

nickevansuk:
Sure @westurner, this sounds like a good addition - either in this PR or a subsiquent PR to improve example coverage.

REST is already covered, however would you be able to create some examples of SPARQL, GraphQL, and LDP? It sounds like if we don't have time to do this right now it could easily go into a subsiquent PR, as the comprehensive example we have already gives people a pretty solid steer, so no rush here, but agreed more examples are always good.

westurner:
I'll add this to my queue. I agree that this can be done later

data/ext/pending/issue-1423.ttl
smrgeoinfo: Not very useful to point at a web page or specification that doesn't really tell you anything about the API.  I think we have a communication problem here.    Did you look at my examples?

nickevansuk: Perhaps it's worth noting that `conformsTo` is an array, so can include both specific and more generally applicable standards or specifications. 

Of your suggested examples, it looks like they cover both `endpointDescription` and `conformsTo` (noting [the distinction](https://www.w3.org/TR/vocab-dcat-2/#Property:data_service_endpoint_description) between these).

| `endpointDescription` | `conformsTo` |
|------------------------|---------------|
| https://mmisw.org/ontapi/ | https://mmisw.org/orrdoc/api/ |
| https://www.ncdc.noaa.gov/paleo-search/api | ? |
| ? | http://docs.opengeospatial.org/is/09-025r2/09-025r2.html |
| http://service.iris.edu/irisws/timeseries/1/ | ? |

In the first of these examples the link to http://sensorml.com/ontapi/ is broken, which leads me to question how up-to-date that document is if we were to reference it?

However, I see your point regarding gRPC, as arguable it fits more neatly within `apiTransport`. Hydra shapes both the design of an API and its documentation, however if controversial it can easily be removed.

Given the deadline for this release is tomorrow, I will remove gRPC and Hydra from the examples for now. We can always create a subsequent PR with more examples later on.

nickevansuk: CC @MikeRalphson

nickevansuk:
Note the gRPC and Hydra examples have been removed from `conformsTo` as per https://github.com/schemaorg/schemaorg/pull/2635#discussion_r470522975

data/ext/pending/issue-1423.ttl
smrgeoinfo: Good point that the distinction between a specification and an endpoint description can be pretty fuzzy, especially in the case of the numerous APIs that are one-ofs.   I thought the logic of endpointDescription was that it should be a link to  machine readable (actionable) document like an OpenAPI or OGC getCapabilities instance (see https://github.com/webapi-discovery/rfcs/issues/13).  If there's a web page that describes the API for people, that's pretty much a specification, but it might better be considered so:documentation (see https://github.com/webapi-discovery/rfcs/issues/10).   In an ideal world someone defining an API that they think someone else might want to use or describe as a so:WebAPI, they should specify a URI (versioned!) to use in so:conformsTo.  This is wishful thinking at this point, so people are going to be guessing what to put in conformsTo or documentation....

let's see how it plays out in the wild when people start using it.

by the way, an endpoint description associated with http://docs.opengeospatial.org/is/09-025r2/09-025r2.html would be a URL for the OGC GetCapabilities instance associated with a service instance (e.g. https://geoservices.gelderland.nl/server/services/Inspire/INSPIRE_PS/MapServer/WFSServer?request=GetCapabilities&service=WFS)

dr-shorthair: So just to spell it out, in this case 

conformsTo = http://docs.opengeospatial.org/is/09-025r2/09-025r2.html (actually http://www.opengis.net/doc/IS/wfs/2.0.2) 

endpointDescription = https://geoservices.gelderland.nl/server/services/Inspire/INSPIRE_PS/MapServer/WFSServer?request=GetCapabilities&service=WFS)

nickevansuk: Yes good catch, indeed I'd confused our own definitions of `endpointDescription` and `documentation` above.

The table should in fact read:

| `documentation` | `endpointDescription` | `conformsTo` |
|------------------|------------------------|---------------|
| https://mmisw.org/ontapi/ | https://mmisw.org/ontapi/swagger_mmiorr.yaml | https://mmisw.org/orrdoc/api/ |
| https://www.ncdc.noaa.gov/paleo-search/api | ? | ? |
| ? | https://geoservices.gelderland.nl/server/services/Inspire/INSPIRE_PS/MapServer/WFSServer?request=GetCapabilities&service=WFS | http://www.opengis.net/doc/IS/wfs/2.0.2 |
| http://service.iris.edu/irisws/timeseries/1/ | ? | ? |

I've clarified the definitions in this PR, to minimise such potential confusion: https://github.com/schemaorg/schemaorg/pull/2635/commits/33b33fe1c26f97becc98d7c982004693dcc42818

Thanks both!

nickevansuk:
Additionally have simplified references to `documentation` and `endpointDescription` in https://github.com/schemaorg/schemaorg/pull/2635/commits/33b33fe1c26f97becc98d7c982004693dcc42818, as per https://github.com/schemaorg/schemaorg/pull/2635#discussion_r471432630

danbri (changes requested):
Why is the property called 'endpointUrl'? isn't it just a property whose value is the endpoint (an 'EntryPoint') that might have an 'url' property of its own?

data/ext/pending/issue-1423.ttl
```diff
@@ -30,3 +45,46 @@
...
+    rdfs:label "endpointUrl" ;
+    :category "issue-1423" ;
+    :domainIncludes :WebAPI ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :EntryPoint,
+        :URL ;
+    :source <https://github.com/schemaorg/schemaorg/issues/1423> ;
+    rdfs:comment "The root location or primary endpoint of the API." .
+
+:conformsTo a rdf:Property ;
```
danbri, line 58: It's been suggested (uk govt folks - I'll ask them to share the comment) that we add a more general 'conformsTo' to Schema.org, broadly in line with the Dublin Core term. I think we should do that - but the definition would need to be more general than APIs. I think it would still be applicable here.

danbri (changes requested):
On the /conformsTo comment - see also https://github.com/schemaorg/schemaorg/issues/1516 for the case to add it more generally than just for APIs. Replace mention of API with "to which the thing described conforms. Broadly equivalent to Dublin Core's "conformsTo" property - https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/conformsTo. We should set domainIncludes to at least /CreativeWork, possibly wider later.

nickevansuk:
> Replace mention of API with "to which the thing described conforms. Broadly equivalent to Dublin Core's "conformsTo" property - https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/conformsTo.

Included in commit https://github.com/schemaorg/schemaorg/pull/2635/commits/c90261b64067a1e1241bf113e4b51d42da079784. @danbri  is that what you meant, or shall we remove the reference to Dublin Core from the description?

nickevansuk:
> Why is the property called 'endpointUrl'? isn't it just a property whose value is the endpoint (an 'EntryPoint') that might have an 'url' property of its own?

So there are a few reasons for this:

1. The naming of the property follows from [DCAT v2's `dcat:endpointURL`](https://www.w3.org/TR/vocab-dcat-2/#Property:data_service_endpoint_url)

2. The inclusion of both `EntryPoint` and `URL` in the range mirrors the existing range of the [`documentation`](https://schema.org/documentation) property and the use of the [`target` property within the Actions document](https://schema.org/docs/actions.html), to allow for [simple usage](https://webapi-discovery.github.io/rfcs/rfc0001.html#examples)
https://github.com/schemaorg/schemaorg/blob/c90261b64067a1e1241bf113e4b51d42da079784/data/ext/pending/issue-1423.ttl#L53-L54

3. The "endpoint" itself is described by machine-readable documentation referenced by the [`endpointDescription`](https://webapi-discovery.github.io/rfcs/rfc0001.html#examples%E2%91%A0) property. Part of the intention of the name `endpointUrl` is to avoid [confusion](https://github.com/webapi-discovery/rfcs/issues/15) around this property potentially being used as a replacement for complete machine-readable documentation. It is strictly "The root location or primary endpoint of the API", potentially covering [multiple environments](https://webapi-discovery.github.io/rfcs/rfc0001.html#examples), rather than attempting to compete with OpenAPI etc.

Alternative names include:
- `endpointPrimaryUrl`
- `endpointBaseUrl`

However depending on the API, the value could represent the base URL or primary endpoint... hence `endpointUrl`...

@danbri interested in your thoughts?

nickevansuk:
@danbri just bumping this, see above - as it seems 99% there?

nickevansuk:
@danbri I've removed references to `endpointUrl` from this PR given your feedback above, and the risks around it being [misunderstood](https://github.com/webapi-discovery/rfcs/issues/15#issuecomment-655364506) if named e.g. `endpoint`. We can easily include this in a future PR after further discussion.

It appears we have consensus on everything else?

nickevansuk:
@danbri just nudging this, seems like it's just a case of you confirming you're happy with the above? https://github.com/schemaorg/schemaorg/pull/2635#issuecomment-724761653

MikeRalphson:
@danbri also popping in to ask if there's anything we can do to get this moving again? Very keen to get `WebAPI` out of Pending, and to adopt at least a subset of the proposed extension.

nickevansuk:
Another nudge on this @danbri - what are the next steps here?

smrgeoinfo:
why is this languishing?

gmackenz (changes requested):
The Knowledge Graph Search API (https://kgsearch.googleapis.com) no longer exists. It was shut down as an external API a few years ago. This WebAPI change needs to be reworked with another example. If you want to stick with Google, I suggest the YouTube API (https://developers.google.com/youtube/v3)

nickevansuk:
Thanks @gmackenz - once we make these changes, and given all other comments have been addressed, would you be happy to merge?

gmackenz:
> Thanks @gmackenz - once we make these changes, and given all other comments have been addressed, would you be happy to merge?

If we are all in agreement, yes.  It doesn't have to be my suggestion for the YouTube API, just one that is in use currently.

PR #2637, opened 2020-07-09
Bundler update

gkellogg:
This updates the Gemfile, which fixes the security warning, but still does not allow SOFT_LINT to be turned off. There are a number of semantic errors with the examples still being reported.

Note that the Ruby lint stage could be made faster if scripts/buildreleasefiles.sh did what it suggests, and not also run unit tests and other things that aren't really necessary for setting up the release files.


danbri:
Thank you @gkellogg - much appreciated!

PR #2642, opened 2020-07-09
Issue 2636 create PracticeProblem schema

lucy-kind:
Create PracticeProblem type, and add hasPart property to existing Quiz and Course types with value of PracticeProblem.


lucy-kind:
hasPart ttl code removed; also added examples file!

PR #2643, opened 2020-07-09
Gemfile update

gkellogg:
This reduces noise in the output leaving things which are actual semantic (rather than syntactic) errors.

It adds a -t option to buildreleasefiles to not perform unit tests and optimizes the use of the script from .travis.yml to substantially improve the build stage.


danbri:
Thanks again, @gkellogg :)

RichardWallis:
@gkellogg  FYI there was a minor fix required to buildreleasefiles.sh (fixed in commit 6a1047a).
Thanks again for diving in here.
If you see benefits from other tweaks in this area, ping me.

Issue #2304, opened 2019-07-11
Rationalisation of dump files required


RichardWallis:
As identified in issues #2301 & #2302; within the definition download files, there is not full segregation between the terms defined in the core and terms defined in sections (extensions).

eg.The pending defined [CssSelectorType](https://schema.org/CssSelectorType), which is defined as the range for [cssSelector](https://schema.org/cssSelector) property, defined in the core vocab. As the _schema_ dump file only contains terms defined in the core, _CssSelectorType_ definition doesn't appear in that file.

Anecdotally, it appears that the main down of dump files are for '_schema_' (core only) the default offered on the [downloads page](https://schema.org/docs/developers.html#defs) and the '_all-layers_' which contains all definitions (including those in the _attic_ section).

I recommend that we rationalise the types of dump files offered (currently 8) down to 2 with the following contents:

- _schema_ - containing all sections except _attic_.
- _all-layers_ - containing all sections including _attic_.

VladimirAlexiev:
#2537 shows that the current reduced `schema` lacks closure (some classes are referenced but not defined)

danbri:
I agree. A few years ago we introduced a fair amount of engineering to support layered extension. In practice it turned out too difficult to disentangle various overlapping areas. Should we have an "extension" for Cultural Heritage? for Galleries / Archives / Libraries / Museums (GLAM)? Or different named efforts for bibliography and archival vocabulary? What about tourism, travel, real estate, and e-commerce? Where should ebooks be described? etc etc.

In practice we have gradually migrated back towards a simple structure.

 * Schema.org is Schema.org
   * It has an area called "attic" for things we did that we can't undo, but don't want to highlight
   * It has an area called "pending" for things we are doing, where we want to highlight a heightened potential for ongoing changes and the need for feedback

We have already downplayed the use of named subdomain extensions ("pending", "bib", etc.) in the site navigation. We have found little use for the defensive layering structure which attributed each triple in the schema definitions to one of those sections. In practice, "all layers" is the only sensible subset to use. There should be a simple triple representation, a consolidated Turtle representation (which looks as similar as possible to the MCF-flavoured Turtle source files), and I guess there is a desire for JSON-LD. Do we believe the CSV files have proved useful?

For naming I'd suggest something like 

 * "schemaorg-current" (everything minus attic)
 * "schemaorg-current-plus-removed" (everything including attic)

RichardWallis:
[...]

Issue #2516, opened 2020-04-01
https everywhere


Bcadej:
On website https://schema.org every example is writen without s -> http://schema.org

Fix all documentation to https://schema.org

RichardWallis:
@Bcadej  What you are seeing is the difference between the URL of a page on the Schema.org website (_https based_), and the the (_http based_) unique identifier (URI) for the structured data vocabulary term that the page describes.

Those http based identifiers have been associated with their terms since the start of Schema.org in 2012.  Those values will be found in datasets, queries, and code far and wide across the web.  Therefore just changing them to new https based identifiers is not as simple as it sounds.  The ramifications could be significant for established processes.

This issue was discussed in detail in issue #2018, in reference to the term canonical value displayed on each page (under the [more...] tab).  Check this out for more information.

danbri:
@RichardWallis actually we could update most of the *examples*. The use of specific URIs in our data dumps and schemas is, as you say, a more complicated business. But we have long had consensus that people can use https: in their markup (see FAQ, https://schema.org/docs/faq.html#19 ).

EmidioStani:
or adding owl:sameAS between the 2 URI so people can effectively use them both

vberkel:
@RichardWallis I no longer see the [more...] tab that shows the HTTP Canonical URI (as well as Leave public feedback and Check for open issues). This is what we used to see:
![image](https://user-images.githubusercontent.com/1944680/85565065-c9c8fd80-b5fc-11ea-8983-c31c955cfe0d.png)

RichardWallis:
See issue #2610 for fix to [more...] issue.

datadavev:
The ESIP science-on-schema.org group is in the process of providing implementation guidance for adoption of the `https://schema.org/` prefix for term definitions in JSON-LD markup for schema.org. This requires a mechanism to reliably dereference the context document for `http` vs. `https` based term definitions. 

What are the plans with respect to dereferencing the context for schema.org markup in JSON-LD? Can we assume a `http` request will return a context with `http` schema.org term definitions, and a `https` request will return a context file with `https` term definitions?

danbri:
See also https://github.com/RDFLib/rdflib/issues/1120#issuecomment-668536109 for discussion on canonicalizing from http to https URIs in Python rdflib

datadavev:
[...]

PR #2654, opened 2020-07-16
Rationalisation of dump files

RichardWallis:
Simplified and expanded dump files and updated associated documentation.

Results in:  
- _schema_.* files now contain all terms.
- _all-layers._* are duplicates of these updated _schema_.* files
- updates to _docs/developers.html_ to reflect these changes

Implements issue #2304

**Note:** The associated tweaks to the documentation assumes this PR will be merged and shipped in V9.0.  If it does not make it into V9.0, there will be a need for a very minor tweak to the wording.

### DIFF
--- a/scripts/Rakefile
+++ b/scripts/Rakefile
@@ -95,7 +95,7 @@ file "spec/schema.rb" => :do_build do
   cmd += " --class-name SCHEMA"
   cmd += " --strict"
   cmd += " -o spec/schema.rb_t"
-  cmd += " #{rel}/schemaorg-current.nq"
+  cmd += " #{rel}/schemaorg-current-http.nq"
   puts "  #{cmd}"
   begin
     %x{#{cmd} && mv spec/schema.rb_t spec/schema.rb}
--- a/scripts/Rakefile
+++ b/scripts/Rakefile
@@ -95,7 +95,7 @@ file "spec/schema.rb" => :do_build do
   cmd += " --class-name SCHEMA"
   cmd += " --strict"
   cmd += " -o spec/schema.rb_t"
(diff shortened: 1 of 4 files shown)


danbri:
My review comments are in https://github.com/schemaorg/schemaorg/issues/2304

RichardWallis:
Following [comments on issue](https://github.com/schemaorg/schemaorg/issues/2304#issuecomment-659426278), this PR has been updated to reflect the following file name changes and contents:
- _schemaorg-current_ - contains all terms from all sections
- _schemaorg-all_ - contains all terms from all sections **plus** attic

danbri:
I've suggested to @RichardWallis that we address the desire for https-based URIs in our data dumps here, but adding parallel definitions (for a year or two) that use https spelling.

* schemaorg-current-http
* schemaorg-current-https
* schemaorg-all-http
* schemaorg-all-https

https://github.com/schemaorg/schemaorg/issues/2516


Regarding mappings between the two, I would be OK with adding OWL https://www.w3.org/TR/owl-ref/#equivalentClass-def and https://www.w3.org/TR/owl-ref/#equivalentProperty-def to the http file, and to a separate http-https-mappings file.ttl for the convenience of those who care. I do not want to burden the general ecosystem with the expectation that this is the right way to do things. It doesn't handle enumeration values, for example. I also do not want to see owl:sameAs used here; there is no reason to be having the debate on whether https://schema.org/Thing and http://schema.org/Thing are the self-same thing or two things with some kind of equivalence. We should show a clear lead that we'll be using https: as our canonical identifiers in the future, and provide utility dumps for those who aren't.

Issue #1931, opened 2018-06-03
Clarify Event's applicability for non-commercial and historical events


RichardWallis:
As part of an in-depth thread on the schema.org mailing list [Historical Events](https://lists.w3.org/Archives/Public/public-schemaorg/2018May/0019.html) it is clear that, although [Event](http://schema.org/Event) is equally applicable to all types of event current, historical, commercial, etc., its current description does not make this fully clear and gives the impression the that it is mainly focused on commercially oriented use cases. 

This impression is reinforced by the spread of examples, only for concerts  and an art event at a museum.

I propose that the description of Event is updated to better reflect  its broad applicability, thus:

> An event happening at a certain time past or future and location physical or virtual, such as a concert, lecture, meeting, festival, or historical event.  Repeated events may be structured as separate Event objects. Where applicable ticketing information may be added via the offers property.

In addition a further example should be added demonstrating this capability:
```
{
  "@context": "http://schema.org",
  "@type": "Event",
  "name": "Signing of the Magna Carta",
  "startDate": "1215-06-15",
  "location": {
    "@type": "Place",
    "name": "Runnymede",
    "sameAs": "https://en.wikipedia.org/wiki/Runnymede"
  },
  "contributor": {
     "@type": "Person",
     "name": "King John of England",
     "sameAs": "http://www.wikidata.org/entity/Q129308"
  },
  "sameAs": "http://www.wikidata.org/entity/Q12519",
}
```
This proposal does not address potential future proposals for new types and/or properties discussed in the thread, but would be a useful enhancement if those come to fruition or not.

Issue #503, opened 2015-05-17
Persons engaged in TheaterEvents


Harald-Effenberg:
Hi!

Currently there are no Persons allowed in TheaterEvents but attendees, organizers and performers.

But usually the perfomers in a TheaterEvent are actors - why can't we markup them as actors?
In fact, TheaterEvents would be rather boring without actors. And they would in most cases be very chaotic without a director.
Actors and directors are allowed in Movies and TVSeries - why not in TheaterEvents?

And there should be at least an itemprop "contributor" for all the other people who engage themselves for the success of a TheaterEvent: stage designers, costume designers, makeup artists, light engineers, sound engineers...

JM2C
Harry

Dataliberate:
Harry,

Much of what you are asking for is what you would expect from a CreativeWork or subtype —  I notice we have not got a stage based set of CreativeWork subtypes, which might be worth progressing in parallel to TV / Movie subtypes (Play, Show, Musical, etc.) in the way you indicate.

My interpretation of TheaterEvent is that it is mainly about the where/when aspects, with its workPerformed property identify the work(s) that are performed at the event.

I suspect that any proposals you might have for theatre and or live performance oriented enhancements would be received with interest.

~Richard

On 17 May 2015, at 16:14, Harald Effenberg <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi!

Currently there are no Persons allowed in TheaterEvents but attendees, organizers and performers.

But usually the perfomers in a TheaterEvent are actors - why can't we markup them as actors?
In fact, TheaterEvents would be rather boring without actors. And they would in most cases be very chaotic without a director.
Actors and directors are allowed in Movies and TVSeries - why not in TheaterEvents?

And there should be at least an itemprop "contributor" for all the other people who engage themselves for the success of a TheaterEvent: stage designers, costume designers, makeup artists, light engineers, sound engineers...

JM2C
Harry

Harald-Effenberg:
Hi Richard,

thank you for your answer!

I thought about the when/where-thing before I posted this issue, but on the other hand: There are perfomers, aren't they?

And often people go to a TheaterEvent because of the well-known name of a famous director - so this can be an important criterion of the Event, like the names of the performers.

And if there are performers and directors, there should be contributors, too, I think.

[...]

PR #2660, opened 2020-07-21
Updated description of Event to reference historical events.

RichardWallis:
Also added historical event example.

Implements issue #1931.

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -1138,7 +1138,7 @@ endorsement rating is particularly useful in the absence of numeric scales as it
 
 :Event a rdfs:Class ;
     rdfs:label "Event" ;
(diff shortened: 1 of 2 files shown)


danbri:
I want to take a look into this topic a bit first, if that's alright...

smrgeoinfo:
While you're editing the definition ... currently we have suggested "Event: An event happening at a certain time, past or future, and location..."  Please can we have a non-circular definition?  Is there such a thing as an event that does not happen at a certain time and location?  Basically the current definition boils down to 'an event is an event'.  

How about something like:
Event:
"Something that occurs in a certain place during a particular interval of time" (https://www.dictionary.com/browse/event);  
"an occurrence; something that happens" (https://wikidiff.com/event/process, https://www.merriam-webster.com/dictionary/event).

RichardWallis:
@danbri can we revisit this - note recent activity in #503

PR #2662, opened 2020-07-21
issue 2658: Supertype of Question should be Comment

lucy-kind:
Supertype of Question should be Comment, so as to be on the same ontological level as Answer (which is currently nested under Comment).


danbri:
LGTM

tmarshbing:
Should we also remove downvoteCount and upvoteCount from Question directly, now that it will get these properties from Comment?

PR #2663, opened 2020-07-21
issue 2636: Create eduQuestionType for questions that act as practice problems

lucy-kind:
Update issue 2636 (see discussion on issue)


data/ext/pending/issue-2636-examples.txt
```diff
@@ -0,0 +1,63 @@
...
+      "text": "This is a hint about the question."
+    },
+    "encodingFormat": "text/markdown",
+    "text": "Which model represents the product $$2 * \frac{2}{3}$$?",
+    "suggestedAnswer": [
+      {
+        "@type": "Answer",
+        "@id": "http://example.org/fractions/practice-problem#answer_2",
+        "encodingFormat": "text/html",
+        "text": "<img src=\"www.test.com/practice-problems-fractions/…image2.png\" alt=\"Answer 2\" />"
```
danbri, line 62: What is this terrifying thing? :) 

HTML inside a JSON(-LD) "text" field, itself likely already inside HTML via <script>. If the goal is to work out how to put small pieces of HTML inside Schema.org graphs, that seems a reasonable goal, but I don't think this combination of /encodingFormat and /text really work. The /text field isn't a good place for complex computer formats. Sorry if I missed this when we talked through this proposal recently.

data/ext/pending/issue-2636.ttl
```diff
@@ -0,0 +1,12 @@
...
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+
+:eduQuestionType a rdf:Property ;
+    rdfs:label "eduQuestionType" ;
+    :category "issue-2636" ;
+    :domainIncludes :Question ;
+    :isPartOf <http://pending.schema.org> ;
+    :rangeIncludes :Text ;
+    :source <https://github.com/schemaorg/schemaorg/issues/2636> ;
+    rdfs:comment "For questions that are part of learning resources (e.g. Quiz), eduQuestionType indicates the format of question being given. Example: "Multiple choice", "Open ended", "Flashcard"." .
```
danbri, line 12: The Travis-CI checks are failing on this as it is not well-formed Turtle: you're using double quotes inside double quotes. You can use """ instead:

>     rdfs:comment """For questions that are part of learning resources (e.g. Quiz), eduQuestionType indicates the format of question being given. Example: "Multiple choice", "Open ended", "Flashcard".""" .

(I checked this at https://www.w3.org/2015/03/ShExValidata/ but there are other similar tools elsewhere)

danbri: I've fixed this one directly in the proposal branch

data/ext/pending/issue-2636-examples.txt
danbri: There are various ways we could put chunks of HTML inside the schema.org graph datamodel:

 * items in the graph could have an /asHTML property pointing to an instance of the existing /WebPageElement type, and then the actual markup could be a /rawHTMLText property from that (or similar names)
 * alternatively, just have /rawHTMLText on items in the graph, like the /Answer (this would be simpler than trying to pair /encodingFormat and /text)

/cc @vholland @rvguha

philbarker: Do you need the HTML inside the graph when the JSON-LD is inside the HTML? You have a fragment id there, why not put the answer in the HTML wrapped in `<div id="answer_2"></div>` That's what we did for [K12-OCX](https://k12ocx.github.io/k12ocx-specs/inpage/structure/) (there are analagous examples at the end)

danbri: Ok, talking to @lucy-kind and colleagues, we'll retract this bit of the example for now and come up with another idiom

danbri:
ok, after discussion - example removed for now. I put a copy in https://gist.github.com/danbri/6373de64dd9820a25aba0d6dd52799b0 so the discussion makes sense to subsequent readers.

Issue #2648, opened 2020-07-13
Version 10 planning (goal: 2020-08-15)


danbri:
Schedule: Aiming to get back on a monthly schedule, so 1 month after predicted version 9.0 release.

Materials to consider for version 1.0: 

 * WebAPI (in Pending) [update draft](https://github.com/schemaorg/schemaorg/pull/2635)
 * [SkiResort should be a subclass of Resort](https://github.com/schemaorg/schemaorg/issues/2647)
 * review discussion around /Observation (see https://github.com/schemaorg/schemaorg/issues/2564)
 * review discussion around Dataset funding (and clarify definition of Dataset so it doesn't get used for everything ever)
 * collaboration with IPTC on properties for images (/cc @bquinn)
 * allow DefinedTerm on keywords to support URLs into taxonomies; #737
* a parentItem of a Comment should be Comment, not Question #851 /cc @lucy-kind 
 * Updated description of Event to reference historical events. (PR #2660)

Merged for 10:
  * a parentItem of a Comment should be Comment, not Question https://github.com/schemaorg/schemaorg/pull/2664
  * eduQuestionType for questions that act as practice problems - https://github.com/schemaorg/schemaorg/pull/2663
 * supertype of Question should be Comment https://github.com/schemaorg/schemaorg/pull/2662

danbri:
It's also time to address /conformsTo, see https://github.com/schemaorg/schemaorg/issues/1516

danbri:
+ https://github.com/schemaorg/schemaorg/pull/2669 (merged)

danbri:
* merged for this release: [learningResourceType now expected on LearningResource](https://github.com/schemaorg/schemaorg/pull/2678); thanks @philbarker /cc @lucy-kind

danbri:
Example fix: https://github.com/schemaorg/schemaorg/pull/2687

danbri:
also included https://github.com/schemaorg/schemaorg/pull/2663

danbri:
https://github.com/schemaorg/schemaorg/pull/2679

danbri:
https://github.com/schemaorg/schemaorg/issues/2690 (improves FloorPlan, currrently still in Pending, by making it more explicit which image is the ... floorplan)

danbri:
I've reviewed the various threads, and done my best to integrate into a v10 release candidate.

Things that aren't quite wrapped up yet:

* WebAPI + conformsTo; I left some review comments on the WebAPI thread
[...]

PR #2664, opened 2020-07-21
Issue 851: a parentItem of a Comment should be Comment, not Question

lucy-kind:
a parentItem of a Comment should be Comment, not Question. planned for v10 release (see issue #2648)

(edited to write " issue 2648" with a # in it so it gets automatically linked to the issue by github --danbri)


RichardWallis:
@lucy-kind  - you edited the wrong file - ```data/releases/9.0/schema.ttl``` is a release specific copy we make of the file you need to edit which is ```data/schema.ttl```.

~Richard

lucy-kind:
Hi Richard, first time editing the core file; thank you for the info! I've reverted the version file and edited the one you pointed to.

danbri:
Thanks @lucy-kind, @RichardWallis - merging! Great to fix this old bug.

Issue #2665, opened 2020-07-21
Add explanation property to Answer


lucy-kind:
Create new property of explanation for Answer type.

Description: A step-by-step or full explanation about Answer. Can outline how this Answer was achieved or contain more broad clarification or statement about it.

Takes Comment/WebContent as value.

danbri:
ping @rvguha on the naming

I am totally sold on having a property that does this (useful for Q/A and Quiz application areas), but the word "explanation" is very general. Should we consider something verbose like /answerExplanation ?

danbri:
From email:

> Guha> I would prefer the more specific term.

PR #2666, opened 2020-07-21
Create issue-2665.ttl (adds /explanation to /Answer)

lucy-kind:
#2665: add explanation Property to Answer


danbri:
Merging, after having changed the property name per discussion in #2665

PR #2686, opened 2020-08-11
Examples Processing

RichardWallis:
Added SchemaExamples module to load, process, and write Examples files and their contents.
Used in ```assign-example-ids.py``` utility script to set valid example ids where not previously assigned.


danbri:
Thanks, it is good that we treat our examples as a serious part of schema.org's content.

nickevansuk:
Just noting that adding global sequential IDs to examples adds complexity for contributors to add new examples, especially with multiple PRs open in parallel... as it's unclear which ID to use in a PR, and the "next" ID may conflict between two PRs?

If sticking with global sequential IDs, perhaps worth documenting the preferred approach when choosing IDs for new PRs?

RichardWallis:
@nickevansuk identifying which ID to use to reference an example has always been a challenge.

Reviewing and confirming new example IDs will now be part of the process of publishing future releases.

danbri:
Thanks @nickevansuk, @RichardWallis 

Yeah it was tempting to use UUIDs or something requiring less coordination, but they're painfully long. It's slightly more work integrating examples but it does force a need to look at examples carefully. Historically they have been easy to neglect, compared to the schema definitions.

PR #2688, opened 2020-08-13
Updated description of billingPeriod

### LINKED ISSUE #2210: billingPeriod of type Duration, but represents an Interval
According to the documentation fo [billingPeriod](https://schema.org/billingPeriod), it represents a time interval, such as `2014-12-21/P30D`. However, the range of billingPeriod includes Duration, which should be of the form `P[n]Y[n]M[n]DT[n]H[n]M[n]S` (such as `P30D`). The value used would be a [Time interval](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals), so really, there should be a datatype Interval which represents this. Or (less correct IMHO), Duration should include values from IS8601 Duration along with Time interval.

See also [validFor](https://schema.org/validFor).

See for examples do-invoice-examples.txt:39.

RichardWallis:
Added invoiceDate property
Updated examples to reflect changes
Fixes #2210

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -4563,7 +4563,7 @@ See also the <a href="/docs/hotels.html">dedicated document on the use of schema
     rdfs:label "billingPeriod" ;
     :domainIncludes :Invoice ;
     :rangeIncludes :Duration ;
-    rdfs:comment "The time interval used to compute the invoice." .
+    rdfs:comment "The time interval used to compute the invoice [[paymentDueDate]]." .
 
 :birthDate a rdf:Property ;
     rdfs:label "birthDate" ;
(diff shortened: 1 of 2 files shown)


danbri:
>  rdfs:comment "The time interval used to compute the invoice [[paymentDueDate]]." .

Adding a reference to paymentDueDate doesn't make a sentence

RichardWallis:
Updated wording for billingPeriod to:
> The time interval used to compute the [[paymentDueDate]] for the invoice.

dr-shorthair (commented):
Terminology is still jumbled here. 
- time-interval = an interval on the time-line with a beginning and end
- duration = distance or amount of time between two instants or temporal positions
- period = usually synonym for interval

So I would suggest that either the property name or the range needs to be changed: 
- `billingPeriodDuration` could have range `Duration`
- `billingPeriod` should have range `Interval` or `Period`, which in turn has a `hasDuration` property whose range is `Duration`
But the proposed state is not coherent.

PR #2697, opened 2020-09-05
Update version of RDF gem to 3.1.6.

gkellogg:
Fixes problem creating script/spec/schema.rb.

@Dataliberate Note that the Travis stage has never passed cleanly, but the results are ignored so as not to impede other work; this has been the case for several years now. Currently, Travis is showing of 8544 example tests, 67 fail with issues, most of which seem like legitimate markup or semantic issues to me. If a push could be made to clean these up, this stage could be allowed to fail, which would help prevent future errors from being introduced, which was the whole point of adding the linting pass.

Random examples:

* Examples issue-1758-examples.txt[255] - ArchiveComponent-457 (jsonld)
   property  schema:dateCreated
          Object "1971-1972"^^<http://schema.org/Date> not compatible with rangeIncludes (schema:Date,schema:DateTime)
* issue-1062-examples.txt[17] - HealthInsurancePlan-PreferredNetwork-NonPreferredNetwork. HealthPlanFormulary-HealthPlanCostSharingSpecification-428 (jsonld)
    property  schema:summaryUrl
          No property definition found
    ...
* Examples issue-2405-examples.txt[286] - Recommendation-414 (jsonld)
   Failed to parse input document: loading document failed: 434: unexpected token at ...

If there are some things you think the linter shouldn't be flagging, let's discuss.


RichardWallis:
Thanks for this Greg

Issue #1655, opened 2017-06-12
Action's target property should list URL as an expected type


RehanSaeed:
According to [https://schema.org/Action](https://schema.org/Action) the target property has an expected type of [https://schema.org/EntryPoint](https://schema.org/EntryPoint). However according to the [docs](https://schema.org/docs/actions.html#part-3) it can also take the form of a URL like so:

```
{
    "@context": "http://schema.org",
    "@type": "Movie",
    "name": "Footloose",
    "potentialAction" : {
    "@type": "WatchAction",
        "target" : "http://example.com/player?id=123"
    }
}
```

The docs seem correct as Google structured data uses also uses the target property as a URL type. In fact reading further, the docs also mention that the target property can take the form of an array of URL/EntryPoint types.

danbri:
I think you're right, thanks

danbri:
@RichardWallis would you mind PR'ing a fix?

RichardWallis:
Implemented in PR #2725

PR #2725, opened 2020-09-29
Added URL to range of Action > target.

RichardWallis:
Issue (#1655)


smrgeoinfo:
How come this hasn't been merged?  The examples for EntryPoint and LinkRole also have URLs as values for target.

PR #2728, opened 2020-09-30
Update schemaorg-all-https-properties.csv

bbaltutat:
Fixed missing ,"" (column) on the second item row.


RichardWallis:
Thanks @bbaltutat for your efforts in this area.
Unfortunately these Vocabulary Definition Files are created by processes at the time of a new release of the vocabulary.  To that end, your change would not be reflected on the Schema.org site.

I am therefore closing this PR. 

However, your observation has been taken into account in some work in this area and should result in correct formatting in the next release.

PR #2737, opened 2020-10-13
Remove subproperty relationship between hasCategoryCode and hasPart

RichardWallis:


whitten:
What criteria was being met by the subproperty relationship between hasCategoryCode and hasPart ?
Clearly removing that relationship reduces the connectivity between items that are currently connected.
Why do you think that connectivity is incorrect ?

PR #2739, opened 2020-10-14
Update the software readme

christianlupus:
The SOFTWARE_README.md file did contain some description how to run a local flask server. Unfortunately these are not working, the server is not starting (at least under Linux).

You get the issue output:
```
$ ./devserv.py 
Traceback (most recent call last):
  File "./devserv.py", line 15, in <module>
    from schemaversion import getVersion
ModuleNotFoundError: No module named 'schemaversion'
```

### DIFF
--- a/SOFTWARE_README.md
+++ b/SOFTWARE_README.md
@@ -23,8 +23,7 @@ Note: The associated configuration scripts are designed to run in a Linux or sim
 
 To work on the vocabulary and run locally firstly clone the repository on a local system:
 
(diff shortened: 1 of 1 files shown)


RichardWallis:
@christianlupus  Thanks for highlighting.

It looks like the problem was a typo in path insert code in ```devserv.py```.  It was targeting "Util" instead of "util".

This has be corrected in commit 333aca3.   If you can confirm that this does indeed fix it, we can then delete this PR.

christianlupus:
Yes, that was it. Now I can run locally without additional path.

Issue #2767, opened 2020-11-17
Clean unused category value from term definitions in pending.


RichardWallis:
Term definitions in the pending area include a category property.   To some extent is duplicated by the source property.
eg.:
```
:Quotation a rdfs:Class ;
    rdfs:label "Quotation" ;
    :category "issue-271" ;
    :isPartOf <http://pending.schema.org> ;
    :source <https://github.com/schemaorg/schemaorg/issues/271> ;
    rdfs:comment "A quotation. Often but not necessarily from some written work, attributable to a real world author and - if associated with a fictional character - to any fictional Person. Use [[isBasedOn]] to link to source/origin. The [[recordedIn]] property can be used to reference a Quotation from an [[Event]]." ;
    rdfs:subClassOf :CreativeWork .
```
Rationalisation in the build processes has made the category property redundant and as such should be be removed from the definitions.

danbri:
and in core files, for any that have been moved into core - thanks!

philbarker:
@RichardWallis should it be `:source` or `dct:source` ?

I don't find `schema.org/source`, and older copies of schema_all.ttl use `dct:source`

RichardWallis:
@philbarker - you are correct, I can't find it either.

The way it is applied in the ttl definition files it assumes that it should be schema.org/source.

To fix we have two options:

1.  Convert all references to dct:source
2. Add schema:source to the schemaorg meta section

@danbri - thoughts?

PR #2772, opened 2020-11-20
Clean unused category values from term definitions.

RichardWallis:
Implements issue #2767 
Mostly in pending, or terms moved from pending. 

Previously category was used for housekeeping reasons.

Also removed code that is now also redundant.


danbri:
Thanks for the cleanup. This is essentially housekeeping so let's get it into the repo ASAP (i.e. alongside v11) /cc @tmarshbing @scor

PR #2788, opened 2020-12-14
legalName subPropertyOf name

apparry:
I propose [legalName][schema:legalName] becomes a [subPropertyOf][rdfs:subPropertyOf] [name][schema:name]

```
:legalName a rdf:Property ;
    rdfs:label "legalName" ;
    :domainIncludes :Organization ;
    :rangeIncludes :Text ;
    rdfs:comment "The official name of the organization, e.g. the registered company name." ;
    rdfs:subPropertyOf :name .
```

[rdfs:subPropertyOf]: https://www.w3.org/TR/rdf-schema/#ch_subpropertyof "rdfs:subPropertyOf"

[schema:legalName]: https://schema.org/legalName "schema:legalName"
[schema:name]: https://schema.org/name "schema:name"

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -6698,7 +6698,8 @@ Typical unit code(s): MTK for square meter, FTK for square foot, or YDK for squa
     rdfs:label "legalName" ;
     :domainIncludes :Organization ;
     :rangeIncludes :Text ;
-    rdfs:comment "The official name of the organization, e.g. the registered company name." .
+    rdfs:comment "The official name of the organization, e.g. the registered company name." ;
+    rdfs:subPropertyOf :name .
 
 :leiCode a rdf:Property ;
     rdfs:label "leiCode" ;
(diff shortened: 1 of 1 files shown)


davux:
Is it common practice to have sub-properties of `name`? In principle it sounds like a nice thing to have.

Also (Noob Question Disclaimer): what does it mean in practice for a property to be a subPropertyOf another? Does it have consequences in terms of expectations during schema validation? For instance, maybe some parsers that look for `name` and find `legalName` will begin to be OK with it? (Not sure if that's a good thing or a bad thing.) Or what about parsers that find both attributes in an object and select `legalName` because it's a `name`?

alex-jansen:
This is a reasonable request but there could be unexpected consequences for schema validations as mentioned by @davux  (triples with `legalName` might show up when looking for triples with the `name` subject). This is probably not a major risk, since `identifier` (which is also a property on `Thing`) actually already has many sub-properties, including some recently introduced ones.

 More importantly, we then also need to consider making other name-related properties, such as `alternateName`, `givenName`, and `familyName`, sub properties of `name`.

Is there still interest in pursuing this pull request?

alex-jansen:
SInce this is also under discussion as part of #3062 I will close this old PR and we can continue the discussion under there.

PR #2793, opened 2020-12-31
Update Link

dhurlburtusa:
TopBraid is now TopQuadrant. Now hosted at datashapes.org.


dhurlburtusa:
Could someone who knows what used to be hosted at http://topbraid.org/schema please confirm that http://datashapes.org/schema is appropriate?

rivettp:
Yes I confirmed it via email with @HolgerKnublauch who's the main man from Top Quadrant in this space.

danbri:
Thanks for this! In fact we have some progress on this front towards integrating SHACL and ShEx exports directly into the Schema.org release process (for use in https://github.com/google/schemarama and elsewhere). I'll merge this fix but hope we have more to show soon too. 

(@RichardWallis - can you remind me where we're up to on this?)

PR #2806, opened 2021-01-13
Fix for nokogiri vulnerability 

RichardWallis:
``


gkellogg (approved):
Apparently, the relative path to the examples has changed?

gkellogg:
Also, as you may know, travis-ci.org is going away, and this whole process can probably run much better using GitHub Actions. Not sure if you’ve considered that, but I’ve moved many of my workflows over.

RichardWallis:
@gkellogg 
Relative path:  Yes this issue fell astride the directory moves in #2800 which confused things a bit
travis-ci.org: Seriously considering move to GitHub Actions - eager to learn from your experiences.

gkellogg:
GH Actions run much more quickly and share the environment created between each step. The YAML structure is quite similar, and you may benefit from published Actions. It has many more capabilities, and is worth diving into. I would be happy to help with this, but it isn't a task that I'm quite prepared to take on on my own, particularly as the Python bits aren't native to me.

Note that you can have several different workflows, not just for CI but also to help automate your release cycles and issue/PR disposition. You can also act on events other than commits and PRs, including Tag/release creation or update and many more. https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows

Issue #2811, opened 2021-01-18
Proposal to add support for structured sizes and measurements


alex-jansen:
In Release 9.0 a new [/size](https://schema.org/size) property was published (see issue #2587) to model product sizes as free-form text.

As a follow-up to address the still open items from issue #2587 and in collaboration with @danbri and @mfhepp we are now proposing to also allow a more structured way of specifying product sizes. 

For example, apparel sizes are often expressed as a combination of one or more of the following attributes:
- a size **system**, typically a country, region, or standard identifier (e.g., "eu", "iso 8559").
- a size **group** (e.g., "big and tall", "petite"), typically used together with age group and gender information to identify a fashion department.
- A size **code** (per the size system and size group), for example "3XL" or "32Wx34L".
- one or more manufacturer-defined **measurements** of the product (e.g. for pants: "waist" with a measurement of 32 inches and "inseam" with a measurement of 34 inches).
- one or more (ranges of) **measurement**(s) for the intended consumer (e.g. "waist" between 30 and 32 inches and "inseam" between 33 and 35 inches). 

We propose to add a few new types and properties to model these size-related attributes, building on prior art such as the GS1 Web Vocabulary ([SizeDetails](https://www.gs1.org/voc/SizeDetails)) and the CPI Ontology ([example](http://www.ebusiness-unibw.org/ontologies/cpi/evaluation/jeans.n3), [reference](http://www.ebusiness-unibw.org/ontologies/cpi/ns)). 

We also propose to generalize the concept of a "measurement" since it applies to many types of products (i.e., not just wearables). For example a bicycle has wheel and frame measurements, and a screw has gauge and thread-distance measurements. 

Given the size of the proposal, an external Google Docs [document](https://docs.google.com/document/d/1MbYXG8wRallckDFbqY2NYc8Y83gUGA4ABDHdMclxzQA) is created for review and inline comments.

Adding @mgh128 and @jvandriel based on the discussion in #2587.

danbri:
/cc @philarcher 

To be explicit, this is a direction that Google both supports for the general ecosystem, and expects (if adopted) to actively implement.

philarcher:
Thanks @danbri & @alex-jansen. I'll take a look and ask @mgh128 for support in that.

mgh128:
[...]

PR #2819, opened 2021-01-24
Draft for structured product sizes (issue #2811)

alex-jansen:
For issue #2811


danbri (approved):
LGTM. I will merge this now and push to webschemas.org, to aid review/feedback process. If nobody finds issues we can push it out as part of Pending in the next release. /cc @mfhepp

PR #2836, opened 2021-02-14
Implementation of GitHub Actions based continuous integration tests 

RichardWallis:
_Re Issue #2835_ 

Removal of .travis.yml Travis-CI integration file

Replaced with ci_tests.yml file in .github/workflows


RichardWallis:
Tests run in less than 5 mins - signifiant improvement over Travis-CI.org

/cc @gkellogg 
Everything Ruby appears to be running the same in this environment -could you have a conformationary look

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      
+      #Run local tests
+      - name: Run local tests
+        run: software/util/buildsite.py -r 
+
+      # Setup Ruby environment  
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+           ruby-version: 2.6
```
gkellogg, line 57: We should update this to at least Ruby 2.8, if not 3.0.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      #Install Ruby dependencies
+      - name: Install Ruby dependencies
+        run: |
+          gem update --system --conservative || (gem i "rubygems-update:~>2.7" --no-document && update_rubygems)
+          gem update bundler --conservative
```
gkellogg, line 63: This two steps are probably not necessary, any longer.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: |
+          gem update --system --conservative || (gem i "rubygems-update:~>2.7" --no-document && update_rubygems)
+          gem update bundler --conservative 
+          bundle install --gemfile=software/scripts/Gemfile
+
+      # Run Ruby tests
+      - name: Run Ruby tests
+        run: (cd software/scripts; bundle exec rake)
+        env:
+          SOFT_LINT: true
```
gkellogg, line 70: With a relatively low number of failures, we could go to a black-list model to mark those tests expected to fail, which would then be useful to note when those tests pass, and to make sure that newly introduced examples pass before being merged.

RichardWallis:
@gkellogg 
Thanks for the reviews.
If I create an issue to implement these would it be OK to assign it to you?

gkellogg:
Sure, I'll take a crack at it.

PR #2843, opened 2021-02-18
Fix example for sizes/measurements and update some enum values for consistency

alex-jansen:
fix example for sizes and measurements (update non-existing enumeration values).
Update WearableSizeTypeEnumerion values for consistency with WearableBodyTypeEnumeration


danbri (approved):
LGTM - thanks for double-checking

Issue #2862, opened 2021-04-01
Bring proposed Bioschemas terms into Pending area


RichardWallis:
This is to introduce the terms from PR #2711 "_Core Types to Support the Discovery of Life Sciences Resources_".  That PR introduced a new 'bio' area to contain term definitions.  

This issue follows the current practice of introducing the new terms into the pending area

PR #2863, opened 2021-04-01
Introduce BioSchemas terms into pending

RichardWallis:
Implements issue #2862 which in turn implements proposed bioschemas terms from PR #2711.

#2711 introduced data/ext/bio for them definitions, this moves those definitions into pending


danbri (approved):
@AlasdairGray and I met today to go through these line by line.

We're merging this, and note for future improvements:

 * linking to design discussions would be very useful, and to works e.g. Darwin Core that have inspired the choices here.
 * acknowledge that the communities brought together around bioschemas can reasonably be expected to be kept in the loop for changes, tweaks, and improvements
 * expect to move Taxon under DefinedTerm, and anticipate potential changes in that area to ensure that other non-lifesci forms of taxonomy can be addressed coherently within schema.org.
 * note that /potentialUse needs clarification and an example.
 * suggest adding /MediaObject as a potential value for /hasRepresentation, to allow mime-typed formats to be easily described.
 * Note that BioChemEntity is a kind of organizational construct rather than an upper level ontology sort of a type.
 * We'll establish quarterly check-ins to review progress, collect implementation experience, review new vocabulary suggestions especially those motivated by data-consuming implementations.

danbri:
https://webschemas.org/hasStatus is too general a property name, we should refine this. Suggestions, @AlasdairGray ?

AlasdairGray:
@danbri, in order to not delay the release, we would suggest dropping the property. 

We had a discussion about this in our telcon today. We can't find the original motivation for the `hasStatus` property and those present noted several issues with the current proposal. However, we also saw the benefit of having a property to capture this sort of information, e.g. to capture if the Gene has been 'knocked out'. 

We've created an issue in the Bioschemas tracker ([Bioschemas:512](https://github.com/BioSchemas/specifications/issues/512)) to come back with a more detailed proposal in a future iteration.

danbri:
Thanks, I'll comment it out for now!

Issue #2895, opened 2021-06-03
Add actor to PodcastSeries


gmackenz:
[actor](https://schema.org/actor) is already present almost every videogame, visual, and audio episodic and series types except for [PodcastSeries](https://schema.org/PodcastSeries). This request is based on active efforts within Google to support cast information for podcast feeds at the episode and series level. I expect this to be actively used in the near future.

_Something to ponder_ :
It seems to me we could land a lot of these commonly used properties (i.e. actor, director, etc.) on a type we are currently missing between CreativeWork/CreativeWorkSeries and the TV, Movie, Radio, Podcast, and VideoGame instance and series types

danbri:
Closing this in favour of #2899.

jhwheeler:
@danbri Is `actor` intended to be the host of the podcast?  The reason I ask is that I haven't heard podcast hosts referred to as "actors", and this is the only property in the `PodcastSeries` schema that might be able to fit the host. As far as I can see, the documentation does not specify who the `actor` property would refer to in relation to the podcast.

Much obliged.

gmackenz:
That was the intent to use the ```actor``` property to supply the person with a specific role for their appearance on the podcast. I know of at least one podcast started by one person and then later on a new person took overthe hosting role while the original creator/host (who certainly could be listed as ```author``` in that context) became a producer.

We should have some examples on podcast to represent the usage.

jhwheeler:
Makes sense, thank you for clarifying. I'll use `author` for the creator/host and `actor` for any guests who may appear in specific episodes.

PR #2896, opened 2021-06-04
Expect existing 'actor' property on existing PodcastSeries 

gmackenz:
Add actor as a property on PodcastSeries.

This is based on active efforts within Google to support cast information for podcast feeds at the episode and series level. I expect this to be actively used in the near future. #2895


danbri:
Closing - I made the change directly to core

see https://github.com/schemaorg/schemaorg/pull/2899

PR #2899, opened 2021-06-07
Update schema.ttl to add PostcastSeries as expecting 'actor' property, for #2895

### LINKED ISSUE #2895: Add actor to PodcastSeries

danbri:
Direct update to core because this is tiny, and Pending is primarily organized around additions.

Fixes #2895 
Obsoletes #2895

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -9531,6 +9531,7 @@ we define a supporting type, [[SpeakableSpecification]]  which is defined to be 
         :Event,
         :Movie,
         :MovieSeries,
+        :PodcastSeries,
         :RadioSeries,
         :TVSeries,
         :VideoGame,


danbri:
Thanks @gmackenz !

Issue #2880, opened 2021-05-20
Proposal to expand support for product return policies


alex-jansen:
**Context** - This is a proposal from Google based on our experience consuming schema.org /[MerchantReturnPolicy](https://schema.org/MerchantReturnPolicy) markup and working with similar data from online merchants. If it were accepted, it would make it easier for us and others to understand eCommerce return policies.

### **Introduction**

Schema.Org currently supports a limited subset of product return policies through type /[MerchantReturnPolicy](https://schema.org/MerchantReturnPolicy). Certain return policies that are common in the industry are not supported through this type, for example product returns to a different country than the shipping country, restocking fees, and different fees based on item condition (unopened or used). We therefore propose to add a few additional types and properties for a more comprehensive coverage of common return policies in the eCommerce industry.

### **Proposal**

We propose to expand schema.org with a few enumeration types and new properties:

*   Support for more return methods, through a new enumeration /_ReturnMethodEnumeration_, with values such as /_ReturnByMail_, /_ReturnInStore_, and /_ReturnAtKiosk_. 
*   Support for different types of return labels through a new enumeration /_ReturnLabelSourceEnumeration_, with values such as /_ReturnLabelDownloadAndPrint_, /_ReturnLabelInBox,_ /_ReturnLabelCustomerResponsibility_
*   Add additional types of return fees to existing enumeration /_ReturnFeesEnumeration_, with common values such as  /_ReturnFeesCustomerResponsibility_ and /_FreeReturn_.

We will use these new enumerations and values to add a few new properties related to return methods, fees, and labels for use on existing type /[MerchantReturnPolicy](https://schema.org/MerchantReturnPolicy):

**<span style="text-decoration:underline;">Return methods</span>**

*   /_returnMethod_ of (new) type /_ReturnMethodEnumeration_. This improves on the existing more limited Boolean /[inStoreReturnsOffered](https://schema.org/inStoreReturnsOffered).
*   /_returnPolicyCountry_ of (existing) type /[Country](https://schema.org/Country) to allow the specification of a country where the return has to be sent to.

**<span style="text-decoration:underline;">Return fees</span>**

*   /_itemDefectReturnFee_ of type /_ReturnFeeEnumeration_ to specify return fees for defect products.
    *   With corresponding property _/itemDefectReturnShippingFeeAmount_ of type /[MonetaryAmount](https://schema.org/MonetaryAmount), to be provided when /_itemDefectReturnFee_ has value /_ReturnShippingFee_
[...]

Issue #2381, opened 2019-10-29
rename pending's ProductReturnPolicy (and related) to MerchantReturnPolicy


danbri:
Based on implementor feedback we received at Google this is confusing people. The intent was that the term is essentially for merchant-level metadata, but the name is encouraging people to add it at the level of per-item Product markup.

/cc @mfhepp

RichardWallis:
Implemented in commit f0ad52c

PR #2905, opened 2021-06-15
Issue 2381: Add some types and properties for product return policies

alex-jansen:
For  #2880  (which follows #2381 etc.)


danbri (approved):
LGTM

PR #2920, opened 2021-07-09
Create Google graf

Gholamrezasadeghyan:


RichardWallis:
This is not a valid update to the vocabulary.

It is the type of data that might be published on a web site using Schema.org

Issue #2927, opened 2021-07-22
Add provider to Action


gmackenz:
This is a proposal from Google

We believe it would be very useful to associate the [provider](https://schema.org/provider) to an [Action](https://schema.org/Action) by adding it as a property. This would allow for identifying the source(s) of the data involved in a potential Action like [WatchAction](https://schema.org/WatchAction) deep links.

Proposal:
- Add proposal as a property to Action

PR #2928, opened 2021-07-22
Create issue-2927

gmackenz:
Add [provider](https://schema.org/provider) as a property to [Action](https://schema.org/Action) #2927


danbri:
To debug this, I will merge, check locally, then strip it down to the one assertion adding provider to Action.

PR #2953, opened 2021-09-07
Consistent formatting of ISSN text

### LINKED ISSUE #2952: ISSN example goes against guidance
According to http://www.bl.uk/help/get-an-isbn-or-issn-for-your-publication#:~:text=Display%20of%20the%20issn

> The number should be printed thus: ISSN 0000-0000
> it should be preceded by the initials ISSN followed by a single space

However, the examples given in https://schema.org/Periodical display it with an errant colon

> ISSN: 0140-6736

See also https://www.issn.org/understanding-the-issn/assignment-rules/the-issn-l-for-publications-on-multiple-media/

Are you happy for me to made (an admittedly pedantic) edit to https://github.com/schemaorg/schemaorg/blob/main/data/sdo-periodical-examples.txt#L28

edent:
Fixes #2952

### DIFF
--- a/data/sdo-periodical-examples.txt
+++ b/data/sdo-periodical-examples.txt
@@ -8,7 +8,7 @@ PRE-MARKUP:
  <p>Volume 376, July 2010-December 2010</p>
  <p>Published by Elsevier
  <ul>
-   <li>ISSN: 0140-6736</li>
+   <li>ISSN 0140-6736</li>
  </ul>
  <h3>Issues:</h3>
  <ul>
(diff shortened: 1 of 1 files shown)


danbri (approved):
LGTM. It feels a little like over-reach on the part of the BL to be saying how people lay out their page. If you were using a bulleted list with foo: bar colons for everything else, writing "ISSN:" instead of "ISSN" seems reasonable, for example. If the string beginning "ISSN" is intended as a computer format, that would be a different matter. Interesting that they insist on the string appearing even in non-latin script language countries, eg. Japan: https://www.ndl.go.jp/en/data/issn/index.html

All that aside, thanks for the edit @edent :) BTW I followed your example a couple weeks ago and applied for an ISSN for my own (much neglected) blog, and it arrived within a few days.

RichardWallis:
See issue #2992.

The code in the branch behind this PR needs rebasing against the main branch in the schemaorg repository.  This will enable it to pass the CI tests which it currently fails.

cc @edent

RichardWallis (approved):

Issue #2982, opened 2021-11-09
Update links to accessibility property vocabs


mattgarrish:
accessibilityAPI, accessibilityControl, accessibilityFeature, accessibilityHazard currently reference the vocabulary we were maintaining in the WebSchemas wiki, but now that we have a CG to maintain those terms we've republished them at https://www.w3.org/2021/a11y-discov-vocab/latest/

We should update the links in those properties (and add to accessMode and accessModeSufficient, which embed some terms) to point to the new vocabulary.

I'd suggest switching the current text from:

> " (WebSchemas wiki lists possible values)."

to:

> Values should be drawn from the [approved vocabulary](https://www.w3.org/2021/a11y-discov-vocab/latest/#).

with links into the appropriate lists of terms for each property.

/cc @clapierre @madeleinerothberg

madeleinerothberg:
+1

clapierre:
Agreed

mattgarrish:
Closing this issue as it's fixed with https://github.com/schemaorg/schemaorg/pull/2985

PR #2985, opened 2021-11-09
Change webschemas wiki links for accessibility properties

mattgarrish:
As mentioned in #2982, this fixes the old links so they point to the new CG document with the vocabularies for the properties.


RichardWallis:
See issue #2992.

The code in the branch behind this PR needs rebasing against the main branch in the schemaorg repository.  This will enable it to pass the CI tests which it currently fails.

cc @mattgarrish

mattgarrish:
I just updated and getting a green check now. Thanks!

PR #2996, opened 2021-11-26
Allow `Text` for `duration`-properties, fixes #3021

### LINKED ISSUE #3021: `MediaObject` does not allow `Text` for `duration`-property, but accordings to examples `Text` should be allowed
As the example in `MediaObject` states, the `duration`-property can be a
`Text` and does not have to be a `Duration`-class:

```
{
  "@type": "AudioObject",
  "contentUrl": "http://media.freesound.org/data/0/previews/719__elmomo__12oclock_girona_preview.mp3",
  "description": "Recorded on a terrace of Girona a sunday morning",
  "duration": "T0M15S",
  "encodingFormat": "audio/mpeg",
  "name": "12oclock_girona.mp3"
}
```

https://schema.org/MediaObject

Maybe I'm missing something and the schema already allows a `Text`-value, but the main reason I'm opening this PR is issue https://github.com/spatie/schema-org/issues/172.

This issue is fixed in PR #2996.

jaylinski:
As examples in `Audiobook` state, the `duration`-property can be a
`Text` and does not have to be a `Duration`-class:

```
{
  "@context": "https://schema.org/",
  "type": "Audiobook",
  "@id": "https://archive.org/id",
  "author": "Lewis Carroll",
  "duration": "PT2H31M21S"
}
```

https://schema.org/Audiobook

Maybe I'm missing something and the schema already allows a `Text`-value, but the main reason I'm opening this PR is issue https://github.com/spatie/schema-org/issues/172.

Thanks for taking your time to review this PR.

Fixes #3021.


jaylinski:
This issue is still relevant. I'm just not sure how I can get anyone from the schemaorg-team involved...

chrisspradling1980:
yeah, this PR should not be moved to inactive and thank you for standing on top of this issue and allowing it to follow through.

alex-jansen:
this looks reasonable, so merging this in

alex-jansen (approved):

rrlevering:
In Google systems at least, all the Quantity subtypes are treated as DataType subtypes.  So Duration/Mass/Energy, etc. (but practically speaking mostly Duration) are primitive strings with "special parsing" like DateTime or URL.  There is no such thing as a "@type": "Duration" object.  I was not around for the context of why that was done like that originally.

PR #3007, opened 2021-12-06
implements #2989 (products intended or suitable for adult audiences)

alex-jansen:


danbri (changes requested):
Please rephrase to say "Item" (or similar) instead of Product, to make it easier to generalize these beyond Product/Offer

Issue #2915, opened 2021-07-01
Support ISO/IEC 6523 as identifier for /Organization


alex-jansen:
The /Organization type has several identification properties, for example, /duns, /leiCode, /globalLocationNumber, /vatID and /taxID.

Many of these existing identifiers can be expressed using the [ISO/IEC 6523](https://en.wikipedia.org/wiki/ISO/IEC_6523) standard. ISO 6523 codes are composed from an International Code Designator (ICD) value, which identifies the type of the code (e.g., 0060 for DUNS, 0199 for leiCode, or 9910+9920-9957 for European /vatID), and an Organization Identifier (OI) which identifies the organization within that frame of reference.

We therefore propose to add a new property /iso6523 with values of type /Text for use on type /Organization. Alternatively we could introduce a structured type with the 2 fields ICD and OI for values of /iso6523, but since /leiCode allows values of type /Text we propose to do the same for /iso6523 for consistency.

VladimirAlexiev:
ISO/IEC 6523 defines the International Code Designator (ICD)
- Official list at http://iso6523.info/icd_list.pdf
- More usable list at https://docs.peppol.eu/poacc/billing/3.0/codelist/ICD/
- Can be resolved at http://oid-info.com/, eg http://oid-info.com/get/1.3.88
- Described at Wikidata: https://www.wikidata.org/wiki/Q28908940
- There is no separate prop on Wikidata because it's subsumed by https://www.wikidata.org/wiki/Property:P3743 ITU/ISO/IEC object ID (`oid`) with prefix `1.3`. Eg ICD=0088 is mapped to `1.3.88` 
- `urn:oid:1.3.88`
- Because ICD is an object identifier, it is used for a lot more than organizations and registered companies: also locations (GS1 GLN), goods (GS1 GTIN), banks, ATMs, application layer protocols, IP address registrants, file & document formats, information objects, local/remote procedures, etc etc.

@danbri or @alex-jansen could you capture some of the above info at https://schema.org/iso6523Code?

VladimirAlexiev:
Here are examples of Wikidata item (Wikidata property) -> ICD. I plan to add them to Wikidata
- US CAGE, NATO NCAGE supplier code: Q1115390 (P5574) -> 0141
- US DoDAAC military unit: Q5285896 -> 0144
- SE Organisationsnummer: Q10611373 (P6460) -> 0007
- FR SIRENE company number: Q3509449 (P1616) -> 0002
[...]

PR #3008, opened 2021-12-06
Implements #2915 (support for ISO6523 organization identifiers)

alex-jansen:


danbri (changes requested):
Can we call this iso6523Id or similar?

wiesmann:
I would name it iso6523Code to be consistent with leiCode. 
The attributes with the 'id' suffix are more generic (taxID, vatID). 
The 'naics' attribute has no suffix.

danbri:
Sgtm - I have merged it and will make the changes I suggested

Issue #3001, opened 2021-12-03
Add country where a MerchantReturnPolicy applies to


alex-jansen:
[MerchantReturnPolicy](https://schema.org/MerchantReturnPolicy) allows the definition of the country ([returnPolicyCountry](https://schema.org/returnPolicyCountry)) where a product must be sent to for returns. 

Merchant return policies often also differ depending on the country where the product is sold (returned from), so we propose to add an additional property /applicableCountry to allow the definition of countries where a particular return policy applies.

Alternatively we can add existing property [countriesSupported](https://schema.org/countriesSupported) to [MerchantReturnPolicy](https://schema.org/MerchantReturnPolicy) for this purpose.

PR #3009, opened 2021-12-06
Implements #3001 (support for country-specific return policies)

alex-jansen:


danbri (changes requested):
s/The Countrry/A Country/

Issue #3028, opened 2022-01-14
Add /OnlineStore as subclass of /Organization


alex-jansen:
**Context** - This is a proposal from Google based on our experience consuming schema.org [/Organization](https://schema.org/Organization) markup and working with similar data from online merchants. If it were accepted, it would make it easier for us and others to recognize which web sites are online sellers.

The /Organization type has many sub-types, but none accurately represents an online seller or store. Detecting the nature of organizations on the web, and e-commerce web sites in particular, is useful for general web site understanding. We therefore suggest to add an /OnlineStore subtype of /Organization.

WeaverStever:
It seems to me that a counterpart to localBusiness, onlineBusiness. might be a better top-level subType so that it might include other businesses that do not offer brick and mortar locations.  

More specific subTypes might someday include:

- onlineStore
- onlineLearning
- onlineConsultant
- onlineEvents
- onlinePersonnel
- onlineData
- onlineLogistics
- onlineSoftware
- onlineInfluencer

alex-jansen:
@WeaverStever Agreed, that makes sense and is indeed more consistent with localBusiness.

alex-jansen:
Thank you @RichardWallis !

PR #3030, opened 2022-01-17
Add OnlineBusiness as subtype of Organization and OnlineStore as subtype of OnlineBusiness (#3028)

alex-jansen:


alex-jansen:
Thanks for the suggestion and agree. Pushed a commit with a changed description. PTAL.

Issue #3069, opened 2022-03-16
Drug should inherit from Product


MatthiasWiesmann:
The [Drug](https://schema.org/Drug) object should inherit from [Product](https://schema.org/Product), as clearly these are _Any offered product or service_. In particular, the following Product attributes are relevant for Drugs: 

* [gtin](https://schema.org/gtin) – NDCs are mapped into GTIN space, and many countries have special GTIN ranges for drugs. 
* [releaseDate](https://schema.org/releaseDate)
* [countryOfOrigin](https://schema.org/countryOfOrigin)
* [offers](https://schema.org/offers)

The following attribute are already present in both objects:

* [manufacturer](https://schema.org/manufacturer)

RichardWallis:
Seems like a good idea

MatthiasWiesmann:
Added Pull Request #3094 that implements this. Also made DietarySupplement inherit from Product, for the same reasons.

danbri:
Tagged for next release

PR #3094, opened 2022-04-19
Make Drug and DietarySupplement inherit from Product.

MatthiasWiesmann:
See Bug #3069 
Both Drug and DietarySupplement are products, in fact DietarySupplement is described as such: _A product taken by mouth that contains a dietary ingredient intended to supplement the diet._ 

Remove the manufacturer field definition, as it is not needed anymore, that property is inherited from Product.
This also means that both `DietarySupplement` and `Drug` inherit relevant fields like `gtin`, `countryOfOrigin` or `offer`.


MatthiasWiesmann:
Gentle Ping.

Issue #3148, opened 2022-08-01
Proposal to add support for product highlights


alex-jansen:
This is a proposal from Google based on our experience consuming schema.org Product markup and working with similar data from online merchants. If it were accepted, it would allow merchants to add product highlights to their product structured data, and allow us and others to better understand products sold online.

## Description

Products advertised or sold online are often displayed not only with a title and description, but also with more detailed information, such as a list of product highlights, which are easy to consume and quick to scan ordered bullet points emphasizing the relevant attributes of a product. For example:

*   _Capture quality 24mp photos and UHD video in an easy to use point and shoot camera with a 2.8" preview screen_
*   _Excellent for underwater photography, submersible up to 16 feet_
*   _Protective plastic cover to protect from scratches_
*   _A great first DSLR camera with enough features for novice photographers of any age_
*   _Uses 2 AAA Batteries (included) and saves videos to a mini SD card (not included)_

To support ordered highlight lists we propose to introduce a new property **/highlights** of type [/ItemList](https://schema.org/ItemList) (for use on [/Product](https://schema.org/Product)). The example given above could then be represented as follows:

```
{
  "@context": "https://schema.org/",
  "@type": "Product",
  "sku": "camera-12345",
  "image": "https://www.example.com/camera.jpg",
  "name": "Nice camera",
  "description": "Compact DSLR camera with pancake lens",
  "gtin": "12345678901234",
  "brand": {
    "@type": "Brand",
    "name": "CameraBrand"
  },
  "color": "black",
  "offers": {
    "@type": "Offer",
    "url": "http://www.example.com/camera_offer",
    "itemCondition": "https://schema.org/NewCondition",
    "availability": "https://schema.org/Backorder",
    "price": "539.99",
    "priceCurrency": "EUR"
  },
  "highlights": {
    "@type": "ItemList",
    "itemListElement": [
      {
        "@type": "ListItem",
        "position": 1,
        "item": "Capture quality 24mp photos and UHD video in an easy to use point and shoot camera with a 2.8\" preview screen"
      },
      {
        "@type": "ListItem",
        "position": 2,
        "item": "Excellent for underwater photography, submersible up to 16 feet"
      },
      {
        "@type": "ListItem",
        "position": 3,
        "item": "Protective plastic cover to protect from scratches"
      },
      {
        "@type": "ListItem",
        "position": 4,
        "item": "A great first DSLR camera with enough features for novice photographers of any age"
      },
      {
        "@type": "ListItem",
[...]

PR #3149, opened 2022-08-02
add /positiveNotes for use on /Product

alex-jansen:
Implements #3148


danbri (approved):
LGTM, thanks for the discussion

danbri (approved):
Only thought is whether to update negativeNotes similarly

PR #3182, opened 2022-09-18
Introduced shex_shacl_shapes_exporter into build site process

RichardWallis:
Drops following files into releases directory:
schemaorg-res.shexj
schemaorg-res.shacl
schemaorg-subclasses.shacl
Which will download from https://schema.org/version/latest/{file}
Eventually will need referencing from the developers documentation page


danbri:
Thanks! What does “res” stand for btw?

RichardWallis:
I just added a 'schemaorg-' prefix to the filenames used in the script presuming they meant something.

Let's pick something more meaningful.

PR #3338, opened 2023-06-18
Make DonateAction a TransferAction instead of a TradeAction

trentlarson:
DonateAction is explicitly "without compensation", so I propose to make it a subclass of TransferAction instead.

All the other TradeActions involve a swap of items or an intention of each party both giving something and receiving something in return. (In fact, in broader contexts, the concept of trade is intentionally separated from gifts and donations.)

~~There are 3 properties of TradeAction that would be lost: price, priceCurrency, and priceSpecification. To include price, one might make the `object` property be a MonetaryValue or TypeAndQuantityNode -- and if there's already an `object` then that might be turned into an array. (I posit that it does not need priceSpecification.)~~ (EDIT: added these as properties in a later commit.)

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -986,7 +986,7 @@ Originally, URLs from [GoodRelations](http://purl.org/goodrelations/v1) were use
 :DonateAction a rdfs:Class ;
     rdfs:label "DonateAction" ;
     rdfs:comment "The act of providing goods, services, or money without compensation, often for philanthropic reasons." ;
-    rdfs:subClassOf :TradeAction .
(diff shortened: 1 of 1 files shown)


trentlarson:
TipAction is another possible difference; it is done in return for good service so I could see an argument for it either way. I'll add that to this PR if that makes sense.

trentlarson:
Who can I pay to look at this? :-)

MatthiasWiesmann (approved):

trentlarson:
Thanks, @MatthiasWiesmann ! 🙇

PR #3348, opened 2023-07-14
Improved unit-testing code.

MatthiasWiesmann:
* Prettier display (color) when outputing to terminal. This uses the `colorama` package which is already a requirement.
* All tests are now defined in terms of unit-tests.
* Use the unit-testing framework skip facility to indicate when a test is skipped.


RichardWallis:
@danbri On the surface no, but it needs checking out as its not just a couple lines of change.

Issue #2918, opened 2021-07-08
Normalize CreativeWork Play


WeaverStever:
Condensing the discussion on "Persons engaged in TheaterEvents #503'" as a proposal for the next release. 

**Problem**
The pending type (https://schema.org/Play) has no independent properties that you would expect to find for a play: director, actor, productionCompany, musicBy, trailer, beginDate, endDate.

**The following types have similar specific properties that are missing from Play:**
Clip
CreativeWorkSeason
Episode
MovieSeries
RadioSeries
TVSeries

The additional properties are needed so that the person offering the metadata can describe a collection of **past** events (aka a production), for inclusion in a contributing artist's resume or CV without having to supply irrelevant (or unavailable) data associated with a pending Event.

Google Rich Snippets and Wikidata appear to understand the Event (type) to be pending at some future time -- in the context of an opportunity to attend still exists.  In WikiData, a past event would be more accurately described as an occurrence.

**Proposed solution:**

Thing > CreativeWork > Play should be similar to "Thing > CreativeWork > Movie", available as a top-level type and not be required to have TheaterEvent or BroadcastEvent as a parent type (just to be able to include the actor property). 

Additionally, because the singular and the plural for "play" is often the same word, Play should be listed as a "more specific type" of CreativeWorkSeries.  

Inclusion in CreativeWorkSeries for disambiguation between a performance in a production and a collection of performances within a production.  (It is not uncommon for a production of a play to also be described as production(s) when performed or engaged in different locations. I.e. the Boston production of _CATS_)

Specific types for CreativeWorkSeries are currently:

BookSeries
MovieSeries
Periodical
PodcastSeries
RadioSeries
TVSeries
VideoGameSeries

Similar to Periodical, a separate instance of "PlaySeries" would not be required if startDate and endDate are included as specific properties within Play -- so the type covers both the singular and plural understanding of the word.

Related discussion:
**Persons engaged in TheaterEvents #503**

A very similar problem addressed and closed last month  (by adding specific properties to the type).
**Expect existing 'actor' property on existing PodcastSeries #2896**

WeaverStever:
Similar issue here #2961

thadguidry:
I've also sent feedback (3rd time now) to Google to get "Plays" added as a filter option on their Events search:

![image](https://user-images.githubusercontent.com/986438/150457870-076373ed-ffae-4bed-9dac-b3e5abe014b5.png)

WeaverStever:
[...]

PR #3352, opened 2023-07-27
Add PerformingGroup as an expected type for the property actor

gmackenz:
As discussed in #3185 cast information can contain groups (comedy, musical, etc.) as well as individual people.

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -9570,8 +9570,9 @@ we define a supporting type, [[SpeakableSpecification]]  which is defined to be 
         :VideoGame,
         :VideoGameSeries,
         :VideoObject ;
-    :rangeIncludes :Person ;
-    rdfs:comment "An actor, e.g. in TV, radio, movie, video games etc., or in an event. Actors can be associated with individual items or with a series, episode, clip." .
+    :rangeIncludes :Person,
+        :PerformingGroup ;
+    rdfs:comment "An actor (individual or a group), e.g. in TV, radio, movie, video games etc., or in an event. Actors can be associated with individual items or with a series, episode, clip." .
 
 :album a rdf:Property ;
     rdfs:label "album" ;


WeaverStever:
Seems to have missed https://schema.org/Play

See also
https://github.com/schemaorg/schemaorg/issues/2918

gmackenz:
> Seems to have missed https://schema.org/Play
> 
> See also #2918

Seems like a misinterpretation (as I see it) of [Play](https://schema.org/Play) as that is the source [CreativeWork](https://schema.org/CreativeWork) which is being performed in a specific [TheaterEvent](https://schema.org/TheaterEvent) or [BroadcastEvent](https://schema.org/BroadcastEvent). So a specific [actor](https://schema.org/actor) is linked to the event not to the play. Usually the Play is a singular conceptual instance that is performed over time in many performances.

salikh:
I believe `.actor` should also allow the mediator type https://schema.org/PerformanceRole as the value, as the JSON-LD example use already has it, but the formal description does not list it.

data/schema.ttl
```diff
@@ -9570,8 +9570,9 @@ we define a supporting type, [[SpeakableSpecification]]  which is defined to be
         :VideoGame,
         :VideoGameSeries,
         :VideoObject ;
-    :rangeIncludes :Person ;
-    rdfs:comment "An actor, e.g. in TV, radio, movie, video games etc., or in an event. Actors can be associated with individual items or with a series, episode, clip." .
+    :rangeIncludes :Person,
+        :PerformingGroup ;
```
salikh, line 9574: How about also adding :PerformanceRole here to make the JSON-LD usage example consistent with schema? (See data/examples.txt LL10845-10860)

gmackenz: Role is meant to be applicable to any property, not just actor, as per the [post announcing PerformanceRole in 2014](http://blog.schema.org/2014/06/introducing-role.html)

salikh: Oh I see! Sorry for missing that. I'll drop the pull request.

salikh: Oh I see! Sorry for missing that, and thanks for the explanation!

gmackenz:
> Where are we with this one?

I believe there has been no _actionable_ pushback on the change to allow **PerformingGroup** as an **actor**, so good to go? I suppose we could update the description a little.

alex-jansen:
From looking at the comments this is good to go, so merging in.

PR #3356, opened 2023-08-07
Fixed all inheritance warnings in schema.org

MatthiasWiesmann:
Fixed all inheritance warnings in the repository, activated the test.

* All domain or range declaration where the target inherits from another target have been removed, or if this would have resulted in an empty declaration, replaced with the common parent. 
* All the related tests are not activated (not expected to fail). In case of failure, each failure case is its own subtest with reporting. 
* Removed some globals that are not needed anymore.

### DIFF
--- a/data/ext/health-lifesci/med-health-core.ttl
+++ b/data/ext/health-lifesci/med-health-core.ttl
@@ -1203,8 +1203,7 @@
 
 :activeIngredient a rdf:Property ;
     rdfs:label "activeIngredient" ;
-    :domainIncludes :DietarySupplement,
-        :Drug,
+    :domainIncludes
         :DrugStrength,
         :Substance ;
     :isPartOf <https://health-lifesci.schema.org> ;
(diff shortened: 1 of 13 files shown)


RichardWallis:
@danbri should review this - replacing specific domain/range declarations with a more generic one may have some merit, it may also diverge from the guiding intent of the the author of the definition and make things more confusing for implementers.

MatthiasWiesmann:
Note that these replacements have no actual effect, as the the more generic type takes precedence anyway when the documentation is generated. So whatever intent was present is not surfaced anyways.

RichardWallis:
I agree that there is no ontological effect, however properties are sometimes referenced in a subtype description, in addition to the inherited generic type, specifically to hint to implementers that it is an appropriate property to use for that subtype.

gmackenz:
PR has conflicts, should you resolve them @MatthiasWiesmann or abandon?

MatthiasWiesmann:
Let's abandon that one.

gmackenz:
Abandoning then...

PR #3427, opened 2023-12-26
Issue 3420 - Clean up physician

hareesh-ms:
Issue 3420 - Clean up physician, add IndividualPhysician,PhysiciansOffice and usNPI property

• Remove the MedicalBusiness hierarchy from Physician as it can be an individual physician or a physician's office. • Add usNPI property which is linked to physician
• Create new nodes IndividualPhysician and PhysiciansOffice to differentiate the physician person and the physician office. • Add practicesAt property that denotes where an IndividualPhysician is practicing.

### DIFF
--- a/data/ext/health-lifesci/med-health-core.ttl
+++ b/data/ext/health-lifesci/med-health-core.ttl
@@ -505,9 +505,6 @@
     rdfs:comment "A process of progressive physical care and rehabilitation aimed at improving a health condition." ;
     rdfs:subClassOf :MedicalTherapy .
 
-:Physician a rdfs:Class ;
-    rdfs:subClassOf :MedicalBusiness .
-
 :PreventionIndication a rdfs:Class ;
     rdfs:label "PreventionIndication" ;
     :isPartOf <https://health-lifesci.schema.org> ;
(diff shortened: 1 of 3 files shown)


jvandriel:
Could we please finish the discussion @3420 first? A pull request seems a bit too early given that the discussion hasn't even finished yet.

jvandriel:
First off, I'm happy to hear your feeling well enough to get behind a computer again. I hope your recovery won't take too long.

As for the continuing the discussion. One of the reasons we came up with turning `Physician` (and others) into a subtype of `Occupation` (`Profession`) is because of the `occupationalCategory` property. 

Both [BLS O*NET-SOC](http://www.onetcenter.org/taxonomy.html) and [ISCO-08](https://www.ilo.org/public/english/bureau/stat/isco/isco08/) have quite large lists containing different types of medical professionals (including several types of physicians). And given that these describe them as professions we felt that having them be other types (`LocalBusiness` and 'Person') would not resolve the long standing issues.

Making `Physician` a subtype of `Occupation` means we don't have to choose any winner between either yet does help with aligning it to other standards - which the vocabulary also refers to.

Issue #3230, opened 2022-12-08
Add a mechanism to support product certifications


MatthiasWiesmann:
Many objects can bear some form of certification, Organizations can be ISO certified, Food products can be Organic or Vegan, a person can be a certified professional. There are certifications for many domains: regulatory, organizational, recycling, food, efficiency, educational, ecological, etc. 

Certifications can be first party (the certification authority makes the claim), second-party (an authorized party makes the claim), or self-certification (recipient of the certification makes the claim). Some certifications have a unique identifier, for instance the FSC (Forest Stewardship Council). 

Some schema.org objects have support for some specific certifications, for instance `Product` has `hasEnergyConsumptionDetails`, `Person` and `Organization` have `hasCredential`, but these are very specific. Adding specific attributes for all possible certification is not feasible. 

A certification would be an object with the following properties:

* A certification unique identifier
* A certification authority
* A certifying organization (for second party)
* A certification measurement (energy efficiency, CO² emissions). 
* The object of the certification (could be `about`). 

Certification could descend from `CreativeWork`

danbri:
Thanks Matthias!

Do you have a sense for how this application area relates to W3C Verifiable
Credentials?

danbri:
I discussed this with @alex-jansen and @MatthiasWiesmann 

Context - This is a proposal from Google based on our experience consuming schema.org [Product](https://schema.org/Product) and [Organization](https://schema.org/Organization) markup and working with similar data from online merchants. If it were accepted, it would make it easier for us and others to understand certifications of products and organisations. 

Another overview:

> Schema.org currently only supports a limited and specific set of product certifications, like energy efficiency labels (see [#2670](https://github.com/schemaorg/schemaorg/issues/2670)). There are many other types of certifications that can be attached to products, but also organisations. Certifications can be legal (FCC mark), administrative (ISO 9001), ethical (organic food, carbon neutral), or religious (kosher or halal food). Since consumers are increasingly interested in knowing the certifications of products and merchants, we believe that adding a small amount of vocabulary to allow a structured way of expressing certification characteristics would benefit the ecosystem. Note that this proposal only a mechanism to describe certifications, not to verify them.

Matthias prepared a doc last year which I think is best to just share. I would like to see some attention from both Verifiable Credentials perspective, digital certification in general, and also (at least in terms of terminology) Education/skills, e.g. various terms relating to [credentials](https://schema.org/docs/search_results.html?q=credential).

[...]

PR #3438, opened 2024-01-05
Issue #3230: Add Certification class

alex-jansen:
Add Certification class per #3230


alex-jansen:
oops, too many commits in this pull request

PR #3441, opened 2024-01-05
Add Certification class (#3230)

alex-jansen:
#3230


alex-jansen:
2 errors fixed (logo and example numbering), but the schemas:name issue is
puzzling...

alex-jansen:
replacing "name" with "schema:name" in the problematic examples fixes the
errors...

Issue #610, opened 2015-06-19
acceptedPaymentMethod should apply to businesses as well as Demands and Offers


tommorris:
The `acceptedPaymentMethod` property currently only applies to Demands and Offers. The most common use case for acceptedPaymentMethod is likely to be to specify the payment methods accepted _at a particular place of business_ rather than for a specific demand or offer.

This type of data is displayed on a lot of websites and is often used by people to determine whether they purchase from a given location or not, and as part of the practicalities of travel and living.

I've started documenting sites that publish this kind of information over on the microformats wiki: see [payment-method-examples](http://microformats.org/wiki/payment-method-examples).

danbri:
Oh, this one fell through the cracks. Sorry! Noticed thanks to @RichardWallis having [fixed up](https://github.com/pocketworks/git2pdf/issues/4#issuecomment-237328803) a "make a PDF of your 100s of open issues" tool, and the red bug color jumped off the page.

I suggest adding Organization and changing the textual definition from "The payment method(s) accepted by seller for this offer." to something like "The payment method(s) that are accepted in general by an organization, or for some specific demand or offer."

/cc @mfhepp @vholland @shankarnat @scor

alex-jansen:
Reopening this one @MatthiasWiesmann

MatthiasWiesmann:
Pull request: https://github.com/schemaorg/schemaorg/pull/3447

gmackenz:
This has been resolved I believe.

PR #3447, opened 2024-01-12
acceptedPaymentMethod can now be attached to Organization.

MatthiasWiesmann:
acceptedPaymentMethod can now be attached to Organization
This should fix issue #610

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -3911,10 +3911,10 @@ See also the <a href="/docs/hotels.html">dedicated document on the use of schema
 :acceptedPaymentMethod a rdf:Property ;
     rdfs:label "acceptedPaymentMethod" ;
     :domainIncludes :Demand,
-        :Offer ;
+        :Offer, :Organization ;
     :rangeIncludes :LoanOrCredit,
         :PaymentMethod ;
-    rdfs:comment "The payment method(s) accepted by seller for this offer." .
+    rdfs:comment "The payment method(s) that are accepted in general by an organization, or for some specific demand or offer." .
 
 :acceptsReservations a rdf:Property ;
     rdfs:label "acceptsReservations" ;


mfhepp:
I see the pragmatic benefit of this, as some entities might want to express that they "in general accept credit cards" or similar. And as for data markup at Web scale, we must make trade-offs and accept that the meaning of any such data can only be interpreted in context and with a limited degree of certainty (no "true, logical assertions"). 

Still, very few businesses will accept methods of payment independent of

- transaction volume,
- types of goods,
- the actual point of sale or service location,

etc.

Two minor improvements worth consideration:

1. Maybe attach it at the [LocalBusiness](https://schema.org/LocalBusiness) rather than [Organization](https://schema.org/Organization)? 
2. The wording could clarify the difference in meaning depending on the scenario, like so

*"When used with Organization (or LocalBusiness), this indicates that the method of payment is in general accepted at the business location. When used with Offer or Demand, it indicates that the method of payment is available for the particular business transaction."*

[Thing](https://schema.org/Thing) > > [LocalBusiness](https://schema.org/LocalBusiness)
[Thing](https://schema.org/Thing) > [Place](https://schema.org/Place) > [LocalBusiness](https://schema.org/LocalBusiness)

Tiggerito:
I'm sure this is of value to OnlineBusiness as well, so I think it is better in Organization

MatthiasWiesmann:
As mentioned by [Tiggerito](https://github.com/Tiggerito), this would also be relevant for online stores. 

I don't think the conditions you mention for an organization are more relevant than on the current objects, like offers. 

For online offers, even if a given offer is marked as accepting credit card transaction, the seller might still decline, for instance because they don't trust a particular card provider, the card does not implement some security features (3D secure), the customers is from an untrusted region, because the volume is too high (say you added 1 million items to the cart). 

The same contextual interpretation applies in all cases…

PR #3469, opened 2024-02-20
Fix typo in claimInterpreter in issue-2450.ttl

compulim:


MatthiasWiesmann (approved):

compulim:
I think there are some more words are missing. Tried to look through Git history but no luck.

Credit to Bing Copilot, I tried to recreate those missing words.

I think there are some inconsistencies across some properties, but we can fix it in another PR or issue.

One thing I spotted is `claimInterpreter` is the entity refining solely the `MediaObject`. But if we look at the `interpretedAsClaim` property closely, it said the `claimInterpreter` is the entity for refining **both** the `MediaObject` and `CreativeWork`.

MatthiasWiesmann (approved):

MatthiasWiesmann (approved):

PR #3474, opened 2024-03-01
Add codespell: config + workflow (to catch new typos) + fix a number of typos using codespell

yarikoptic:
Initially encountered typos in another representation of past version of linkml elsewhere, decided to fix them "in the root" ;)   more on codespell: https://github.com/codespell-project/codespell . Resulted in fixing some code including 1 interface but IMHO was ok.


data/ext/health-lifesci/med-health-core.ttl
MatthiasWiesmann, line 1: I don't think this fix is correct:

- reoccurrence = a further or repeated occurrence. 
- recurrence = the fact of occurring again.

MatthiasWiesmann (changes requested):

data/ext/health-lifesci/med-health-core.ttl
yarikoptic: thanks for spotting!

Filed
- https://github.com/codespell-project/codespell/pull/3378
to address "upstream". 

Redid fixing, force-pushed.

MatthiasWiesmann (approved):

PR #3503, opened 2024-04-03
Added digital link attribute to Thing

MatthiasWiesmann:
Adding digital link attribute to Thing. 
This attribute was discussed on [bug 3475(]https://github.com/schemaorg/schemaorg/issues/3475).
A [demo version of the repository is available](https://digitallink-dot-schemadotorg1.ew.r.appspot.com/Product)


alex-jansen:
#3475 provides an example of a digital link, but it would be nice to add a full Product example eventually as well.

alex-jansen:
Seeing no more comments on #3475 so it looks we have agreement. I will move forward with merging this in.

PR #3546, opened 2024-06-11
Extend 'MerchantReturnPolicySeasonalOverride' with additional attribu…

### LINKED ISSUE #3542: Extend 'MerchantReturnPolicySeasonalOverride' to enable overrides of more properties
The current [MerchantReturnPolicySeasonalOverride](https://schema.org/MerchantReturnPolicySeasonalOverride) is quite limited in the set of attributes that it can override on [MerchantReturnPolicy](https://schema.org/MerchantReturnPolicy). For example, having free/no returns, store credits as refund or keeping the product as return method would be useful to be able to override during holiday/sale events.

On the extreme, the seasonal override *could* itself be an `MerchantReturnPolicy` with a start and end date, but simply extending the seasonal override with a few properties should be sufficient for most use cases. I propose extending with the following properties:

* `/refundType` - https://schema.org/RefundTypeEnumeration
* `/returnMethod` - https://schema.org/ReturnMethodEnumeration
* `/returnFees` - https://schema.org/ReturnFeesEnumeration

`/returnFees` then implies also adding:

* `/restockingFee` - https://schema.org/MonetaryAmount or https://schema.org/Number
* `/returnShippingFeesAmount` - https://schema.org/MonetaryAmount 

depending on the selected return fee.

jlndr:
…tes.

Closes #3542 

This adds:

* `/refundType`
* `/restockingFee`
* `/returnFees`
* `/returnMethod`
* `/returnShippingFeesAmount`

to properties which can be overridden by the seasonal override.

Also update .gitignore with common python enviroment ignores from https://github.com/github/gitignore/blob/main/Python.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,12 @@
 /rspec-failures.txt
 software/site/*
 /tmp
+
+# Environments (from https://github.com/github/gitignore/blob/main/Python.gitignore)
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
--- a/data/ext/pending/issue-2381.ttl
+++ b/data/ext/pending/issue-2381.ttl
@@ -131,7 +131,7 @@
 
 :refundType a rdf:Property ;
     rdfs:label "refundType" ;
-    :domainIncludes :MerchantReturnPolicy ;
+    :domainIncludes :MerchantReturnPolicy, :MerchantReturnPolicySeasonalOverride ;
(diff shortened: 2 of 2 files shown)


alex-jansen:
can you build a test site for review?

jlndr:
Here is a test site: https://1-dot-schemadotorg-expr-jlinder.ew.r.appspot.com/MerchantReturnPolicySeasonalOverride

alex-jansen (approved):

PR #3561, opened 2024-07-03
Update releases.html

hartca:
Removed duplicate 27.01 2024-06-24 release info.


hartca:
Screenshot 2024-07-04 01 01 13

PR #3593, opened 2024-08-26
Jsonld context ld

MatthiasWiesmann:
Added unit-test to validate that the `sdojsonldcontext` library produces valid JSON.
The JSON is currently produced by concatenating strings in Python, so there is no guarantee to be valid.


alex-jansen:
codespell error is not relevant, it should not check Python syntax.

alex-jansen (approved):
codespell error is irrelevant

Issue #3537, opened 2024-05-22
Clean-up PaymentMethod


MatthiasWiesmann:
The list of enum values for [PaymentMethod](https://schema.org/PaymentMethod) points to the [goodrelations web-site](http://www.heppnetz.de/ontologies/goodrelations/) which seems to have been updated last in 2008, the last messages on its mailing lists were in 2016. 

The list of payments is incomplete, it is missing big players like ApplePay while including dead projects like PaySwarm. 

I don't think that maintaining a complete list of online payment lists would scale. Instead I would suggest adding a `PaymentMethod` class between [`PaymentCard`](https://schema.org/PaymentCard) and [`FinancialProduct`](https://schema.org/FinancialProduct), which would include additional fields like:

* Currency
* Geographical scope

For insance 
```json 
{
  "@context": "https://schema.org",
  "@type": "PaymentMethod",
  "name": "Twint",
  "currency": "CHF",
  "provider": {
    "@type": "Organization",
    "name": "TWINT AG",
    "url": "https://www.twint.ch/"
   }
}
```
I would also update PaymentCard with a property to reflect security features like 3D secure.

MatthiasWiesmann:
The following methods of payment also need to be added:

* Carrier payments
* Pay in convenience stores (very common is Asia)

rrlevering:
The problem with this change is that it's now written that the existing goodrelation enums are valid but that's not true if you make PaymentMethod structured.  Enums are essentially modeled as primitive values in schema.org (strings or urls) in practice and if you make the type a non-enum then it cannot support raw strings.  So all the existing schema.org online is now invalid.  Maybe consider adding Text to the range of acceptedPaymethodMethod with this list added there as supported legacy values and mention that you can represent the same thing (the goodrelation URLs) with a suggested structured value?

MatthiasWiesmann:
Pull request #2996 should fix this.

jvandriel:
If you don't mind me asking, why is this change needed? 

I was always under the assumption schema.org's data model specifies that every value can be a `Text` or `URL` value, even if it hasn't been specified.

MatthiasWiesmann:
> I was always under the assumption schema.org's data model specifies that every value can be a `Text` or `URL` value, even if it hasn't been specified in vocabulary definitions.

Technically, this is not the case, although my parsers apply Postel's principle and accept way more than the schema specifies.

rrlevering:
[...]

PR #3606, opened 2024-09-02
Cleanup of the PaymentMethod structure

MatthiasWiesmann:
See Issue #3537 this includes the following changes:

* PaymentMethod is now a proper class.
* Added a PaymentMethodType enumeration with the actual payment method types, including things like carrier-payment or in store payment which are common in Asia and not present in the good relation ontology. 
* Marked the old `http://purl.org/goodrelations/v1#` enumeration values as deprecated, the website behind these URLs has not been updated since 2011, it is missing many payment methods and mixes payment method _type_ and payment method _providers_. 
* Added examples of payment methods attached to an online store. 

A test instance can be viewed here: https://payment-methods-dot-schemadotorg1.ew.r.appspot.com/PaymentMethod

### DIFF
--- a/data/ext/pending/issue-3537-examples.txt
+++ b/data/ext/pending/issue-3537-examples.txt
@@ -1,0 +1,43 @@
+TYPES: #eg-3537 PaymentMethod,OnlineStore,PaymentService
+
+PRE-MARKUP:
+
+Possible Payment: Cash On Delivery, SplashPay 🌊, SEPA bank transfer and prepayment in all Splash stores 🌊.
+
+MICRODATA:
+
+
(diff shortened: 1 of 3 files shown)


Tiggerito:
The JSON-LD example in your test instance is not valid. There should be just one "acceptedPaymentMethod" containing an array of the entries.

Some entries are of type "PaymentService", which does not seem right.

MatthiasWiesmann:
I fixed the JSON examples. 
The PaymentService entries are intended, added some descriptions to make it clearer.
PTAL.

alex-jansen (approved):

Issue #3488, opened 2024-03-18
In `duration`, `s/date format/duration format/`


hoehrmann:
https://schema.org/duration says »in [ISO 8601 date format](http://en.wikipedia.org/wiki/ISO_8601)« but surely means the duration format; schema.org properly refers to the duration format in many other places.

PR #3608, opened 2024-09-03
Fixed bug where duration was described in terms of time 

MatthiasWiesmann:
See bug #3488 duration was described in terms of ISO time, and not duration.


MatthiasWiesmann:
Note that the changes are the Certification are just cause by whitespace / editing.

alex-jansen (approved):

PR #3625, opened 2024-09-16
Clean-up assign-example ids.

MatthiasWiesmann:
Clean up the assign-example-ids script so it can be called directly in python.

* move all the logic into a function.
* use logging systematically
* use a dictionary structure to avoid iterating twice over all examples
* only import the library, not the classes.
* raise an error in the case where two examples have the same file, instead of overwriting.

### DIFF
--- a/software/SchemaExamples/utils/assign-example-ids.py
+++ b/software/SchemaExamples/utils/assign-example-ids.py
@@ -1,60 +1,49 @@
 #!/usr/bin/env python3
 # -*- coding: UTF-8 -*-
 
-import sys
 import os
-for path in [os.getcwd(),"./SchemaExamples","./software/SchemaExamples"]:
-  sys.path.insert( 1, path ) #Pickup libs from shipped lib directory
-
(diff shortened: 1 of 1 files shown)


jlndr (approved):

software/SchemaExamples/utils/assign-example-ids.py
```diff
@@ -1,60 +1,49 @@
...
-        log.info("Writing %s" % fn)
-        f = open(fn,"w")
-    f.write(ex.serialize())
-    f.write("\n")
-if f:
-    f.close()
-sys.exit(0)
+
+for path in [os.getcwd(),'./SchemaExamples','./software/SchemaExamples']:
+    sys.path.insert(1, path) #Pickup libs from shipped lib directory
```
pascalfleury, line 9: #Pickup -> # Pickup

MatthiasWiesmann: Done.

PR #3646, opened 2024-09-27
Fix imports and breakage

MatthiasWiesmann:
Fix the various imports and the breakage introduced by the changes in the script directory.

* All imports are now done relative from the root.
* All imports are done at the start of the module, so we get errors immediately. 
* Fixed various breakage.
* Only module/libraries are imported.
* Moved pretty logging into its own file.
* Added a context managed to log entry/exit of a particular task.
* Uncommented some logging statements. 
* Ran the `ruff` formatter on all files.
* Removed superfluous log initialisation statements.

### DIFF
--- a/software/SchemaExamples/schemaexamples.py
+++ b/software/SchemaExamples/schemaexamples.py
@@ -9,30 +9,39 @@ import re
 import threading
 
 IDPREFIX = "eg-"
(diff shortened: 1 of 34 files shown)


pascalfleury (approved):
I would have welcomed a review for just the formatting changes before the fix.

PR #3657, opened 2024-10-01
Add tooling to annotate a part of the schema with 'isPartOf'

pascalfleury:
This addition in tooling enables to manage the 'isPartOf' mechanically, so avoid errors when moving schema parts from eg. pending to general availability.


software/util/reorg.py
```diff
@@ -6,67 +6,104 @@
 """
 
 from collections.abc import Sequence
-from rdflib import Graph, Namespace
+from rdflib import Graph, Namespace, RDF, RDFS, DCTERMS, URIRef
```
MatthiasWiesmann, line 9: See the Python style guide, only import packages, not classes.

software/util/reorg.py
```diff
@@ -6,67 +6,104 @@
...
+import argparse
+import itertools
 import logging
 import sys
-import argparse
 
-def SchemaOrgGraph():
-  """Creates a graph with usual namespaces set as we want in schemaorg."""
-  g = Graph()
-  g.bind("dc", Namespace("http://purl.org/dc/terms/"), replace=True)
```
MatthiasWiesmann, line 17: Maybe add a TODO to have these in a central place.

software/util/reorg.py
```diff
@@ -6,67 +6,104 @@
...
-  return g
+class SchemaOrgGraph(object):
+  def __init__(self, filename = None, format = "turtle"):
```
MatthiasWiesmann, line 16: Consider adding type signatures.

software/util/reorg.py
```diff
@@ -6,67 +6,104 @@
...
+    return getattr(self.g, *args, **kwargs);
+
+  def IdenticalTo(self, other):
+    only_in_other = other.g - self.g
+    only_in_self = self.g - other.g
+    if len(only_in_other) or len(only_in_self):
+      raise f"Graphs differ: {only_in_other + only_in_self}"
+    return True
+
+  def FullyContains(self, graph):
```
MatthiasWiesmann, line 35: Weird function signature, either returns True or an exception. Do  we need the True?

software/util/reorg.py
```diff
@@ -6,67 +6,104 @@
...
+    only_in_subset = graph - self.g
+    if len(only_in_subset):
+      raise f"Graph does not contain it all: {only_in_subset}"
```
MatthiasWiesmann, line 38: Please raise an Exception subclass.

software/util/reorg.py
```diff
@@ -6,67 +6,104 @@
...
+    return True
 
-def GraphsAreIdentical(g1, g2):
-  only_in_g1 = g1 - g2
-  only_in_g2 = g2 - g1
-  # TODO: print sampled diffs if any.
-  return len(only_in_g1) == 0 and len(only_in_g2) == 0
+  def ListSubjects(self, subject_type):
+    return set([s for s, p, o in self.g.triples((None, RDF.type, subject_type))])
```
MatthiasWiesmann, line 42: Nitpick, return frozenset.

software/util/reorg.py
pascalfleury: Done

software/util/reorg.py
pascalfleury: Added type annotations to this class.

software/util/reorg.py
pascalfleury: The goal is to use it in constructs like:

```
if my_graph.FullyContains(other_graph):
  # do something
```

and still get a sensible error message when that fails.

software/util/reorg.py
pascalfleury: Done.

software/util/reorg.py
pascalfleury: done.

software/util/reorg.py
pascalfleury: Done.

pascalfleury:
git mess. dropping this.

PR #3684, opened 2024-10-11
Updated the `sdojsonldcontext` module to Json package to generate Json

MatthiasWiesmann:
* Migrated the `sdojsonldcontext` to use the standard python json package to generate json
* Fixed bug where the code would not handle multiple types correctly (not present in actual schema)
* Fixed bug where the code would not output correct Json for an empty set of terms (not present in the actual schema)
* Added back logic to produce individual files
* Updated unit-tests, removed one bogus one.
* Moved file unit-tests into its own file.
* Use block logging

### DIFF
--- a/software/util/sdojsonldcontext.py
+++ b/software/util/sdojsonldcontext.py
@@ -7,6 +7,7 @@ import json
 import logging
 import os
 import sys
+import typing
 
 # Import schema.org libraries
 if not os.getcwd() in sys.path:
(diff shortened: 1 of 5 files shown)


software/util/sdojsonldcontext.py
```diff
@@ -16,99 +15,68 @@
...
+    global CONTEXT
+    if not CONTEXT:
+        CONTEXT = createcontext()
+    return CONTEXT
 
-    """
-    log.info("Creating JSON-LD context")
 
-    SCHEMAURI = "http://schema.org/"
+def _convertTypes(type_range):
```
pascalfleury, line 34: can you add a type hint to `type_range` it would help understand this function.

pascalfleury (approved):

software/util/sdojsonldcontext.py
MatthiasWiesmann: Done

pascalfleury (approved):

Issue #3673, opened 2024-10-08
Make Casino Inherit from Organization


MatthiasWiesmann:
See discussion #3449

MatthiasWiesmann:
`Casino` does not inherit from [LocalBusiness](https://schema.org/LocalBusiness), instead it inherits from [EntertainmentBusiness](https://schema.org/EntertainmentBusiness) which inherits from `LocalBusiness`. 

I don't think `EntertainmentBusiness` needs to be a local business. 
The question is, how about its direct descendants:

* [AdultEntertainment](https://schema.org/AdultEntertainment) should probably be just an organisation (online version exists). 
* [AmusementPark](https://schema.org/AmusementPark) should probably be a LocalOrganization
* [ArtGallery](https://schema.org/ArtGallery) should probably be just an organization (online version exists).
* [ComedyClub](https://schema.org/ComedyClub) probably local
* [MovieTheater](https://schema.org/MovieTheater) local
* [NightClub](https://schema.org/NightClub)

MatthiasWiesmann:
I have set-up a test instance with these changes: https://casino-dot-schemadotorg1.ew.r.appspot.com

alex-jansen:
(added this same comment to the PR) 

A problem I see, for example for Casino, is that moving EntertainmentBusiness under Organization means that it is going to lose all the properties from Place and LocalBusiness. However, some large well-known casino's for example in Las Vegas such as the Wynn and Caesar's Palace use this markup and provide properties from Place, for example the address property. I therefore don't think we can do this as designed. Would it be an option instead to add a new OnlineCasino class under OnlineBusiness?

alex-jansen:
We will just add OnlineCasino as a subtype of Organization.

PR #3689, opened 2024-10-14
Decouple EntertainmentBusiness from LocalBusiness

MatthiasWiesmann:
See bug #3673 

Basically, AdultEntertainment, ArtGallery, and Casino are just Organizations.

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -128,7 +128,7 @@ Should not be used where the nature of the alignment can be described using a si
 :AmusementPark a rdfs:Class ;
     rdfs:label "AmusementPark" ;
     rdfs:comment "An amusement park." ;
-    rdfs:subClassOf :EntertainmentBusiness .
+    rdfs:subClassOf :EntertainmentBusiness, :LocalBusiness .
 
 :AnimalShelter a rdfs:Class ;
     rdfs:label "AnimalShelter" ;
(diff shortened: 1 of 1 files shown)


alex-jansen (commented):
A problem I see, for example for Casino, is that moving EntertainmentBusiness under Organization means that it is going to lose all the properties from Place and LocalBusiness. However, some large well-known casino's for example in Las Vegas such as the Wynn and Caesar's Palace use this markup and provide properties from Place, for example the address property. I therefore don't think we can do this as designed. Would it be an option instead to add a new OnlineCasino class under OnlineBusiness?

gmackenz:
> AdultEntertainment, ArtGallery, and Casino

I disagree that all of these are not also LocalBusiness--especially ArtGallery which still mostly exist more as a physical places than as an online business. To a lesser degree the same is true for AdultEntertainment and Casino. I would prefer multityping myself over creating OnlineCasino, OnlineArtGallery, etc.

alex-jansen:
Instead we will just add OnlineCasino as a subtype of Organization.

alex-jansen:
Reopening for further discussion. Given the comments from @mfhepp and @gmackenz and the current complex subtyping structure under Organization it seems multi-typing is probably better in this case?

gmackenz (approved):
Approval

PR #4292, opened 2024-11-04
Clean-up the term accumulation logic

MatthiasWiesmann:
The SdoTerm source class has some logic to accumulate term informations from queries. 
Rename tmpTerm to TermAccumulator to reflect what the class actually does.
Simplified the accumulation code.


pascalfleury (approved):
LGTM.

Issue #3572, opened 2024-07-24
New Schema to Represent Energy and Transportation Incentives


dswen:
**Motivation**
Governments, utilities, and other organizations offer financial incentives to help make efficient appliances and vehicles more affordable for consumers. These incentives save people money, but can be difficult to find and understand. Information is fragmented across multiple sources and is often formatted inconsistently, sometimes in ways that are hard to understand by typical consumers. This can limit awareness and adoption.

**Description**
We propose a new type, SustainabilityFinancialIncentive, which contains information about financial incentives related to electrification, carbon emissions reduction, or other sustainability initiatives. This schema aims to balance ease of adoption while still providing consumers with the highest priority information they need to understand incentives.

Our goals in this effort are to develop schema guidance that:
*   Is easy for incentive providers to adopt
*   Applies globally as well as domestically in the United States
*   Provides unique incremental value over existing data sets
*   Will generate data that can be shared broadly for public use (e.g. via DataCommons.org)

We do not want to:
*   Develop competing standards 
*   Design a schema that fits every possible use case
*   Design a schema that only works for specific domains (e.g. only solar incentives)

The core use case we have in mind is providing basic information about available incentives. In particular, we want to answer these questions:
*   What general category of purchase (e.g. solar panels or heat pumps) is being incentivized?
*   How much money can I access to help defray the cost of purchase and installation? We can approximate that with:
    *   How much money do people typically receive?
    *   What's the maximum amount of money possible?
*   Is it available in my area?
*   Is the incentive still active or has it expired or lost funding?
*   Where can I learn more about this incentive?

Here are the proposed fields:

## FinancialIncentive

| Field | Data type | Description |
| :---- | :---- | :---- |
| display\_name | string | A human-readable display name for this incentive  e.g. "Faster Adoption and Manufacturing of (Hybrid and) Electric Vehicles (FAME-II)" |
| eligible\_region | [DefinedRegion](https://schema.org/DefinedRegion) | The region in which the incentive is applicable. |
| value | [IncentiveValue](\#incentivevalue) | The amount of money the consumer can receive. |
| status | enum | Whether this incentive is currently active. ACTIVE | ON\_HOLD | IN\_DEVELOPMENT | RETIRED |
| provider | [Provider](http://schema.org/provider) | The entity / organization that is providing this incentive.  |
| url | string | A URL where people can find more information about this incentive, including a description of what it is, eligibility criteria, and how it can be redeemed. |
[...]

PR #4322, opened 2024-12-20
Draft for financial incentives (issue #3572)

ArtixBot:
For issue #3572. The schema itself has had a few changes since the last comment on there; assume that the schema presented here is the latest version. (The changes are minor.)

### DIFF
--- a/data/ext/pending/issue-3572.ttl
+++ b/data/ext/pending/issue-3572.ttl
@@ -315,6 +315,7 @@ Use Text for incentives that are limited based on other criteria, for example if
 
 # Additions to existing types.
 :provider :domainIncludes :FinancialIncentive.
+:publisher :domainIncludes :FinancialIncentive.
 
 :areaServed :domainIncludes :FinancialIncentive.
 
--- a/data/ext/pending/issue-3572.ttl
+++ b/data/ext/pending/issue-3572.ttl
@@ -1,0 +1,231 @@
+@prefix : <https://schema.org/> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix xml: <http://www.w3.org/XML/1998/namespace> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+# FinancialIncentive, the main class for all financial incentives.
+:FinancialIncentive a rdfs:Class ;
+    rdfs:label "FinancialIncentive" ;
+    :isPartOf <https://pending.schema.org> ;
+    :source <https://github.com/schemaorg/schemaorg/issues/3572> ;
+    rdfs:comment "Represents financial incentives for goods/services offered by an organization (or individual)." ;
+    rdfs:subClassOf :Intangible .
+
+:incentiveType a rdf:Property ;
+    rdfs:label "incentiveType" ;
+    :domainIncludes :FinancialIncentive ;
+    :rangeIncludes :IncentiveType ;
+    :isPartOf <https://pending.schema.org> ;
+    :source <https://github.com/schemaorg/schemaorg/issues/3572> ;
+    rdfs:comment "The type of incentive offered." .
+
+:incentiveStatus a rdf:Property ;
+    rdfs:label "incentiveStatus" ;
+    :domainIncludes :FinancialIncentive ;
+    :rangeIncludes :IncentiveStatus ;
+    :isPartOf <https://pending.schema.org> ;
+    :source <https://github.com/schemaorg/schemaorg/issues/3572> ;
+    rdfs:comment "The status of the incentive (active, on hold, retired, etc.)." .
+
+:eligibleWithSupplier a rdf:Property ;
+    rdfs:label "eligibleWithSupplier" ;
+    :domainIncludes :FinancialIncentive ;
+    :rangeIncludes :Organization ;
+    :isPartOf <https://pending.schema.org> ;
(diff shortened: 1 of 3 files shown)


data/ext/pending/issue-3572.ttl
```diff
@@ -0,0 +1,323 @@
...
+:IncentiveQualifiedExpenseTypeUtilityBill a :IncentiveQualifiedExpenseType ;
+    rdfs:label "IncentiveQualifiedExpenseTypeUtilityBill" ;
+    :isPartOf <https://pending.schema.org> ;
+    :source <https://github.com/schemaorg/schemaorg/issues/3572> ;
+    rdfs:comment "This incentive applies to utility bills." .
+
+
+# Additions to existing types.
+:provider :domainIncludes :FinancialIncentive.
+:publisher :domainIncludes :FinancialIncentive.
```
alex-jansen, line 318: the description of publisher only refers to CreativeWorks, so that needs to be updated too

data/ext/pending/issue-3572.ttl
```diff
@@ -0,0 +1,323 @@
...
+<li>Use [[UnitPriceSpecification]] for incentives based on a per-unit basis (e.g. net metering).</li>
+<li>Use [[LoanOrCredit]] for loans/credits.</li>
+</ul>.""" .
+
+:incentivizedItem a rdf:Property ;
+    rdfs:label "incentivizedItem" ;
+    :domainIncludes :FinancialIncentive ;
+    :rangeIncludes :ItemList,
+        :DefinedTerm,
+        :Product ;
```
alex-jansen, line 122: Would it also make sense to have an individual Offer to be an incentivizedItem, e.g., one specific merchant selling the product providing the incentive while another merchant selling the same product not providing the incentive?

data/ext/pending/issue-3572.ttl
```diff
@@ -0,0 +1,323 @@
...
+# FinancialIncentive, the main class for all financial incentives.
+:FinancialIncentive a rdfs:Class ;
+    rdfs:label "FinancialIncentive" ;
+    :isPartOf <https://pending.schema.org> ;
+    :source <https://github.com/schemaorg/schemaorg/issues/3572> ;
+    rdfs:comment """<p>Represents financial incentives for goods/services offered by an organization (or individual).</p>
+
+<p>Typically contains the name of the incentive ([[name]]), the incentivized good/service ([[incentivizedItem]]), the amount offered by the incentive ([[incentiveAmount]]), the incentive's status ([[incentiveStatus]]), how the incentive is applied ([[incentiveType]]), the [[provider]] of the incentive, and what supplier the incentive applies to ([[eligibleWithSupplier]]).</p>
+
+<p>Optionally contains criteria on whether the incentive is limited based on ([[purchaseType]]), purchase price ([[purchasePriceLimit]]), income limits ([[incomeLimit]]), and whether the incentive applies to goods/services/both/utilities ([[qualifiedExpense]])
```
alex-jansen, line 16: some properties have links, some have not (e.g., [[purchase type]] but not "purchase price"). Please make consistent.

data/ext/pending/issue-3572.ttl
```diff
@@ -0,0 +1,323 @@
...
+
+In summary: <ul><li>Use [[QuantitativeValue]] for absolute/percentage-based incentives applied on the price of a good/service.</li>
+<li>Use [[UnitPriceSpecification]] for incentives based on a per-unit basis (e.g. net metering).</li>
+<li>Use [[LoanOrCredit]] for loans/credits.</li>
+</ul>.""" .
+
+:incentivizedItem a rdf:Property ;
+    rdfs:label "incentivizedItem" ;
+    :domainIncludes :FinancialIncentive ;
+    :rangeIncludes :ItemList,
```
alex-jansen, line 120: Is an ItemList really necessary? Do the elements need to be ordered in some particular use cases?

alex-jansen (commented):
Looks generally good, just left a few small comments to be addressed and then we can merge it in and put in the next release.

data/ext/pending/issue-3572.ttl
ArtixBot: The demo site has a link for each property in the description; which property is missing a link?

image

data/ext/pending/issue-3572.ttl
ArtixBot: For the example you mentioned, if two merchants are selling the same product but only one is offering an incentive, the merchant offering the incentive could markup their page (and not the merchant not offering the incentive) and specify themselves as both the provider and publisher.

We did consider using Offer, but it doesn't really work well in consideration of how financial incentives are implemented (e.g. net metering).

data/ext/pending/issue-3572.ttl
ArtixBot: Updated.

data/ext/pending/issue-3572.ttl
alex-jansen: I see, in some cases there is a term followed by the link in round brackets. In some cases there is not (it seems when the term matches the link exactly), which looked confusing to me when reading. A convention of links between brackets is not followed anywhere else in Schema.org type descriptions (at least I could not find any instances), maybe this can be changed to direct links? (just one example: https://schema.org/transitTimeLabel)

ArtixBot: Changed to direct links.

alex-jansen (approved):

alex-jansen:
Looks good now, approved and will include in next release

alex-jansen:
@ArtixBot An example still refers to ItemList causing the build to fail,  can you remove that?

alex-jansen:
thanks, merged.

PR #4384, opened 2025-03-30
DateTime in jsonldcontext

madbob:
File at https://schema.org/docs/jsonldcontext.jsonld misses all "DateTime" types for every property, mostly due a mispelled case in the building script.

### DIFF
--- a/software/tests/test_basics.py
+++ b/software/tests/test_basics.py
@@ -513,8 +513,18 @@ class BasicJSONLDTests(unittest.TestCase):
             "dateModified" in self.ctx["@context"], "dateModified should be defined."
         )
         self.assertTrue(
-            self.ctx["@context"]["dateModified"]["@type"] == "Date",
-            "dateModified should have Date type.",
+            len(self.ctx["@context"]["dateModified"]["@type"]) == 2,
+            "dateModified should have two types.",
+        )
+        self.assertIn(
+            "Date",
+            self.ctx["@context"]["dateModified"]["@type"],
+            msg="dateModified should have Date type.",
+        )
+        self.assertIn(
+            "DateTime",
+            self.ctx["@context"]["dateModified"]["@type"],
+            msg="dateModified should have DateTime type.",
         )
 
     def test_sameas_jsonld(self):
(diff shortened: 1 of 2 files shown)


gmackenz (approved):
Approved, thanks!

gmackenz:
Fails in the build test.

madbob:
Seems that the test itself was broken: [documentation says](https://schema.org/dateModified) that `dateModified` may be both `Date` or `DateTime`, while `software/tests/test_basics.py` just tested for `Date` and failed with the "fixed" version of `jsonldcontext.jsonld`.

gmackenz:
tests now breaks for auditDate

gmackenz (changes requested):
Test has another breakage for "auditDate"

madbob:
Ok, I totally failed to properly check my pull request... Sorry...

But now, I'm a bit confused.

JSON-LD specification [says something](https://www.w3.org/TR/json-ld/#specifying-the-type) about expressing a `type` "with an IRI", as "a term defined in the active context", or "using an array", but omits details about the contents of the array (the only available example contains IRIs, it do not mentions terms).

On the other hand, given my limited understanding of Ruby, seems that `json-ld` gem [only accepts](https://github.com/ruby-rdf/json-ld/blob/develop/lib/json/ld/context.rb#L594) a single string, in form of URI, assigned to `type`. Arrays seems not to be handled at all, with any content.

I've tried replacing `Date` and `DateTime` with `schema:Date` or fully qualified URLs as `http://schema.org/Date`, with no success in the validation by the `json-ld` gem. Probably I'm missing something, but don't know what...

gmackenz:
Not sure what value the test_basics for "Date" has and if we should keep it.

gmackenz:
Anyone else have some background on test or what should be done?

rrlevering:
Based on my read, the @type in JSON-LD context is a data/value type (or it would be type @id), so it's like the type of a typed literal in RDF.  So it really can't have multiple values in that definition.  It's not like a node type where JSON-LD allows multiple values.  So I don't really understand what you would put here if it could be either data type.  In Google's internal context file, we don't generate these types at all.  So I guess I would rather we just get rid of the declaration in that case.  Maybe someone with more background can take a stab here though.

PR #4395, opened 2025-04-07
Add referee property to SportsEvent (Issue #4318)

### LINKED ISSUE #4318: Add `referee` property to SportsEvent
It seems rather obvious to include this `Person` as a `SportsEvent`-specific property.  Granted, while multiple sports have multiple referees, there is usually one head match official who makes the final decision.

### New Property Request
Name: `referee`
Type: `Person`
Scope: `SportsEvent`

fthobe:
Adding referee to sports event as requested in #4318 .

Closes #4318

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -8296,6 +8296,12 @@ While such policies are most typically expressed in natural language, sometimes 
     :domainIncludes :MusicRelease ;
     :rangeIncludes :Organization .
 
+:referee a rdf:Property ;
+    rdfs:label "referee" ;
+    rdfs:comment "An official who watches a game or match closely to enforce the rules and arbitrate on matters arising from the play such as referees, umpires or judges. The name of the effective function can vary according to the sport." ;
+    :domainIncludes :SportsEvent ;
+    :rangeIncludes :Person .
+
 :referenceQuantity a rdf:Property ;
     rdfs:label "referenceQuantity" ;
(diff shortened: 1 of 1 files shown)


fthobe:
@DaveyJake @gmackenz can you both take a look?

data/schema.ttl
```diff
@@ -8296,6 +8296,12 @@ While such policies are most typically expressed in natural language, sometimes
     :domainIncludes :MusicRelease ;
     :rangeIncludes :Organization .
 
+:referee a rdf:Property ;
+    rdfs:label "referee" ;
+    rdfs:comment "The referee in a sports event." ;
```
gmackenz, line 8301: A referee can also be known as a umpire or judge and is a sports official who is in the position of being the final authority in administrating a sports event.

fthobe: Extend the comment?

gmackenz: Some suggestions, but yes extend the comment a little. Maybe we are fine as it is with `referee` being very loosely defined?

fthobe: How about "impartial" or "judge"?

gmackenz: I just thought we should allow for the common synonyms as some sports have judges and some sports have umpires but act the same as referees

fthobe: Ok, but do you want to edit the comment or rename the property? Both is fine, just to understand each other better.

fthobe: Or both.

fthobe: Updated the wording, should be better now. Can you squash on merge or do I need to?

fthobe: Even squashed with love

gmackenz (approved):
Approved

Issue #4397, opened 2025-04-08
Proposal new properties sdo:PostalAddress


RJMourits:
For exchanging historical and contemporary data from multiple countries, we use sdo:streetAddress. This causes problems for the automatic extraction of address data, as three elements of the address can appear in any order and not all elements are always present. These elements are: 1) street name, 2) house number, and 3) neighborhood. 
Would it be possible to add three properties (streetName, houseNumber, and neighborhood) to https://schema.org/PostalAddress, so that we can present the data without ambiguation? We feel the problem is universal enough for sdo. Moreover, we checked for other vocabularies, but found no viable alternative to sdo.

fthobe:
Hey, 
I can see where you are coming from, addresses across various countries are tough. 

Looking at schema for https://schema.org/PostalAddress going for a full address

Street: [streetAddress](https://schema.org/streetAddress)
ZIP Code: [postalCode](https://schema.org/postalCode) 
City: [addressLocality](https://schema.org/addressLocality)
Region: [addressRegion](https://schema.org/addressRegion)

Now you might get that region is tricky and I can tell you why: 

- Italy: Region would be sustituted with the Province;
- Germany: Region would be used with the real region, but nobody does that;
- US: Region would be used for state

In fact, that difference is reflected in schema as the description says: 
> The region in which the locality is, and which is in the country. For example, California or another appropriate first-level [Administrative division](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country).

So you should go crazy and use region for neighborhood and I promise you it will be valid and nothing will break. Many people do that.

RJMourits:
Thanks for the prompt reply!

Would this not result in a new ambiguity problem? In our case, we already use [addressRegion](https://schema.org/addressRegion) to describe Bundesländer, districts, and provinces. 

What we like about [postalCode](https://schema.org/postalCode), [addressLocation](https://schema.org/addressLocation), [addressRegion](https://schema.org/addressRegion), and [addressCounty](https://schema.org/addressCountry) is how disambiguous they are, which allows for stringent shacls and ontologies.

We appreciate that it is easy to have all address information in one line with [streetAddress](https://schema.org/streetAddress), but it enforces ambiguity in our data, which in turn causes ever-growing extraction problems.

fthobe:
Just keep in mind that not all data neccessarily has an appropriate place in schema. 

For practical reasons usually the combination 
- Address
- Postal Code
- City
- Country

[...]

PR #4402, opened 2025-04-11
Add extendedAddress to Address (concerns #4397)

fthobe:
Aligning the address with a second Address line as discussed in #4397.

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -5991,6 +5991,12 @@ This property should not be used where the nature of the alignment can be descri
     :rangeIncludes :Date,
         :DateTime .
 
+:extendedAddress a rdf:Property ;
+    rdfs:label "extendedAddress" ;
+    rdfs:comment "An address extension such as an apartment number, C/O or alternative name." ;
+    :domainIncludes :PostalAddress ;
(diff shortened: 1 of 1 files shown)


fthobe:
@rrlevering here you go.

fthobe:
@rrlevering can we pass that given the precedent in vcard and related address standards?

rrlevering (changes requested):
Generally looks good to me.

"appartment" -> "apartment" and probably "CO" -> "c/o"

What does "alternative name" mean?

fthobe:
@rrlevering should be fine now.
> "appartment" -> "apartment" 

Sorry, my italian heritage got the better of me.

> probably "CO" -> "c/o"

I made the change, anything to consider regarding escaping of any characters in TTL files?

> What does "alternative name" mean?

We run flagship stores for consumer electronic brands. Our company name does not neccessarily coincide with the name of the store. 
vCard standard says the following in [RFC 6350 (Section 6.3.1](https://www.rfc-editor.org/rfc/rfc6350.html#section-6.3.1): 
> the extended address (e.g., apartment or suite number);

rrlevering (approved):

Issue #4392, opened 2025-04-06
Allow automated testing also without approval


fthobe:
The automated testing workflow does not start before being triggered by a maintainer. 

It might make the workflow more fluid if automated testing is triggered immediately.

fthobe:
@gmackenz can I make also a PR for a broken link checker on MDs?

gmackenz:
MD as in mark down? A general broken link checker would be good.

fthobe:
yes, markdown via Github actions.

fthobe:
@gmackenz 
This seems to work very well: 
https://github.com/schemaorg/schemaorg/compare/main...fthobe:schemaorg:4392-update-github-actions
It needs a config file to exclude the release folders and localhost. But if this is a good road for you, I can finish it up and make a PR. 

The output looks like this: 

```
FILE: ./README.md
[✓] https://schema.org/
[✓] https://www.w3.org/community/schemaorg/
[✓] https://www.w3.org/2006/tools/wiki/Github
[✓] https://github.com/schemaorg/schemaorg/issues/1
[✓] https://schema.org/docs/releases.html
[✓] https://staging.schema.org/docs/releases.html
[✓] https://schema.org/docs/faq.html#18
[✓] http://schema.org/docs/releases.html
[✓] https://lists.w3.org/Archives/Public/public-schemaorg/2015Dec/0016.html
[✓] http://schema.org/docs/extension.html
[✓] https://ref.gs1.org/voc/
[✓] https://groups.google.com/g/ontolog-forum
[✓] https://blog-schema.org/2012/11/08/good-relations-and-schema-org/
[✓] https://schema.org/docs/news.html
[✓] https://iptc.org/standards/rnews/
[✓] https://www.poynter.org/product-category/fact-checking/
[✓] https://thetrustproject.org/
[✓] https://www.bbc.co.uk/ontologies
[✓] http://musicontology.com/docs/getting-started.html
[✓] https://musicbrainz.org/doc/LinkedBrainz/RDF
[✓] https://github.com/schemaorg/schemaorg/issues?q=is%3Aissue+label%3A%22Closed+and+Noted+%28and+possibly+Queued%29%22+is%3Aclosed
[✓] https://help.github.com/articles/about-pull-requests/
[✓] https://github.com/schemaorg/schemaorg/pulls
[✓] http://wikidata.org/
[✓] https://lov.linkeddata.es/dataset/lov/
[...]

PR #4403, opened 2025-04-11
RFC Broken link Checker

fthobe:
After a discussion in #4392 I hereby propose a broken link checker based on the work of @gaurav-nelson (Redhat). 

Gaurav has developed a broken link checker that works via Github actions and allows us to exclude: 
- folders like [release](https://github.com/schemaorg/schemaorg/tree/main/data/releases) given the frequently outdated links there;
- localhost for automated testing. 

Given that his work is licensed under MIT License, I'd like to
1. open up a repo in the Github organization of schmema.org;
2. configure the system properly with a config.json to exclude irrelevant folders and files;
3. add the workflow to schemaorg/schemaorg/

### DIFF
--- a/.github/workflows/broken-links.yml
+++ b/.github/workflows/broken-links.yml
@@ -1,0 +1,10 @@
+name: Check Markdown links
+
+on: push
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
--- a/.github/workflows/broken-links.yml
+++ b/.github/workflows/broken-links.yml
@@ -6,5 +6,5 @@ jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@main
     - uses: gaurav-nelson/github-action-markdown-link-check@v1


mfhepp:
Thanks - three comments:
1. There seems to be a newer tool by the same author: https://github.com/UmbrellaDocs/action-linkspector
2. I am a bit hesitant to add an external Github action to the workflow of such a high-visibility project, as supply-chain attacks and other bad things could happen (if e.g. the external action's repo gets compromised, e.g. introducing spam or malware links to documentation etc.
3. We should not automatically *remove* links that are "broken", e.g. temporarily return a non-200 HTTP status. E.g. a temporary outage of purl.org should not automatically remove all respective links.

fthobe:
> 1. There seems to be a newer tool by the same author: https://github.com/UmbrellaDocs/action-linkspector

I saw no tangibile benefit for schema and the old Solution is more lightweight. 

> 2. I am a bit hesitant to add an external Github action to the workflow of such a high-visibility project, as supply-chain attacks and other bad things could happen (if e.g. the external action's repo gets compromised, e.g. introducing spam or malware links to documentation etc.

Yes, i agree, That's Why i'd like to hard fork it into a Schema.org repo and not allow any modifications to schema by the action. I checked the license (permissive MIT) and that would be the road to go. 


> 3. We should not automatically *remove* links that are "broken", e.g. temporarily return a non-200 HTTP status. E.g. a temporary outage of purl.org should not automatically remove all respective links.

The Runner does not:
- change the Code of Schema
- solely Outputs a List in the Terminal of the GitHub Action with the broken links

fthobe:
@mfhepp any thoughts?

MatthiasWiesmann:
I also don't like having such external dependencies. This could be trivially be added to the python tests.

fthobe:
OK, closing this one and when I have some time see how to work it out with python.
> I also don't like having such external dependencies. This could be trivially be added to the python tests.

PR #4398, opened 2025-04-08
Remove twitter from readme (closes #4388)

fthobe:
Due to being unused by schema.org, twitter has been removed from readme.

closes #4388

PR #4427, opened 2025-04-18
Update README.md to add mention of the blog

gmackenz:
We should mention that we do have a blog (blog.schema.org). Mentioned in #4398

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ Every month or so, after final review by the Schema.org Steering Group and wider
 
 Regarding CC and opensource licenses for documents and software, see our [FAQ entry](https://schema.org/docs/faq.html#18).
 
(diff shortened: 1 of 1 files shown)


fthobe:
@gmackenz wouldn't it be better if we have it straight on top as a link list: 

- Official Website
- Blog
- Releases

In that way it's immediately visible.

gmackenz:
> @gmackenz wouldn't it be better if we have it straight on top as a link list:
> 
> * Official Website
> * Blog & Updates
> * Releases
> 
> In that way it's immediately visible.

Just wanted to mention for now so throw in a quick link in the README. Any more general site improvements can discussed in a separate new issue.

fthobe:
ok, thank you :)

alex-jansen (approved):

PR #4432, opened 2025-04-22
Fix the wrongly labeled 'pending' properties.

pascalfleury:
From the addition of shipping information a few properties were labelled as being pending, which is wrong as they have been in used already.

this fixes the offenders.


alex-jansen (approved):
added some manual edits to fix the htps: prefixes as well.

Issue #4396, opened 2025-04-07
Invalid prefixes added for 'http://schema.org/rdfs:' in 29.0


nickmacdonald7:
The prefix `http://schema.org/rdfs:` was added to several data files in the 29.0 release (e.g. [schemaorg-current-http.ttl](https://github.com/schemaorg/schemaorg/blob/194c5a3046cf9fc8815db58041e6c779cf8400ec/data/releases/29.0/schemaorg-current-http.ttl#L5),  [schemaorg-all-http.ttl](https://github.com/schemaorg/schemaorg/blob/194c5a3046cf9fc8815db58041e6c779cf8400ec/data/releases/29.0/schemaorg-all-http.ttl#L5), [schemaorg-all-http.rdf](https://github.com/schemaorg/schemaorg/blob/194c5a3046cf9fc8815db58041e6c779cf8400ec/data/releases/29.0/schemaorg-all-http.rdf#L4))

It's used by the `schema:rdfs:label` property of `schema:weight`, which looks to be an accidental duplicate/typo of the `rdfs:label` property.

nickmacdonald7:
This was fixed in the v.29.1 update 🥳

PR #4433, opened 2025-04-23
Fix the faulty added ns1: namespace and the wrong prefix with rdfs.

pascalfleury:
This fixes Issue #4396.


pascalfleury:
I see this was also fixed already.

PR #4435, opened 2025-04-23
Re-add 'Mass' as a range type for 'weight'.

pascalfleury:
This was wrongly removed while fixing the pending state of the property.


pascalfleury:
Did not see that it was already added, just moved in the file.

pascalfleury:
this is unneeded now.

PR #4441, opened 2025-04-29
extended address example (#4397)

fthobe:
Added sample data for extendedAddress added in #4402 .

@rrlevering @gmackenz I am unsure where to put sample data as neither of the sample files seems to be versioned or anything. I took a large approach. Could you check if I understood it correctly?

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -7,6 +7,7 @@ Jane Doe
 
 Professor
 20341 Whitworth Institute
+Suite 123
 405 Whitworth
 Seattle WA 98052
 (425) 123-4567
@@ -29,6 +30,7 @@ MICRODATA:
   <div itemprop="address" itemscope itemtype="https://schema.org/PostalAddress">
     <span itemprop="streetAddress">
       20341 Whitworth Institute
+      Suite 123
       405 N. Whitworth
     </span>
     <span itemprop="addressLocality">Seattle</span>,
@@ -59,6 +61,7 @@ RDFA:
   <div property="address" typeof="PostalAddress">
     <span property="streetAddress">
       20341 Whitworth Institute
+      Suite 123
       405 N. Whitworth
     </span>
     <span property="addressLocality">Seattle</span>,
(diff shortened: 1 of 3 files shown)


gmackenz:
I think that is the location? I would look at past PRs for evidence of examples appearing in schema.org?

I wonder if it would be better to not update existing examples of address usage and instead add a specific new example of extended address usage?

fthobe:
@gmackenz I think your european colleagues are all out on vacation due to the long weekend 👯

MatthiasWiesmann:
Hi, sorry this pull request got ignored.
Was it necessary to change the examples in the code directory?

fthobe:
Up to you

fthobe:
@MatthiasWiesmann yes, I went through the changes and would add the example to show that it's not meant to insert corners or anything but "C/O" or similar fields.

MatthiasWiesmann:
> @MatthiasWiesmann yes, I went through the changes and would add the example to show that it's not meant to insert corners or anything but "C/O" or similar fields.

It's just that the examples are not used anywhere…

fthobe:
Can you elaborate? If you write me two lines where you'd like to see them I can add them.

data/examples.txt
```diff
@@ -29,6 +30,7 @@ MICRODATA:
   <div itemprop="address" itemscope itemtype="https://schema.org/PostalAddress">
     <span itemprop="streetAddress">
       20341 Whitworth Institute
```
MatthiasWiesmann: Just wondering if this is consistent, basically, I would do one per street address line. 
```
<span itemprop="streetAddress">20341 Whitworth Institute</span>
<span itemprop="extendedAddress">Suite 123</span>
<span itemprop="streetAddress">405 N. Whitworth</span>
```

fthobe:
Ahh shit, literally I understand now, sorry, so if we don't use them let's drop the examples fully and I edit the ones we use. 
@matthiaswiesmann

fthobe:
@MatthiasWiesmann should be fine now and be visible on respective pages as example.
https://schema.org/PostalAddress
Screenshot 2026-01-04 at 18 48 55

fthobe:
@gmackenz I think this can be approved now (it's only providing an example)

Issue #380, opened 2015-03-13
Representing VAT component of a price


danbri:
* There is currently no way to indicate the VAT portion of a PriceSpecification
* We do have a boolean http://schema.org/valueAddedTaxIncluded
* Most countries have a VAT system, http://en.wikipedia.org/wiki/Value-added_tax#Different_systems
* We should somehow make it possible to indicate how much of a price is from VAT

/cc @vholland

sandstrom:
:clap:

chaals:
+1 to Sandstrom

danbri:
Rescheduled for Q2. @vholland do you have a sense how close we are?

vholland:
I was thinking something along the lines of @sandstrom's suggestion: a vatAmount. I think I could get something together to propose with examples after sdo-gozer.

danbri:
Thanks, that would be great. I'll leave it as Q2 for now. We'll need a codename for the next release soon btw...  how about sdo-tully? per http://ghostbusters.wikia.com/wiki/Louis_Tully QOTD ""Who does your taxes...?"

danbri:
Ping - where are we with this?

mfhepp:
We are almost one year older, wiser, and a few hairs have changed their colors to a grayish tone ;-)
Other than that, the issue is still pending...

mfhepp:
But in any case, we need something that can handle relative price components, like a percentage on the unit price. A quick fix like vatAmount won't help IMO. Maybe a new subtype of PriceSpecification like TaxSpecification with properties like

taxPercentage
applicableRegion
exemptRegions

We will also need a mechanism to link the tax to the price components. This can either be done by adding a taxInformation property that links PriceSpecifications to TaxSpecifications or Number, or by a
priceReference property to link to the price components that are subject to the tax.

If we opt for the second route, we should encourage the use http://sdo-hotels.appspot.com/CompoundPriceSpecification in such scenarios.

sandstrom:
Feel free to disagree with me, but would a relative tax percentage really be needed?

Mostly a `vatAmount` would be enough to back out the vat percentage if needed (` vat-amount / (total – vat-amount) = vat-rate`).

If the total price is a compound of multiple items (with different rates) this won't work. In those cases a per item price (with the vat amount) would work though. I assume it's already possible to specify prices both in aggregate and with specificity.

***
[...]

PR #4447, opened 2025-05-07
boolean to indicate sales tax exemption support for offers

fthobe:
This is a first step to incorporate taxes into offers laying the ground work to define acceptance of sales tax reductions. 

In many jurisdictions there are several reasons to be exempted from sales tax. Below some examples: 
- in many countries people with disability receive sales tax / vat discount (eg Italy blind people 4% instead of 22%);
- the entire European union supports VAT exemption for businesses dealing directly (eg one company selling a service to another); 
- in the US sales tax can be exempted in many cases between businesses.

This is achieved through two booleans on offers: 

1. `acceptsTaxExemptionsDisability` allows businesses to markup disability sales tax reductions; 
2. `acceptsTaxExemptionsBusiness` allows businesses to markup B2B sales tax reductions; 

This PR introduces the ability to markup sales tax exemption support for merchants.

Concerns (but does not solve):  #380

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -3998,6 +3998,18 @@ See also the <a href="/docs/hotels.html">dedicated document on the use of schema
         :Text,
         :URL .
 
+:acceptsTaxExemptionsDisability a rdf:Property ;
+    rdfs:label "acceptsTaxExemptionsDisability" ;
+    rdfs:comment "Indicates whether the seller processes orders with disabilty related tax exemptions." ;
+    :domainIncludes :Offer ;
+    :rangeIncludes :Boolean .
+
+:acceptsTaxExemptionsBusiness a rdf:Property ;
+    rdfs:label "acceptsTaxExemptionsBusiness" ;
+    rdfs:comment "Indicates whether the seller processes orders with business tax exemptions." ;
+    :domainIncludes :Offer ;
+    :rangeIncludes :Boolean .
+
 :accessCode a rdf:Property ;
     rdfs:label "accessCode" ;
     rdfs:comment "Password, PIN, or access code needed for delivery (e.g. from a locker)." ;


fthobe:
@alex-jansen this is a proposal to add some groundwork for VAT (the ability of the merchant to process orders without VAT / Sales tax). 

Examples would be: 

- [Amazon Sales Tax Exemption Program](https://www.amazon.com/gp/help/customer/display.html?nodeId=G2UQTUL5CHRCA7BL)

- [Candle Science](https://support.candlescience.com/hc/en-us/articles/201199700-How-do-I-become-tax-exempt-at-CandleScience)

It would be for me a sensible first step, I have some questions though: 
1. locally sales taxes have different names (VAT in Europe, GST in India and many Asian countries, Sales Tax in the US), how could we account for the differences in naming this?
2. What would be a good comment in your opinion?

MatthiasWiesmann:
I really don't think adding a flurry of booleans to [Offer](https://schema.org/Offer) is the way to go. 

I would suggest we define a separate object named TaxInformation or something similar which would contain various tax bits, but also meta-information, in particular:

* Who this tax information applies to: where (country codes, zip codes in the US), who (people with disabilities, business, schools, etc). The second one would be a enum and that can be extended as we discover that shops painted red have a discount.

fthobe:
> I really don't think adding a flurry of booleans to [Offer](https://schema.org/Offer) is the way to go. 
> 
> 
> 
> I would suggest we define a separate object named TaxInformation or something similar which would contain various tax bits, but also meta-information, in particular:
> 
> 
> 
> * Who this tax information applies to: where (country codes, zip codes in the US), who (people with disabilities, business, schools, etc). The second one would be a enum and that can be extended as we discover that shops painted red have a discount. 
> 
> 

Hey, 

Ok, I am struggling in between what I'd like to see, what you'd like to see and what's possible: 

Generally calculating taxes and shipping and displaying it on the frontend is "complicated". There are three approaches: 

- In the US you calculate taxes after you have a shipping address for consumers and B2B alike as even a zip code can be insufficient to calculate taxes precisely (yes, people have even been fined for that). There's an entire ecosystem of companies calculating taxes for eCommerce. 
- in Europe for consumer and mixed shops you usually calculate taxes for your country and if you ship intracomunity (other EU countries) or outside EU, you recalculate after shipping address
- on B2B shops you don't show taxes at all in Europe as you assume businesses either do not pay them or pay at checkout and get them back. 

Your multiple taxes system is great in theory, I just don't believe it's relevant for schema or structured data as the info will not be available.

alex-jansen:
The problem is still that these tax exemptions are often locale-dependent (e.g., as you mentioned the reduction for the blind in Italy), so should be specifiable by country (similar to how shipping costs are specified by destination). Just having a single exemption boolean at the Offer level cannot generally be sufficient.

fthobe:
@alex-jansen please don't take this the wrong way. 

I fully agree with what you say in general and I am big fan overdeclaring compared to underdeclaring, but the reality is that Offers does not define a locale and is usually used following guidelines of regionally limited platforms (eg Google Shopping with one feed for one market). I fully get your point (really I do), but the point is that usually different regionals have different pricings (and different market feeds) and that's usually solved either via subfolders for locales (eg /de-DE/) or fully separate domains. Where there are cross border transactions, taxes are not to be applied (and usually removed during checkout). 

In sense that, yes, you are right, but no, price and offer currently do not work that way. If thats supposed to be changed the amount of work would probably be different as we are talking about introducing a regional component to offer or price and I'd suggest that I close this. 

So, if we want to make this a defined term and extend the information we can totally do it, I just don't believe that, considering the ecosystem, anything other than declaring B2B tax exemptions and declaring disability tax exemptions has any chance to be adopted (by google and bing). 


So if you want me to expand it I do it, I just wonder who can consume the additional fields as of right now (as proposed) they'd work anywhere (eg US and all of Europe, but also eg Brazil's new vat regulation), but there's no taker for multi market price specifications in schema as of right now (and also no platform from Shopify to BC or magento doing this by default).

alex-jansen:
Note that for delivery costs and times Schema.org does support complex region-dependent lookups through [ShippingConditions](https://schema.org/ShippingConditions) and [OfferShippingDetails](https://schema.org/OfferShippingDetails). 

If there is a concrete use-case to add a few booleans to support tax-exemptions that's fine and we should add them. But, forward-looking, it would still be good to encapsulate those in a separate object, similar to what was done for delivery cost and times, for example using a new *OfferTaxDetails* object that we can easily expand in the future.

(Also note that prices specified in Schema.org for non-US locales usually include VAT (unless the [valueAddedTaxIncluded](https://schema.org/valueAddedTaxIncluded) property on [PriceSpecification](https://schema.org/PriceSpecification) is explicitly specified as False))

rashworld-max (approved):

fthobe:
@alex-jansen did I have a stroke or did someone just approve this?

PR #4461, opened 2025-05-27
Expand recipeIngredient range

rrlevering:
Expand range on recipeIngredient to allow for more structured data.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -1408,7 +1408,7 @@ JSON:
   "recipeIngredient": [
     "3 or 4 ripe bananas, smashed",
     { "@type": "PropertyValue", "value": 1, "name": "egg" },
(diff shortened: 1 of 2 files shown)


gmackenz (approved):
Should we add a new example showing use of ItemList or add onto to the existing example? Approved otherwise.

rrlevering:
> Should we add a new example showing use of ItemList or add onto to the existing example? Approved otherwise.

I actually purposely didn't.  ItemList is most there as an ordered list because schema.org doesn't allow order technically and I feel like that may change in the future.

redphoenix2uk:
I'm getting errors trying to use ItemList for the recipeIngredient - ItemList is not a known valid target type for the recipeIngredient property:
image

rrlevering:
Sorry, the rollout into the validator stalled on Friday.  It's hopefully going now so should be available in the few hours.

Issue #4483, opened 2025-08-21
Missing property isDropShipped


wf-frank:
Often, eCommerce websites will offer items for sale by third parties. Adding "DropShipped" to the allowed ItemAvailability values could make it clear that a product is shipped by a third party and might arrive separately.

alex-jansen:
How would this work in the case of an online retailer (a drop shipper) who marks their products as ItemAvailability=InStock, based on the fact that they know their supplier has the product available in their inventory (and OutOfStock if not available). Could drop-shipping be considered a separate property from ItemAvailability? 

For example: Drop shipping means that the item happens to be shipped from a 3rd party supplier's warehouse instead of the retailer's own warehouse, which makes it perhaps more a fulfillment/shipping-related property?

wf-frank:
Yes, it makes sense to have a separate property isDropShipped (boolean), as this could signify that a customer should refer to the merchants terms and conditions for drop shipped items versus regular stocked items.

Then they can use ItemAvailability: OnlineOnly, which may cover these cases:

1. The merchant does not know how much stock is available or when it will be available.
2. The merchant does not know how long it will take to deliver.
3. The merchant wants the customer to know that an order will arrive in separate shipments.

I updated the PR.

wf-frank:
What should we have for `domainIncludes` on the isDropShipped property ?

Some suggestions:
:Order
:Offer
:OfferShippingDetails

alex-jansen:
Since drop shipping is fundamentally related to fulfillment I think we need it at least on [:OfferShippingDetails](https://schema.org/OfferShippingDetails) and [:ShippingService](https://schema.org/ShippingService)? We could consider adding it to [FulfillmentTypeEnumeration](https://schema.org/FulfillmentTypeEnumeration), but that is more about how the consumer can pick up the product, so probably not as appropriate.
@pascalfleury FYI

Issue #2651, opened 2020-07-14
Proposal for adding BackOrder as an enumeration member of ItemAvailability


chollma:
## Context
Online retailers are especially dependent on schema.org vocabulary as a standardized way to describe the evolving status of their products to a variety of systems including search engines. Today, there are a number of ways that a product can be described using the `ItemAvailability`attribute.

## Problem
I believe there's a significant gap in this list, especially given the current state of consumer demand for online shopping, the encumbered state of our global supply chains during a pandemic, and our commitment to provide the best possible information available to customers.

The current version of the schema.org vocabulary includes `PreOrder` as a enumeration member of `ItemAvailability`. As I understand it, `PreOrder` refers to an item that is available for purchase but has **not been carried** by a business before.

## Proposal
I would like to propose the addition of `BackOrder` to this list of members, because I believe there is a unique difference between `PreOrder` and `BackOrder`. 

## Details
As I understand it, `BackOrder` refers to an item that is out of stock, still able to be purchased and will be available again for customers at some point in the future, rather than being `Discontinued`.

Inclusion of `BackOrder` in the list of enumeration members for `ItemAvailability` will allow consumers of schema.org to describe the status of an item with more granularity than `OutOfStock`. In addition, it will allow other major business to utilize an updated vocabulary to better serve customers with their products (ie. https://support.google.com/merchants/answer/6324448?hl=en)

RichardWallis:
See issue #1977 near by.

chollma:
Thank you, @RichardWallis. I think #1977 is similar in nature to this request but not identical.

philsward:
Yes, this is still a request as per my original request in #1632 

Like I stated in my original request, BackOrder is different than OutOfStock.  Neither of which, have anything to do with PreOrder.

> **Backorder vs. out of stock**
> Out of stock means that a product does not currently have any inventory available and does not have a date for resupply, while ‘backordered’ implies there is a determined date for products to arrive. 
> 
> It’s the difference between “This item is currently unavailable” and “This item won’t ship [until 2 weeks from now, for 10 business days, etc.].” In other words, there is hope in the foreseeable future with a backorder. It might take a while, but you will receive the product. 
> 
> When a product is ‘out of stock,’ there’s a chance that’s the case permanently, or at least will be for so long that the seller can’t predict when they will have it again.

Source: https://www.shipbob.com/blog/backordered-meaning/

[...]

PR #4484, opened 2025-08-21
Add ItemAvailability value "DropShipped" (#4483)

frankforte:
Add the value "DropShipped" for ItemAvailability, to signify products shipped by a third party.

### DIFF
--- a/data/schema.ttl
+++ b/data/schema.ttl
@@ -1024,10 +1024,6 @@ Region = state, canton, prefecture, autonomous community...
     rdfs:comment "The act of producing a visual/graphical representation of an object, typically with a pen/pencil and paper as instruments." ;
     rdfs:subClassOf :CreateAction .
 
-:DropShipped a :ItemAvailability ;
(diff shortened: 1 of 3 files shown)


MatthiasWiesmann:
Please do not edit the core schema file, create new files in https://github.com/schemaorg/schemaorg/tree/main/data/ext/pending with a number matching the relevant bug.

fthobe:
@frankforte there was a discussion about that here in #2651

There's some point in having a distinction between dropshipping stock and regualar stock, but keep in mind that this difference can also be solved by definition of delivery windows other added information you can provide around a product (e.g. a well defined shipping method).

data/schema.ttl
```diff
@@ -6662,6 +6658,12 @@ In the context of [[ShippingService]], use the [[ServicePeriod]] format, that co
     :domainIncludes :Product ;
     :rangeIncludes :Product .
 
+:isDropShipped a rdf:Property ;
```
alex-jansen: please move to pending as suggested

Issue #3697, opened 2024-10-18
BookFormatType, GraphicNovel


joric3:
https://bib.schema.org/BookFormatType lists the following types:

- [AudiobookFormat](https://bib.schema.org/AudiobookFormat)
- [EBook](https://bib.schema.org/EBook)
- [GraphicNovel](https://bib.schema.org/GraphicNovel)
- [Hardcover](https://bib.schema.org/Hardcover)
- [Paperback](https://bib.schema.org/Paperback)

EBook, Hardcover and Paperback are publication formats for various types of content, mostly for (written) text or text with illustrations. It might be images and nothing else. In other words, these types describe an outer form and not the content. The primary mode for the content presentation is visual, but otherwise content is undefined. (Some books include multimodal content like haptic elements or sound, e.g. in children's books, but that is beside my point here.)

An Audiobook is for spoken text, obviously using auditory modality. I can understand that is practical to list it as a BookFormatType even though it is primarily used for text and cannot adequately convey images. (It can convey sound beyond spoken word, obviously - sound effects or music.)

A graphic novel is a longform comic, a specific kind of content or medium (in which text and images combine in specific ways). A graphic novel can be published in print (Hardover or Paperback) or electronic format (EBook). To me, it does not make sense to conceptualize as a BookFormatType next to the others. It's like listing "artbook" or "short story collection" on the same level as EBook, Hardcover, Paperback. That's mixing type of content with output formats. Therefore, I think graphic novel should be removed here.

(I'm new to this, apologies if I made mistakes.)

LokiMidgard:
This would be nice.

gmackenz:
GraphicNovel came from BIB as an extension in 2014. It might have been a mistake to make it a BookFormatType instead of a subtype of Book as the source intended it seems [Periodicals and Comics synthesis - Graphic Novels](https://www.w3.org/community/schemabibex/wiki/Periodicals_and_Comics_synthesis#Notes_and_Explanation_of_Terminology).

LokiMidgard:
What would be the implications changing this?

joric3:
And what would be the implications of not changing this? ;)

LokiMidgard:
If I understand it correctly, bookFormat has a single value. So having a hardcover graphic novel, would mean I can only define it as hardcover or graphic novel (but it is both).

[...]

PR #4531, opened 2025-10-14
Deprecate GrapicNovel in BookFormatType and add a seperate LiteraryForm wich includes this.

LokiMidgard:
This is a proposal in regards to #3697. Feel free to comment and edit ;)

---

Just removing (depricating) the GrapicNovel seems not to be a good idea, since someone created it because he needed it.
So instead there is now a subtype of Book, ComicBook.

This pull request has 3 changes:

- The GrapicNovel in BookFormatType was depricated. (In comment, I havn't seen any actually syntax to depricate stuff)
- A new subclass of book that is ComicBook.
- Extend Genre Attribute 

---

I have yet to check if it builds, since I currently have no python enviroment…

### DIFF
--- a/data/ext/bib/bsdo-1.0.ttl
+++ b/data/ext/bib/bsdo-1.0.ttl
@@ -93,7 +93,7 @@
 
 :GraphicNovel a :BookFormatType ;
     rdfs:label "GraphicNovel" ;
-    rdfs:comment "This type is deprecated: GraphicNovel does not fit the BookFormatType enumeration, as it can appear in multiple formats (e.g., Hardcover, eBook). It is not mutually exclusive and therefore deprecated. Use standard BookFormatType values instead.\n\nBook format: GraphicNovel. May represent a bound collection of ComicIssue instances." ;
+    rdfs:comment "This type is deprecated: GraphicNovel does not fit the BookFormatType enumeration, as it can appear in multiple formats (e.g., Hardcover, eBook). It is not mutually exclusive and therefore deprecated. Use standard BookFormatType values instead in combination with the SequentialArt.\n\nBook format: GraphicNovel. May represent a bound collection of ComicIssue instances." ;
     :isPartOf <https://bib.schema.org> .
 
 :abridged a rdf:Property ;


MatthiasWiesmann:
I think it's disputable, as graphic novels are printed on different media than regular book, different paper for one.

LokiMidgard:
I would not say that this proposal is "final" I'm happy to discuss and change it (or other change it). Or discard it all together if it is not good enough.

I agree that it is disputable. Reading Wikipedia it seems clear, that it is not a well defined Word.

>The term represents an attempt to distinguish longer comics—often promoted as thematically more sophisticated—from traditional Western comic books and albums, a distinction that is also reflected in their broad distribution through the book trade.

- *Translated from the German [Wikipedia site](https://de.wikipedia.org/w/index.php?title=Graphic_Novel&oldid=252979240)*


But from my perspective It would still make sense to change it, to better catalog different Versions of Graphic novels.

Discussions:
 - https://github.com/bookwyrm-social/bookwyrm/issues/2743
 - https://github.com/schemaorg/schemaorg/issues/3697

*(And sorry that the text below got so long…)*

---

The original Description in the schema said:
>Book format: GraphicNovel. May represent a bound collection of ComicIssue instances.

Wich is more a description of the content of the book, 

While EBook
> Book format: Ebook.

It describes the *physical* or in this case *non-physical* form of the book, but can be any content.


---

There are also different versions of graphic Novels that could be distinct from each other by the BookFormatType but are also all Visual Novels.

E.g. there are different Versions of `Little Monsters Volume 1` wich I would say is a Graphic Novel  (It is also in the Amazon category)
The versions that exist are (That I'm aware of)
- Paperback (listed on Amazone)
- EBook (listed on Amazone, Kindle)
- Hardcover (Is in my Collection)

image

- [Source](https://www.amazon.de/Little-Monsters-Vol-1-English-ebook/dp/B0B82YVNB1/ref=sr_1_4?__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&s=books&sr=1-4&language=en_GB#)

And while Graphic Novels tend to have different paper (pure white, compared to a more yelowish tone in standard novels) that is not a good distinction.

The non Fictional book Immun by Philipp Dettmer has  pretty much the same paper as the graphic Novel Vampire Boy by Carlos Trillo and Eduardo Risso (The reprint by 2019). The reason they both have probably the same paper is that the first one has many color full illustrations that are better  printed on pure white paper, while normal novels that consists only of text do not need that.
I would say it is more a correlation not a direct causality. The paper is not used because it is a Graphic novel,  but because pictures look better on this paper. On the other hand Novels probably use not pure white because it is nicer to read (and maybe cheaper?).

---

So I would not say HardCover and Graphic Novel are mutual exclusive Book types, but that a Graphic Novel is a different type (wich I  called LiteraryForm) that is in distinction to Novel.

Though my proposal also dose not cover something like a biography written as graphic novel…

---

Also I find it much harder to differenciate what is a Comic, Manga and what an Graphic Novel.

A Comic is a  Story told in Pictures, with (mostly) direct speech (other wise its a picture book)
A Manga is the same but from Japan.
A Graphic Novel is the first (so not from Japan) but with more Pages. Sometimes just some Comics glued together. Or just a comic marketed as serious story.

---

Maybe an even better from would be:
A sample graphic novel:
```
Name: Little Monsters Volume 1
BookFormat: HardCover
LiteraryForm: Novel
PrimaryPresentation: Graphical
```

compared to
```
Name: Erkental
BookFormat: HardCover
LiteraryForm: Novel
PrimaryPresentation: Textual
```

and Audio books would be

```
Name: Erkental
BookFormat: AudioBook
LiteraryForm: Novel
PrimaryPresentation: Audio
```

But this may getting a little bit to detailed…

MatthiasWiesmann:
I would decouple [bookFormat](https://schema.org/bookFormat) which represent the physical (or virtual) format of a book, I agree there is a need for a child of book for comic-books like media, which would be appropriate for manga, bande dessinée, comic-books. I also agree that we need some more bookFormat values to represent different physical book formats, but mainly we need more documentation on how to use the existing attributes. 

I also agree the use of the audio-book format should be discouraged, mostly because the relevant attributes for an audio-book ([duration](https://schema.org/duration) and [readBy](https://schema.org/readBy)) are defined in the [Audiobook](https://schema.org/Audiobook) object. The relationship with the original book can be expressed using the [isBasedOn](https://schema.org/isBasedOn) property. 

As to how to represent various comic-book styles. 

* Thick manga: ComicBook, bookFormat: [Paperback](https://schema.org/Paperback)
* Classical belgian Bande Dessinée: ComicBook, bookFormat: [Hardcover](https://schema.org/Paperback)
* US style weekly comic-book: ComicBook, bookFormat:  Pamphlet (aka floppy), to be added.
* US style graphic novel: ComicBook, [GraphicNovel](https://schema.org/GraphicNovel) (glued binding)
* Online comic-book:  ComicBook, bookFormat: [EBook](https://schema.org/EBook)

The LiteraryForm property can be expressed using the existing [genre](https://schema.org/genre) attribute. 
The length can be expressed using the existing [numberOfPages](https://schema.org/numberOfPages) property. 

The **dimensions** of the comic book can be specified with the size attribute, so a classic franco-belge will have this property:

```
size: {
  "@type": "QuantitativeValue",
   "name": "Height",
   "value": 32
   "unitCode": "CMT",
}
size: {
  "@type": "QuantitativeValue",
   "name": "Height",
   "value": 24
   "unitCode": "CMT",
}
```

You could also use standard terminology for sizes, i.e. "A4", "Bonellide".

Does this make sense?

LokiMidgard:
I like the addition of `Pamphlet`, I would have used Paperback for those, but this would make much more sense.

Using an existing Property is better then making a new one, but I'm not quite sure how genre works since it seems to be an URL? And I would have expected it to be something like Sci-Fi, but the comment reads
> Genre of the creative work, broadcast channel or group.

And I'm not sure what the genre of a group is… So yeah explaining these things could help, especially if it is expected to be used by people not in the domain, like me. I only have some comic books and I tried to add some of those to bookwyrm and ended up here 🙃

If I understood your comment correct you would add a sub-type of book, similar to audio-book? This would enable adding aditional propertis special for comic books while having everything of book. Wich seems like a good approach. 

I still don't see the difference between:
- Classical belgian Bande Dessinée: ComicBook, bookFormat: [Hardcover](https://schema.org/Paperback)
- US style graphic novel: ComicBook, [GraphicNovel](https://schema.org/GraphicNovel) (glued binding)
If I look at it by physical property's.

Do you have two examples for those? The comments should help for people like me to be able to asigne a book to one of these properties by some simple properties. That could help.

Those I think I understand I would define like this
- Pamphlet 
  - A booklet normaly bind by staples, no defined spine. Often fewer pages, but that is not a criteria.
- Softcover
  - Normaly glued pages, with a spine and Front and Back are of thicker paper
- Hardcover
  -  A robust front, back and spine. Normaly the back and front are not bendable.

MatthiasWiesmann:
I created a small pull request https://github.com/schemaorg/schemaorg/pull/4591 which adds the pamphlet format and clarifies the existing ones. 
What I would suggest for following steps

* extend the [genre](https://schema.org/genre) attribute to accept a DefinedTerm, then you can define a DefinedTermSet with literary genres.  
* Add the ComicBook subclass. 
* Deprecate the GraphicNovel book type.

LokiMidgard:
I added the ComicBook subclass, The GraphicNovel book type is already depricated (by comment, I don't know if there is a better way)

I also tried to extend the genre attribute, but I must admit I don't known how it is actually done…
I used an LLM to describe me how it is done (hopefully correct). But I'm not that confident in it. But at least it is in the last commit wich can easily be reverted :)

MatthiasWiesmann:
I gave this a bit more though and would also propose the following: 

* Call the subclass "SequentialArt" which is the formal term that encompasses comic-books, manga, bande dessinée etc. 
* Add attributes for `inker`, `colorist` and `letterer` as those can be different  (author and illustrator are already present in [Book](https://schema.org/Book)).

MatthiasWiesmann:
The genre attribute is extended in https://github.com/schemaorg/schemaorg/pull/4592
Generally, it pays to split work into small PR which have limited scope.

LokiMidgard:
I reverted the genre. I also noticed that inker, colorist and letterer are already present in VisualArt. So I Thought it would make more sense to extend SequentialArt with VisualArt and Book

MatthiasWiesmann:
> I reverted the genre. I also noticed that inker, colorist and letterer are already present in VisualArt. So I Thought it would make more sense to extend SequentialArt with VisualArt and Book

100% agree.

data/ext/bib/bsdo-1.0.ttl
```diff
@@ -62,6 +62,13 @@
     rdfs:subClassOf :CreativeWork ;
     :isPartOf <https://bib.schema.org> .
 
+:SequentialArt a rdfs:Class ;
+    rdfs:label "SequentialArt" ;
+    rdfs:comment "An art forms that use images deployed in a specific order for the purpose of graphic storytelling (i.e., narration of graphic stories) or conveying information. The best-known example of sequential art is comics." ;
```
MatthiasWiesmann: I would replace this with something like:

"Examples of SequentialArt are Franco-Belgian Bande Dessinée, Comics in the USA and <ruby>漫画<rp>)</rp><rt>manga</rt><rp>)</rp></ruby> in Japan.

LokiMidgard: Ok, changed it

PR #4547, opened 2025-10-23
migrated schema to use CurrencyCode, CountryCode, and LanguageCode

MatthiasWiesmann:
Added `Text` subclasses for `CountryCode`, `LanguageCode` and `CurrencyCode`. 
Migrated all relevant attributes to use these.
Functionally, this makes no difference, but makes the documentation much clearer, and could allow for better validation.

### DIFF
--- a/software/SOFTWARE_README.md
+++ b/software/SOFTWARE_README.md
@@ -80,7 +80,9 @@ Locally one can run tests that assert that the schema is self-consistent, and th
 
     software/util/buildsite.py -a -r --rubytests
 
-This will auto-build the entire site (`-a`), run the tests (`-r`) as well as validate the examples (`--rubytests`).
+This will auto-build the entire site (`-a`), run the tests (`-r`) as well as
+validate the examples (`--rubytests`). If you get some Ruby errors when testing,
+please check the installation steps in the `Build Notes` section below.
 
 
 Deploying to GCloud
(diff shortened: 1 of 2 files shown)


data/releases/3.3/ext-health-lifesci.ttl
pascalfleury: same here. what is in releases we just leave as historical trail.

data/schema.ttl
```diff
@@ -4169,8 +4169,7 @@ See also the <a href="/docs/hotels.html">dedicated document on the use of schema
         :GeoCoordinates,
         :GeoShape,
         :PostalAddress ;
-    :rangeIncludes :Country,
-        :Text ;
+    :rangeIncludes :CountryCode, :Country ;
```
pascalfleury: this essentially dark launches the change, doesn't it?

pascalfleury:
Can you also remove the edits to the past releases 3.2 and 3.3? These schemas would no longer work if we do these edits.

pascalfleury:
it seems you pushed another change ([2dc8f4e](https://github.com/schemaorg/schemaorg/pull/4547/commits/2dc8f4e2c1870b67e02d5959f0ae09d5b86dec8f)) which undoes your reverting the editing of released files...

MatthiasWiesmann:
It should be fixed now.

pascalfleury (approved):

Issue #4513, opened 2025-09-29
Proposal — New property `onDisplay`


vascosommernunes:
## Problem statement (user need)

**“Where can I try this now?”**  
This is the everyday question for people looking to experience products before they buy and also for visitors deciding what art to see in person. Today, Schema.org can tell us if a thing is purchasable in a store (InStoreOnly), but it cannot say whether the item is actually on view for the public — on a showroom floor or on a gallery wall — right now.

- `InStoreOnly` ≠ on display. It simply says you can buy it offline, but not whether you can view and experience it.  
- For consumers and visitors, this leaves the key question unanswered.  

We have tried to make use of schema with an association of 40 design furniture retailers. But found that the schema ontology, which has such a great trait of being able to being explicit, lacks this addition which we feel would be of benefit for many.

### New property: `onDisplay` (Boolean)

- **domainIncludes:** `Product`, `VisualArtwork`  
- **rangeIncludes:** `Boolean`  
- **Definition:** Indicates whether the item is currently **publicly viewable** (e.g., on a showroom floor or on a gallery wall).  
- **Usage note:**  
  - Use `onDisplay: true` when the public can physically see or try the item at a `Place`.  
  - This complements commerce signals; it does not replace them.  
  - To indicate the place, reuse existing properties (e.g. `location` on the item page, or `hasPart` on the place page).  

## Affected types

- `Thing > Product` → add `onDisplay`  
- `Thing > CreativeWork > VisualArtwork` → add `onDisplay`  

## Compatibility

- This is a **non-breaking extension**. Existing mark-up remains valid.  
- `onDisplay` is complementary: it provides an answer to a distinct, high-value user question while preserving established commerce semantics.  

## Value to society

A minimal addition, but a meaningful one:  
- **For consumers:** answers the practical question “Where can I try this now?” across places.  
- **For retailers and museums:** makes visibility of their on-floor or on-wall inventory indexable, discoverable, and AI-friendly.  
- **For the web ecosystem:** promotes structured answers to a question that otherwise requires scraping, guessing, or manual updates.  

## References

- Blog post:  
  [See it in person: Proposing onDisplay for schema.org](https://blog.showroom.fm/index.php/2025/09/26/113/)

- Namespace definition (draft):  
  [https://blog.showroom.fm/index.php/152-2/](https://blog.showroom.fm/index.php/152-2/)

- Museum precedent: Rijksmuseum legacy OAI-PMH documentation (deprecated, but shows “on display” flag):
[...]

PR #4550, opened 2025-10-29
Add display location property

vrugtehagel:
Specifies a place at which a product or creative work can be viewed or experienced. Note that this does not imply that the item can be purchased at this location, nor that the item is purchaseable altogether.

Resolves https://github.com/schemaorg/schemaorg/issues/4513

### DIFF
--- a/data/ext/pending/issue-4513-example.txt
+++ b/data/ext/pending/issue-4513-example.txt
@@ -1,0 +1,41 @@
+TYPES: #eg-4513 Product, displayLocation
+
+PRE-MARKUP:
+
+An example of a product with a displayLocation.
+
+MICRODATA:
+
+TODO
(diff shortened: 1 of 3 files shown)


MatthiasWiesmann:
Please don't change the main schema file, create a file numbered after the issue in `data/ext/pending/` with the extensions.

MatthiasWiesmann:
Also consider adding an example file that shows how this property is used.

vrugtehagel:
Sorry about that, moved it to `data/ext/pending/` and added an example for how it's used. Thanks!

vascosommernunes:
Thank you @vrugtehagel and @MatthiasWiesmann for your support! The internet will be a big tiny bit better with this one.

Issue #4579, opened 2025-11-11
Error subclass of AtomicEvent to use with Actions


MatthiasWiesmann:
Currently, [Action](https://schema.org/Action) instances have an [error](https://schema.org/error) field which is defined as a Text field. This field should also accept some form of structured error information. This could be achieved by subclassing the AtomicEvent class.

Issue #4527, opened 2025-10-10
Proposal: Schemas for Systems' States and Error Messages


AdamSobieski:
## Proposal

Hello. I would like to propose creating schemas for [error reports](https://en.wikipedia.org/wiki/Exception_handling), [stack traces](https://en.wikipedia.org/wiki/Stack_trace), [call stacks](https://en.wikipedia.org/wiki/Call_stack), [stack frames](https://en.wikipedia.org/wiki/Call_stack#STACK-FRAME), and related concepts.

A motivating use case is enabling and enhancing AI-related scenarios via JSON-LD representations of systems' [states](https://en.wikipedia.org/wiki/State_(computer_science)) and [contexts](https://en.wikipedia.org/wiki/Context_(computing)). There are several other interesting uses to consider; some are indicated below.

Firstly, there are Web scenarios to consider such as embedding JSON-LD representations of systems' states and error messages in `<script>` element [data blocks](https://html.spec.whatwg.org/multipage/scripting.html#data-block) within HTML resources.

```html
<html>
  <head>
    <script type="application/ld+json">
      {
        ...
      }
    </script>
  </head>
  <body>
    <p>Hello world!</p>
  </body>
</html>
```

Secondly, today, developers and end-users can share error-report data in Web-based discussion forums. This sharing often involves copying and pasting formatted text output by software systems for these systems' developers to make use of.

What if software systems' states and error messages could be output into, processed into, and/or pasted into standard machine-readable formats or schema?

Perhaps new software tools could, then, be created to enhance technical communication, enabling the indexing of, searching for, retrieval of, analysis of, and visualization of these data.

Perhaps, resembling [Mermaid diagrams](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) on GitHub, developers and end-users could paste systems' states and error messages into [markdown code blocks](https://markdown.land/markdown-code-block) for convenient formatting and visualization features.

````markdown
```state
{
  ...
}
```
````
````markdown
```error
{
  ...
}
```
````

Thirdly, having a standard means of representing systems' states and error messages would enhance a number of protocols, e.g., [MCP](https://modelcontextprotocol.io/docs/getting-started/intro), which include error-handling capabilities, enabling or enhancing the exchange of systems' states between AI agents and/or systems' other components.

[...]

PR #4589, opened 2025-11-17
Added error class

MatthiasWiesmann:
Added a basic error class. 
It currently only has one attribute `errorCode`. 
attributes to specify things like stack traces can be added later. 

See https://github.com/schemaorg/schemaorg/issues/4579


MatthiasWiesmann:
The author of https://github.com/schemaorg/schemaorg/issues/4527 closed the bug as good, so I assume this would satisfy the requirement.

PR #4594, opened 2025-11-20
Add schema snapshotting util and a snapshot test

jlndr:
Adding:

* minimal snapshotting utilities which exports the `all-https` turtle version of the current schema.
* 'golden' file of the current schema
* snapshot test which checks that the current snapshot is up to date with the current schema

This ensures that changes to the complete schema can be observed easily, including changes/additions to the schema itself as well as any code building the schema.

The test could be a fail fast test to run in CI. We can add an github action to run it separately in a follow up. The test runs with the rest of the tests already, but this is quite slow.

demo:

Screenshot 2025-11-19 at 14 55 14

### DIFF
--- a/software/tests/test_snapshot.py
+++ b/software/tests/test_snapshot.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import difflib
+import logging
+import os
+import sys
+import sys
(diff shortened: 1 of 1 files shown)


MatthiasWiesmann (approved):

pascalfleury (approved):
that's a good start :-)

PR #4596, opened 2025-11-21
Moved phonetic examples away from PhoneticText

MatthiasWiesmann:
Two more ruby text exceptions down! 

The PhoneticText class never really worked, causes errors in the ruby tests, and is not really needed, as the same effect can be achieved with multiple @language tagged values. This pull request migrates the examples away into the relevant classes and attributes (name), the actual PhoneticText class will be migrated in another pull request.

### DIFF
--- a/data/ext/pending/issue-2108-examples.txt
+++ b/data/ext/pending/issue-2108-examples.txt
@@ -1,8 +1,8 @@
-TYPES: #eg-0270 PronounceableText, phoneticText
+TYPES: #eg-0270 City, name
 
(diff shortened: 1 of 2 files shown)


rrlevering:
Ok, I'm ok with this for fixing the tests and because data types with properties breaks many systems but I don't like this reliance on JSON-LD syntax for language.  We need to discuss some better textobject class or to propose changes to the syntax specs.

MatthiasWiesmann:
100% agree

PR #4620, opened 2025-12-01
Fix equivalentClass typo

dpb587:
Hello - I noticed a case mismatch with `owl:equivalentClass` and assume it is unintentional.

For what it's worth, I appreciate the recent efforts to document more related ontologies here - thanks.


jlndr (approved):
Thanks!

Issue #4621, opened 2025-12-02
Compiled schema should add rdf:type triples for external classes


rrlevering:
Our import process chokes with the newer external schema class references (in the upcoming release) because they are not recognized as classes.  I could change it to assume that subClassOf always points to a type, but I think it would be cleaner if they were typed for completeness.

For example:
```
<https://spec.edmcouncil.org/fibo/ontology/FND/Arrangements/Documents/Certificate> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2000/01/rdf-schema#Class> .
```

pascalfleury:
Let's add them for the time being, but I think it would make sense to implement your suggestion, as the classes are essentially defined by their linking to the original schema. A complete system would even go there and verify that they are classes.

MatthiasWiesmann:
Created https://github.com/schemaorg/schemaorg/pull/4625 that addresses this.

MatthiasWiesmann:
> Let's add them for the time being, but I think it would make sense to implement your suggestion, as the classes are essentially defined by their linking to the original schema. A complete system would even go there and verify that they are classes.

My reading of the code is that the post-processing phases strips these out try to run: `./software/util/buildfiles.py All` and check the content of `software/site/releases/29.3/schemaorg-all-http.ttl`

PR #4625, opened 2025-12-03
Defining external classes as RDFs Classes in the schema.

pascalfleury:
Some tool choke on an externally defined class that is not marked as such in the schema. This PR adds these as classes.

This is a mitigation of #4621.


MatthiasWiesmann:
My reading of the code is that the post-processing phases strips these out try to run: ./software/util/buildfiles.py All and check the content of software/site/releases/29.3/schemaorg-all-http.ttl

pascalfleury:
This is unneeded now as the list of external classes is auto-computed and added in code.

PR #4656, opened 2025-12-22
Hardened docs page building logic

MatthiasWiesmann:
Hardened the doc page building logic.

- Better error reporting if a term is not found.
- Removed global variables.
- Use sets for set logic
- Templates now output full types if a non schema type is encountered.
- Added type information to functions


fthobe:
Can we run any broken link checker also on the releasenotes page, that's where most of the broken links come from.

pascalfleury (approved):

Issue #4666, opened 2025-12-30
Allow more types for suitableForDiet


MatthiasWiesmann:
[suitableForDiet](https://schema.org/suitableForDiet) should allow more different diets.

* It should accept the [Diet](https://schema.org/Diet) type. 
* It should allow [DefinedTerm](https://schema.org/DefinedTerm) for exact diet definitions  (Halal vs JAKIM/MUIS/BPJPH).
* Some missing enum values could be added.

fthobe:
Do we want to add any of the diet types to the enum?

I can carry the changes over to a correct PR. GS1 has this and we could reference the vocabulary: https://ref.gs1.org/voc/DietTypeCode

MatthiasWiesmann:
You can add the mapping, with the caveat that on the GS1 side, these are type-codes (strings) and not enums.

fthobe:
Is the any example of a currently valid external mapping how you would imagine it?

MatthiasWiesmann:
> Is the any example of a currently valid external mapping how you would imagine it?

Well, let's stay the pastafarian church decides to define two diets:

```
{
  "@context": "https://schema.org/",
  "@type": "DefinedTermSet",
  "@id": "https://example.org/pastafarian-dietary-canon",
  "name": "Pastafarian Dietary Canon",
  "description": "The official classification of dietary paths recognized by the Church of the Flying Spaghetti Monster.",
  "hasDefinedTerm": [
    {
      "@type": "DefinedTerm",
      "@id": "https://example.org/terms/OrthodoxNoodly",
      "name": "Orthodox Noodly Diet",
      "termCode": "FSM-01",
      "about": {
        "@type": "Diet",
        "name": "Orthodox Noodly Diet",
        "description": "A diet centered on the regular consumption of long-strand pasta to honor the Noodly Appendages of the Creator.",
        "dietFeatures": "Requires the use of a colander for draining; requires 'Ramen' prayer.",
        "recognizingAuthority": {
          "@type": "Organization",
          "name": "Church of the Flying Spaghetti Monster"
        },
        "keywords": "Pasta, Spaghetti, Meatballs, Carb-loading"
      }
    },
    {
      "@type": "DefinedTerm",
      "@id": "https://example.org/terms/HighSeasPirate",
      "name": "High-Seas Pirate Diet",
      "termCode": "FSM-02",
      "about": {
        "@type": "Diet",
        "name": "High-Seas Pirate Diet",
[...]

Issue #3176, opened 2022-09-03
Proposal: RestrictedDiet :: LowCarbDiet


tomkenkel:
Thing > Intangible > Enumeration > RestrictedDiet :: LowCarbDiet

A diet focused on reduced carbohydrate intake.

A member value for enumeration type: [RestrictedDiet](https://schema.org/RestrictedDiet)

chrisspradling1980:
Crystalline 
-please take note of the word not tokenized
Gel: resulting as substance pertaining to a diet calling for RestrictedDiet
I agree with the inverse Domain requests particular to the EnvironmentType self incrimination occurs because of USCA titles and federal Scheduling for possession. Symbolic expression is needed to timecode token experiences. This will never stop with humans.

RichardWallis:
See also #3211

GinMorgWireless:
#3211 offers more information than total carbs. People who need the tags LowSucrose and LowMaltose need the lower resolution than just carb levels. They often need to eat glucose and fructose to make up for the reduced sucrose, maltose and starch. The LowCarb tag is not suitable on its own as foods can be "low carb" and still contain too much sucrose, maltose or starch for safe digestion in this population.

fthobe:
@GinMorgWireless wouldn't that be different kinds of diets?

GinMorgWireless:
Yeah I’d love to get these tags added as low carb and low disaccharide
tags, even low monosaccharides like fructose would be helpful for lots of
people. It would actually make a huge difference for people who are looking
for these specific medical dietary needs.

I’m not sure how to progress something like this further through the schema
though??

fthobe:
Can you make a complete list inside a table with: 

- name of the diet
- comment describing the diet

You can find the full list of currently supported diets here: 
[RestrictedDiet](https://schema.org/RestrictedDiet)

And if you click on any of the elements in the list you can find the comment here: 
<img width="657" alt="Image" src="https://github.com/user-attachments/assets/b779dc5f-532e-46e8-888f-60380c593364" />

If you provide me with a name and comment I will check if I get interest on the formal PR to add it to schema.org.

GinMorgWireless:
Carbohydrate Malabsorption Special Medical Diets

LowLactoseDiet
A diet appropriate for people with lactose intolerance.
LowSucroseDiet A diet appropriate for people with sucrose intolerance. Some
people with CSID only need to restrict sucrose
[...]

PR #4667, opened 2025-12-30
allow Diet for suitableForDiet

MatthiasWiesmann:
See [issue 4666](https://github.com/schemaorg/schemaorg/issues/4666)

### DIFF
--- a/data/ext/pending/issue-4666.ttl
+++ b/data/ext/pending/issue-4666.ttl
@@ -1,0 +1,5 @@
+@prefix : <https://schema.org/> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+
+:suitableForDiet a rdf:Property ;
+  :rangeIncludes :Diet.


alex-jansen (approved):

fthobe:
Could you add samples?

alex-jansen:
Before I create a PR with an example, does the snippet below give a good idea of the intention? (the key here is the use of the [sameAs](https://schema.org/sameAs) property to link to one or more web pages with an official definition of the specific named diet (the `description` property  under `Diet` is optional and I just added it for illustration)

```
{
  "@type": "Recipe",
  "name": "Avocado panzanella",
  "image": "https://wwww.examplemenus.com/avopan.jpg",
  "description": "A simple recipe for a quick and easy meal with avocado.",
  "suitableForDiet": [
    {
      "@type": "Diet",
      "name": "Vegan",
      "description": "A plant-based eating pattern that excludes all foods derived from animals",
      "sameAs": [
        "https://www.nhs.uk/live-well/eat-well/how-to-eat-a-balanced-diet/the-vegan-diet/",
        "https://en.wikipedia.org/wiki/Veganism",
        "https://snomed.info/sct/24930006"
      ],
      "code": {
        "@type": "MedicalCode",
        "codeValue": "24930006",
        "codingSystem": "SNOMED-CT"
      }
    },
    {
      "@type": "Diet",
      "name": "Mediterranean",
      "description": "Diet based on the traditional eating habits of people in countries bordering the Mediterranean Sea.",
      "sameAs": "https://en.wikipedia.org/wiki/Mediterranean_diet"
    }
  ]
}
```

As far as I understand there is no "official" dictionary (for example, maintained by a government or health organization) with Diet codes that can be referenced. The best I found is some diets that are specified as medical (SNOMED) codes . I illustrated that as well above for Vegan diet to show how official codes for diets can be added if available. One additional thing we can do here is allow `DefinedTerm` as value for the `code` property to reference other authoritative well-defined and non-medical external enumerations of Diets, but only if these are available.

@TechHappyNZ and @fthobe based on #3176

fthobe:
Here's what I'd do: 
- there are some diet types defined in gs1, I'd actually provide content parity with them as it seems extremely low hanging fruit, it might one day be handy for recipies and also food / grocery shopping applications, gs1 made the template, I would only migrate the missing ones over into the current diet set (which seems to derive from there anyway)

- everything else be left to custom implementation citing external standards like you suggested

The reasoning behind that: 
- all gs1 labels are commercially established for any boxed food, why not keep them as most of them are already in the enum and only few are missing and they do seem to cover the majority of lifestyle diet related searches (low carb, keto, low fat,...)
- i understand what Matthias pointed out (schema.org is not a nutrition content platform), so for all non gs1 purposes i think what you suggested is brilliant

ps813957-prog:
### b_4666_allow_diet

Issue #4469, opened 2025-07-18
Move floorLevel up from Accommodation to its parent Place


Coffee2CodeNL:
Since any physical place can be on a different floor, malls have different floors too for instance, so a local business needs a floorLevel too.

gmackenz:
Not opposed to this but some examples of usage? Up to now this is almost always used with real estate listings ([example](https://validator.schema.org/#url=https%3A%2F%2Fwww.99acres.com%2Fstudio-apartment-flat-for-sale-in-vital-east-point-mhada-colony-mulund-central-mumbai-280-sq-ft-r1-spid-O81052701)).

I don't encounter floor level for places other than apartments or hotel room.

chaals:
Would be helpful for a huge number of business locations in many places.

Do you mean you don't encounter people using it?

gmackenz:
Not that I've observed but it might be a Catch-22 situation because it is not available for all subtypes of Place. Again can we come up with some examples of markup for allowing floor level on Place (maybe a pull request)?

Coffee2CodeNL:
The only reason I'm proposing to pull it up is because a Place can be on a different Floor.

If it's better to just add floorLevel to a LocalBusiness, that's fine too.

I have no experience with PRs for Schemas, it's just something I ran into while working on data where we have LocalBusinesses on different floors.

MatthiasWiesmann:
I would add the  [floorLevel](https://schema.org/floorLevel) property to [LocalBusiness](https://schema.org/LocalBusiness) and [Residence](https://schema.org/Residence) . I'll prepare a pull request.

PR #4669, opened 2025-12-30
added floorLevel to LocalBusiness and Residence

MatthiasWiesmann:
See https://github.com/schemaorg/schemaorg/issues/4469


MatthiasWiesmann:
> Maybe add to issue-2373.ttl which is also in pending (and just add this issue to the :source) ? (to avoid distributing this in different places)

Done.

alex-jansen (approved):

Issue #4490, opened 2025-09-04
Proposal: previousStartDate rangeIncludes DateTime


dpb587:
Hello, I propose adding [DateTime](https://schema.org/DateTime) as an expected data type of [previousStartDate](https://schema.org/previousStartDate). Currently it expects a [Date](https://schema.org/Date).

In name and description, it seems like it should support the same values as [startDate](https://schema.org/startDate) (which expects both Date and DateTime).

> ... This property contains the previously scheduled start date. For rescheduled events, the startDate property should be used for the newly scheduled start date. ...

The addition of DateTime would help it match apparent usage, too.

* Only a couple examples include this property ([eg-0172](https://schema.org/Event#eg-0172), [eg-0202](https://schema.org/MusicEvent#eg-0202)), but they both use a DateTime value.
* Externally, Google's [Event Structured Data docs](https://developers.google.com/search/docs/appearance/structured-data/event#previous-start-date) describe it as a DateTime.

It looks like [Release v3.1](https://schema.org/docs/releases.html#v3.1) (2016) added DateTime to startDate, but previousStartDate was unchanged.

dpb587:
Closing - implemented via #4679 and released in [v30.0](https://schema.org/docs/releases.html#v30.0).

PR #4679, opened 2026-01-08
Add previousStartDate rangeIncludes DateTime

dpb587:
This would bring it inline with conventional usage, existing repository examples, and seeming-intended usage. Without this DateTime inclusion, schema-based validation of previousStartDate assume that only a Date is valid and emit an error.

Since this is an amendment of an existing property, I updated `schema.ttl` directly rather than a dedicated `data/ext/pending/*` file that seem to be for new definitions, but let me know if I should change it.

This is a follow-up to my original issue about this in #4490, although there was not further discussion on it.


pascalfleury:
Indeed, `startDate` and a whole set of other `*Date` (`datePublished`, `dateCreated`, `dateModified`, `validThrough`) do accept both `Date` and `DateTime`, so this makes sense.

PR #4684, opened 2026-01-13
Adding example and documenting member tiers.

pascalfleury:
In any places, they can be specified as multiple tiers, which is now reflected in the documentation and the examples.


MatthiasWiesmann (approved):

pascalfleury:
Modified the example to use an annotation of an organization with shipping services and multiple member tiers.

Issue #3548, opened 2024-06-13
Adding coherent examples of JSON-LD implementation of SiteNavigationElement


ChiefRA:
Hello,
Since I've seen that there are no examples on SiteNavigationElement implementation at the bottom of the Schema page https://schema.org/SiteNavigationElement I've decided to add one as it might be of help to anyone trying to implement this: 

```
<script type="application/ld+json">
  {
  "@context": "https://schema.org",
  "@graph":
    [
      {
        "@context": "https://schema.org",
        "@type":"SiteNavigationElement",
	"@id":"#table-of-contents-1",
        "position": 1,
	"name": "Page 1 Item",
        "url": "https://www.example.com/page-1"
      },
      {
        "@context": "https://schema.org",
        "@type":"SiteNavigationElement",
        "@id":"#table-of-contents-2",
	"position": 2,
        "name": "Page 2 Item",
        "url": "https://www.example.com/page-2"
      },
      {
        "@context": "https://schema.org",
        "@type":"SiteNavigationElement",
        "@id":"#table-of-contents-3",
        "position": 3,
        "name": "Page 3 Item",
        "url": "https://www.example.com/page-3"
      },
      {
        "@context": "https://schema.org",
        "@type":"SiteNavigationElement",
        "@id":"#pagination-4",
        "position": 4,
        "name": "Page 4 Item",
        "url": "https://www.example.com/page-4"
      },
      {
        "@context": "https://schema.org",
        "@type":"SiteNavigationElement",
        "@id":"#pagination-5",
        "position": 5,
        "name": "Page 5 Item",
        "url": "https://www.example.com/page-5"
      }
    ]
  }
</script>
```

This implementation successfully validates on https://validator.schema.org/ 

Please add it to the rightful section? 
Thanks,
Arthur

WeaverStever:
@ChiefRA 

On the schema validator, only two of your 5 example properties pass (each @Id needs to be unique). 


I played around with doing this with a stronger type. 

This seems to pass the validator.

(Note that I have type with both the "SiteNavigationElement" and "Webpage")

~~~
<script type="application/ld+json">
{
    "@context":"https://schema.org",
    "@type": "website",
    "@id": "/#",
    "name": "Title of the website",
    "description" : "The official website for example.com", 
     "url": "https://example.com/",
     "hasPart": [
      {
        "@type": ["SiteNavigationElement", "Webpage"],
        "@id": "/#navlink1",
        "name": "webpage-1",
[...]

PR #4686, opened 2026-01-14
json Example for SiteNavigationElement (#3548)

fthobe:
Adds an example to SiteNavigationElement.
Credits for example go to @ChiefRA .
Closes: #3548


fthobe:
@gmackenz this also only provides examples

PR #4775, opened 2026-02-27
Clarify the definitions of accessMode and accessModeSufficient

mattgarrish:
It's come to our attention that our original definitions lack some clarity when it comes to adaptations of the content. We've kept the rules around setting the access modes and sufficient access modes from the original Access for All properties these were derived from in epub usage documentation, but feel it would be helpful to improve the wording of the schema.org definitions to make clearer that access modes do not take into consideration adaptations while sufficient access modes do.

The new wording has been discussed in https://github.com/w3c/a11y-discov-vocab/issues/254 as well as in the publishing community group.

/cc @clapierre @madeleinerothberg


mattgarrish:
Thanks!

PR #4776, opened 2026-03-03
Update schemaorg.css to fix searchbox styling issue

maybeiambatman:
Programmable Search Engine (CSE) has changed their default styling which was clashing with schema.org searchbar. When entering new text the first few characters are obscured by the search icon. I've removed the custom padding that causes this issue. Also removed the static heigh which makes the icon look off center vertically.

#### Mobile
Before:
image
After:
image

#### Desktop
Before:
image
After:
image


rrlevering:
I think you need to sync your fork to pick up the test I just fixed.  I'm not sure if I can do that, but my github-fu is not strong.

PR #4777, opened 2026-03-05
Change Quantity into a DataType instead of an Intangible.

rrlevering:
Practically all of the online data shows all these subtypes being primitives, not nodes.  So make this consistent.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -3807,8 +3807,9 @@ JSON:
     "name": "Test Candidate One"
   },
   "object": {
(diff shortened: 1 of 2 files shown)


rrlevering:
I could have made this a subtype of Text which is somewhat appropriate for Mass, etc. but then I'd have to pull out Duration separately to be consistent with DateTime and that was sounding messier.

alex-jansen:
Looks reasonable, if there are no objections I will merge this PR in on Friday in preparation for a new release

PR #4782, opened 2026-03-19
[Docs] Getting Started: Match previously mentioned URL for movie trailer

jezmck:
Otherwise it's just unnecessarily confusing.


jlndr:
Thanks!

PR #4808, opened 2026-04-24
feat(MedicalSpecialty):

jasonshanks:
Added a missing MedicalSpecialty for Ophthalmologic. This differs from existing Optometric which is solely for the diagnose of, not surgical treatment of eye issues.

### DIFF
--- a/data/ext/health-lifesci/med-health-core.ttl
+++ b/data/ext/health-lifesci/med-health-core.ttl
@@ -942,6 +942,12 @@
     rdfs:comment "A trial design in which the researcher knows the full details of the treatment, and so does the patient." ;
     :isPartOf <https://health-lifesci.schema.org> .
 
+:Ophthalmologic a :MedicalSpecialty ;
+    rdfs:label "Ophthalmologic" ;
+    rdfs:comment "A specific branch of medical science that pertains to the diagnosis and treatment of disorders of the eye and its associated visual structures, including medical and surgical interventions." ;
+    rdfs:subClassOf :MedicalBusiness ;
+    :isPartOf <https://health-lifesci.schema.org> .
+
 :Optometric a :MedicalSpecialty ;
     rdfs:label "Optometric" ;
     rdfs:comment "The science or practice of testing visual acuity and prescribing corrective lenses." ;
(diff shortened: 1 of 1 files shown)


rrlevering (approved):

jasonshanks:
Apologies, before maintainer approves and merges I made a minor correction to the terminology to the more prevalent 'Ophthalmic'. Is there any known timeline on when this might occur?

jasonshanks:
Hello. I was what the status/timeline might be for getting this workflow approved and the PR merged into main branch?

data/ext/health-lifesci/med-health-core.ttl
```diff
@@ -942,6 +942,12 @@
     rdfs:comment "A trial design in which the researcher knows the full details of the treatment, and so does the patient." ;
     :isPartOf <https://health-lifesci.schema.org> .
 
+:Ophthalmologic a :MedicalSpecialty ;
```
jlndr: The term name and label should match. Should this be 'Ophthalmology'? See [Dermatology](https://schema.org/Dermatology), [Pathology](https://schema.org/Pathology)

Ophthalmology seems to be the relevant [wikipedia entry](https://en.wikipedia.org/wiki/Ophthalmology)

jasonshanks: Yes 'Ophthalmology' is correct. Do you need me to resubmit with change or will you?

jlndr: Feel free to do it and I'll be able to merge

jasonshanks: Thanks @jlndr . I've updated term and label now to: Ophthalmology

Comparing this more closely to the others, doesn't it seem like there is a mix of adjectives and nouns within MedicalSpecialty? For instance should these also change from:
- Optometric -> Optometry
- Osteopathic -> Osteopathy
- Obstetric -> Obstetrics
- Neurologic -> Neurology
- Homeopathic -> Homeopathy
several more, etc. etc.

jlndr: > doesn't it seem like there is a mix of adjectives and nouns within MedicalSpecialty?

Yes definitely - seeing the differences between all them I'm not sure if we have/had any strict guidelines regarding this. Unfortunately it's hard to rename these once it's released and in use

PR #4819, opened 2026-05-19
Move Python files that are *scripts* into /scripts/ and adjust.

pascalfleury:
Scripts (Python files expected to be called on the command line) are moved into the script dir.

- we also adjust all other Python files, documentation;
- we remove all the import path handling from non-script files;
- we further assume the directory from which these are run is the root of the project (i.e. `schemaorg`) and therefore loading other Python files can be greatly simplified;
- we moved any import that is not at the top of the file to ... the top of the file;
- the example-code is also adjusted so it works like described in the documentation.

### DIFF
--- a/software/SchemaExamples/example-code/out
+++ b/software/SchemaExamples/example-code/out
@@ -1,4 +1,4 @@
-TYPES: #eg-313 DefinedTerm, termCode
+TYPES: #eg-0314 DefinedTerm, termCode
 
 PRE-MARKUP:
 
@@ -32,7 +32,7 @@ JSON:
  "inDefinedTermSet": "http://id.loc.gov/vocabulary/resourceTypes"
 }
 
-TYPES: #eg-314 DefinedTerm, DefinedTermSet, inDefinedTermSet
+TYPES: #eg-0315 DefinedTerm, DefinedTermSet, inDefinedTermSet
 
 PRE-MARKUP:
 
@@ -58,7 +58,7 @@ MICRODATA:
 
 <div>
 	<div itemscope itemtype="https://schema.org/DefinedTermSet">
-	<meta itemprop="additionalType" content="https://schema.org/Book">
+	<link itemprop="additionalType" href="https://schema.org/Book">
 	 <h1><a itemprop="url" href="http://openjurist.org/dictionary/Ballentine"><span itemprop="name">Ballentine&apos;s Law Dictionary</span></a></h1>
 	</div>
 	<div itemscope itemtype="https://schema.org/DefinedTerm">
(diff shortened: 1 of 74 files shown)


software/SchemaExamples/example-code/out
```diff
@@ -32,7 +32,7 @@ JSON:
  "inDefinedTermSet": "http://id.loc.gov/vocabulary/resourceTypes"
 }
 
-TYPES: #eg-314 DefinedTerm, DefinedTermSet, inDefinedTermSet
+TYPES: #eg-0315 DefinedTerm, DefinedTermSet, inDefinedTermSet
```
jlndr, line 35: the 0 padding I wouldn't be so surprised by - do we know where the +1 increment is coming fro?

software/SchemaExamples/example-code/out
```diff
@@ -58,7 +58,7 @@ MICRODATA:
 
 <div>
 	<div itemscope itemtype="https://schema.org/DefinedTermSet">
-	<meta itemprop="additionalType" content="https://schema.org/Book">
+	<link itemprop="additionalType" href="https://schema.org/Book">
```
jlndr, line 61: Does the change which cause this diff impact the site generation also?

software/scripts/compareterms.py
```diff
@@ -1,52 +1,40 @@
...
-dev_appserver.fix_sys_path()
-
-from testharness import *
+if os.getcwd() not in sys.path:
+    sys.path.insert(1, os.getcwd())
+import software
 
-# Setup testharness state BEFORE importing sdo libraries
-setInTestHarness(True)
+# Ensure that the google.appengine.* packages are available
```
jlndr, line 22: this is a test? can this be in main?

software/scripts/compareterms.py
```diff
@@ -1,52 +1,40 @@
...
-    read_extensions,
     getMasterStore,
+    read_extensions,
+    read_schemas,
 )
 from apirdflib import getNss
+from testharness import *
 
+# Setup testharness state BEFORE importing sdo libraries
+setInTestHarness(True)
```
jlndr: same here, can this be in main?

software/SchemaExamples/example-code/out
pascalfleury: it does count from a number, based on the number of valid types we have. As I let Gemini verify the examples it could be that these were outdated anyway. and there is no test to assert that the examples a) result in this, and b) actually work.

software/SchemaExamples/example-code/out
pascalfleury: This is the same reason, the examples code has been verified.

software/scripts/compareterms.py
pascalfleury: Removed.

jlndr (approved):

PR #4832, opened 2026-05-29
feat(decoupled-stats): implement server-side baked HTML stats badging

williamgslima:
Implement compile-time monthly stats rendering on term pages, decoupling statistics presentation completely from client-side JS loader modules to improve performance, stability, and security.

### DIFF
--- a/software/scripts/buildsite.py
+++ b/software/scripts/buildsite.py
@@ -7,6 +7,7 @@ import argparse
 import contextlib
 import datetime
 import glob
+import json
 import logging
 import os
 from pathlib import Path
@@ -33,6 +34,7 @@ import util.fileutils as fileutils
 import util.paths as paths
 import util.pretty_logger as pretty_logger
 import util.schema as schema
+import util.stats as stats
 
 
 log: logging.Logger = logging.getLogger(__name__)
@@ -126,32 +128,44 @@ def initialize() -> argparse.Namespace:
         action="store_true",
         help="create page for term (repeatable) - ALL = all terms",
     )
+    parser.add_argument(
+        "--buildrelease",
+        default=False,
+        action="store_true",
+        help="Build the release files and save them into the releases data directory",
+    )
+    parser.add_argument(
+        "--buildsite",
+        default=False,
+        action="store_true",
+        help="Build the website from the releases directory + static content + stats",
+    )
 
     args: argparse.Namespace = parser.parse_args()
 
     op: List[str]
     for op in args.buildoption:
-        schema.constants.BUILDOPTS.extend(op)
+        schema.config.BUILDOPTS.extend(op)
 
     ter: List[str]
     for ter in args.terms:
-        schema.constants.TERMS.extend(ter)
+        schema.config.TERMS.extend(ter)
 
     pgs: List[str]
     for pgs in args.docspages:
-        schema.constants.PAGES.extend(pgs)
+        schema.config.PAGES.extend(pgs)
 
     fls: List[str]
     for fls in args.files:
-        schema.constants.FILES.extend(fls)
+        schema.config.FILES.extend(fls)
 
     if args.output:
-        schema.constants.OUTPUTDIR = args.output
+        schema.config.OUTPUTDIR = args.output
 
-    if args.autobuild or args.release or args.shacltests:
-        schema.constants.TERMS = ["ALL"]
-        schema.constants.PAGES = ["ALL"]
-        schema.constants.FILES = ["ALL"]
+    if args.autobuild or args.release or args.shacltests or args.buildrelease:
+        schema.config.TERMS = ["ALL"]
+        schema.config.PAGES = ["ALL"]
+        schema.config.FILES = ["ALL"]
 
     SchemaTerms.localmarkdown.Markdown.setWikilinkCssClass("localLink")
     SchemaTerms.localmarkdown.Markdown.setWikilinkPrePath("/")
@@ -164,7 +178,7 @@ def initialize() -> argparse.Namespace:
 
 def clear() -> None:
     if args.clearfirst or args.autobuild:
-        output_dir: Path = Path(schema.constants.OUTPUTDIR)
+        output_dir: Path = Path(schema.config.OUTPUTDIR)
         log.info(f"Clearing {output_dir} directory")
         if output_dir.is_dir():
             item: Path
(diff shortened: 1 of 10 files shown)


software/util/schema.py
```diff
@@ -84,3 +93,94 @@ def setVersion(ver: str, date: str) -> None:
     paths.DefaultInputLayout().domain_file(
         paths.Domain.ROOT, "versions.json"
     ).write_text(json.dumps(sort_dict(versiondata), indent=4))
+
+
+def load_stats_map(stats_dir: Optional[Path] = None) -> Dict[str, str]:
```
pascalfleury: Can we put this code in a new file, called buildstatspages.py or so?

software/util/schema.py
```diff
@@ -84,3 +93,94 @@ def setVersion(ver: str, date: str) -> None:
...
+    """Loads and parses the latest Google public stats JSON file into a dictionary."""
+    stats_map: Dict[str, str] = {}
+    
+    if stats_dir is None:
```
pascalfleury: This path logic should be added to the InputLayout class, the other files should have no business in computing paths like here.

templates/terms/InfoBlock.j2
```diff
@@ -3,8 +3,18 @@
...
     <div id="infohead">
-        <h1>{{ term.label }}</h1>
+        <h1 class="page-title">{{ term.label }}</h1>
         <div class="infotype">A Schema.org {{ TERMTYPE }}</div>
+        {% if google_stats_badge %}
+        <div class="public-stats-row">
+            Usage: 
+            <span class="public-stats-pill hover-tooltip" aria-label="Based on monthly aggregations from Google's web index.">
+                {{ google_stats_badge }} Domains
+                <span class="tooltiptext">Based on monthly aggregations from Google's web index.</span>
```
pascalfleury: should we already have the "Google" part coming from the data directly? it should be from the CSV file name, or directory, or in the CSV file directly. So if we add another provider of stats, this would come in directly.

software/util/schema.py
```diff
@@ -84,3 +93,94 @@ def setVersion(ver: str, date: str) -> None:
...
+            )
+        logging.warning(
+            f"Stats directory does not exist or is not a directory: {stats_dir}"
+        )
+        return {}
+
+    json_files = glob.glob(str(stats_dir / "????_??.json"))
+    json_files = [f for f in json_files if "summary" not in os.path.basename(f)]
+    if not json_files:
+        if not dev_mode:
```
pascalfleury: do we need this devmode at all? feels very clunky to me. Especially for releasing, this is one more dimension that is different between the release and the dev.

software/util/schema.py
```diff
@@ -15,6 +19,9 @@
 from util.sort_dict import sort_dict
 
 
+dev_mode = os.environ.get("SDO_DEV_MODE") == "1" or "unittest" in sys.modules
```
pascalfleury: elsewhere I claim we do not need this dev_mode. Aslo using the environment like so will be another way to shoot oneself in the foot.

software/util/buildtermpages.py
```diff
@@ -64,7 +65,12 @@ def termtemplateRender(term: sdoterm.SdoTerm, examples: List[schemaexamples.Exam
         elif ex.hasJsonld():
             exselect[3] = "selected"
         ex.exselect = exselect  # type: ignore
-
+    google_stats_badge: Optional[str] = None
```
pascalfleury: do not call it 'google_' in the code if we want to have more stats providers.

software/util/buildtermpages.py
```diff
@@ -64,7 +65,12 @@ def termtemplateRender(term: sdoterm.SdoTerm, examples: List[schemaexamples.Exam
...
+    if term and term.uri:
+        cleaned_uri = unicodedata.normalize("NFC", term.uri.strip())
+        if cleaned_uri.startswith("http://"):
```
pascalfleury: there is a function named protocolSwap somewhere for this.

software/util/buildtermpages.py
```diff
@@ -64,7 +65,12 @@ def termtemplateRender(term: sdoterm.SdoTerm, examples: List[schemaexamples.Exam
...
+            cleaned_uri = "https://" + cleaned_uri[7:]
+        google_stats_badge = schema.constants.STATS_MAP.get(cleaned_uri)
```
pascalfleury: can we pass the entire stats object in the constructor instead of passing through globals? I am trying to get away from these globals that make a lot of thigs difficult.

software/scripts/buildsite.py
```diff
@@ -229,8 +243,20 @@ def loadTerms() -> None:
         return
     LOADEDTERMS = True
     if not sdotermsource.SdoTermSource.SOURCEGRAPH:
-        with pretty_logger.BlockLog(logger=log, message="Loading triples files"):
-            sdotermsource.SdoTermSource.loadSourceGraph("default")
+        if FORCE_DEV:
```
pascalfleury: I would much prefer, as you have the 2 flags now, that we have to run the --buildrelease first (taking the latest) then have the --buildsite take from the latest built files.

i.e. no FORCE_DEV or such. It just reads from the last built version (which are dev files).

software/scripts/buildsite.py
```diff
@@ -229,8 +243,20 @@ def loadTerms() -> None:
...
+            with pretty_logger.BlockLog(logger=log, message="Loading development triples files (force_dev)"):
+                sdotermsource.SdoTermSource.loadSourceGraph("default")
+        else:
+            # Try to load from the release file first to build the site from release
+            protocol: str = "https" if sdotermsource.SdoTermSource.vocabUri().startswith("https") else "http"
+            release_file: str = f"data/releases/{schema.getVersion()}/schemaorg-all-{protocol}.ttl"
```
pascalfleury: thi should be read from the InputLayout class, no path computation in any other class. Also, you should get the version directly from there, no need to deal with it here.

software/scripts/buildsite.py
```diff
@@ -229,8 +243,20 @@ def loadTerms() -> None:
...
+            
+            if os.path.exists(release_file):
+                with pretty_logger.BlockLog(logger=log, message=f"Loading triples from release file {release_file}"):
+                    sdotermsource.SdoTermSource.loadSourceGraph(release_file)
+            else:
```
pascalfleury: I think we should hard fail here. this is also an indication that we did probably not create the buildrelease first. Silently loading something else makes releases more difficult.

software/util/schema.py
```diff
@@ -84,3 +93,94 @@ def setVersion(ver: str, date: str) -> None:
...
+                    if normalized_name.startswith("http://"):
+                        normalized_name = "https://" + normalized_name[7:]
+                    stats_map[normalized_name] = bucket
+            except Exception:
+                continue
+
+    return stats_map
+
+
+constants.STATS_MAP = load_stats_map()
```
pascalfleury: :-) I named it 'constants' so we get away from using this part as an argument passing...

software/util/paths.py
```diff
@@ -42,6 +42,8 @@ class Domain(str, Enum):
     ROOT = "root"
     DOCS_COLLAB = "docs/collab"
     DOCS_TERMFIND = "docs/termfind"
+    PUBLIC_STATS = "public_stats"
+    PUBLIC_STATS_GOOGLE = "public_stats_google"
```
pascalfleury: nit: I would call it "public_stats/google"

pascalfleury (approved):

software/util/paths.py
williamgslima: Addressed!

PR #4843, opened 2026-06-15
Add documentation for multiple values/order in the data model section

rrlevering:
Addressing #4806

### DIFF
--- a/docs/datamodel.html
+++ b/docs/datamodel.html
@@ -63,7 +63,7 @@ alongside other vocabulary that shares our basic datamodel and our use of underl
 and RDFa.
 </p>
 
-<h2>Multiple Values and Collections</h2>
+<h2 id="multiple-values">Multiple Values and Collections</h2>
 <p>
   By default, any property in Schema.org can have multiple values. In Microdata and RDFa,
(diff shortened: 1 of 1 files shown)


docs/datamodel.html
```diff
@@ -63,6 +63,65 @@ <h1>Data model</h1>
...
+  this is typically represented by repeating the property tag. In JSON-LD, it is
+  represented as an array of values.
+</p>
+<p>
+  Because of how the underlying RDF model works (representing data as independent triples),
+  these multiple values are technically treated as an unordered set by default. However, there are cases
+  where the order of values is important (e.g., the authors of a book, the steps in a recipe,
+  or items in a list). While practically most parsers assume source order, it is not explicit and there are
+  cases where that can be incorrect. To better support these cases, publishers may use
+  <a href="https://www.w3.org/TR/rdf-schema/">standard RDF collection and container semantics:
```
pascalfleury: this <a> tag is not closed.

pascalfleury (approved):

docs/datamodel.html
rrlevering: Hmm...this doesn't take into account the final version.  I fixed that in a follow-up commit.

PR #4847, opened 2026-06-21
Monthly Schema.org Stats Update

williamgslima:
Monthly Schema.org Stats Update

Automated PR to update Schema.org adoption stats based on the Google's latest internal aggregations.

### DIFF
--- a/data/public_stats/google/summary_2026_06.json
+++ b/data/public_stats/google/summary_2026_06.json
@@ -1,0 +1,62 @@
+[
+  {
+    "Class":"Itemtype",
+    "Domain Bucket":"< 1K",
+    "count":475
+  },
+  {
+    "Class":"Itemtype",
+    "Domain Bucket":"1K - 10K",
+    "count":241
+  },
+  {
+    "Class":"Itemtype",
+    "Domain Bucket":"10K - 100K",
+    "count":150
+  },
+  {
+    "Class":"Itemtype",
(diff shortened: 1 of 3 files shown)


data/public_stats/google/summary_2026_06.json
pascalfleury, line 1: what's the difference with the 'summary_2026_06.json' file?

rrlevering: I don't understand this comment.  This is the aggregation file for the total rollup of each bucket.

rrlevering (approved):

data/public_stats/google/summary_2026_06.json
pascalfleury: typed the wrong name. the files `2026_06.json` and `summary_2026_06.json` seem to have the same contents.

williamgslima: Actually, that is not the case. Here is the structure used for each item in these files:

2026_06.json (granular URL mappings showing the respective bucket for that schema):

json
  {
    "Class": "Itemtype",
    "Name": "http://schema.org/AboutPage",
    "Domain Bucket": "100K - 1M"
  }
  
  
summary_2026_06.json (aggregated stats showing the number of schemas within that bucket):

json
  {
    "Class": "Itemtype",
    "Domain Bucket": "100K - 1M",
    "count": 43
  }

PR #4881, opened 2026-08-12
examples: Minor fixes for syntax and typos.

pascalfleury:
(originally from @dpb587)

Data-related value changes:

Duration: use P prefix for valid ISO 8601 syntax.
temporalCoverage: use correct / delimiter for valid ISO 8601 ranges.
Enumeration: fix outdated/incorrect references.
Encoding-related syntax changes:

Microdata: add missing itemscope attributes.
RDFa: add missing vocab attributes; fix incorrect attributes; drop rdfa:usesVocabulary conversion artifact from JSON-LD.
JSON-LD: fix duplicate object keys for valid syntax; drop unused @context entry conversion artifacts.
And a couple other corrections to match sibling examples and to use more-valid stub values.

### DIFF
--- a/data/examples.txt
+++ b/data/examples.txt
@@ -399,7 +399,7 @@ MICRODATA:
  <meta itemprop="contentUrl" content="http://media.freesound.org/data/0/previews/719__elmomo__12oclock_girona_preview.mp3" />
 
 <span class="description">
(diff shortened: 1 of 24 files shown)


pascalfleury:
dpb587@ has amended his PR, so this one is now moot.

PR #4896, opened 2026-09-02
Fix a few glitches in the file and site production

pascalfleury:
Some outputs contain minor issues, none of them changing the semantics of the schema though.

1. Inverse properties are now populated earlier (aka `resolve`d), so that they generate the needed info in pages (see `AutoRental` and its `funding` property).
2. The double breadcrumbs (see `minPrice`) is now fixed (does not produce the breadcrumb if it points outside of schema.org).
3. The ordering in XML files for certain nodes is now deterministic (by manually sorting the nodes).
4. Numbering of examples got more stability when new examples are added by checking in more modes.

### DIFF
--- a/software/SchemaTerms/sdoterm.py
+++ b/software/SchemaTerms/sdoterm.py
@@ -42,6 +42,12 @@ class SdoTermOrId:
         self._term_id = term.id if term else term_id
         self._term = term
 
+    def _resolve(self) -> Optional["SdoTerm"]:
+        if not self._term and self._term_id:
+            from SchemaTerms.sdotermsource import SdoTermSource
+            self._term = SdoTermSource.getTerm(self._term_id)
+        return self._term
+
     @property
     def expanded(self) -> bool:
         return not self._term_id or bool(self._term)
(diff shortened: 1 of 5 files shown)


software/SchemaTerms/sdoterm.py
```diff
@@ -42,6 +42,12 @@ def __init__(self, term_id: Optional[str] = None, term: Optional["SdoTerm"] = No
         self._term_id = term.id if term else term_id
         self._term = term
 
+    def _resolve(self) -> Optional["SdoTerm"]:
+        if not self._term and self._term_id:
+            from SchemaTerms.sdotermsource import SdoTermSource
```
jlndr, line 47: qq: why is this needed? Shouldn't `_term` or `_term_id` be set in the constructor (if any)? Is the usage of this class not doing what we expect, i.e. why should `.pending()` work if the class was created with only a `term_id`?

If we think it should, we could just do this resolving in the constructor perhaps?

jlndr (approved):

software/SchemaTerms/sdoterm.py
pascalfleury: It has to do with the the fact that when being loaded, the backlinks are not populated, and then the information that some properties are actually still in pending does not make it to certain generated pages.

